PR #17, opened 2017-03-28
Added base conversions.

shivhek25:


dynamitechetan:
@shivhek25  please check the merge conflicts

Bubble Sort.java
```diff
@@ -2,22 +2,24 @@
 
 class BubbleSort
 {
-	public static void main(String[] args) 
+	public static void main(String[] args)
 	{
-		int array[]=new int[6];
+		int n;
+		int array[]=new int[n];
```
DocJava: "n" has no value here, this will break

Bubble Sort.java
```diff
@@ -2,22 +2,24 @@
...
 		Scanner input=new Scanner(System.in);
-
+    System.out.println("Enter length of array:");
+		n=sc.nextInt();
```
DocJava: now that n has a value we can initialize the "array" variable down here

Decimal to Hexadecimal.java
```diff
@@ -0,0 +1,24 @@
...
+{
+    public static void main(String args[])
+    {
+        Scanner sc=new Scanner(System.in);
+        int n,k,i,d;
+        String s=" ";
+        int a[]={'A','B','C','D','E','F'};
+        n=sc.nextInt();
+        k=n;
+        while(k!=0)
```
DocJava: infinite while loop, the value of k is never modified

shivhek25:
Made the requisite changes. Kindly check.

Decimal to Hexadecimal.java
shivhek25: Updated it.

Bubble Sort.java
shivhek25: Made the changes.

shivhek25:
Can the admins and reviewers please look into the updations ? Thank you. @dynamitechetan @DocJava @AnupKumarPanwar

DocJava:
Looks fine, I would suggest using readable variable names (instead of i k n j or other random letters).
I still vote taking a look at the Java code style guide https://google.github.io/styleguide/javaguide.html 
The GCD file (looks fine, I did not run it, but here are some notes): 
-The main method throws exception , but the entire code is caught in a try catch..so nothing would ever be thrown.
-I'm pretty sure the try catch would never actually catch anything? 
-The catch has "return" in it, but it is the last thing to happen in a block, so really you could just leave it empty or something and have the same effect?

shivhek25:
Sure. Should I change the variable names for these codes as well, or should I keep this pointer in mind next time onwards ? @DocJava

DocJava:
That is up to you, the coder. (I vote to change it now though)

r0hit-gupta:
@shivhek25 Please resolve the conflicts.

shivhek25:
I have already made all the changes. Could you please specify which conflicts are to be resolved ?

0oo0:
There are conflicts in BubbleSort.java
You can see them on Github.
Merge conflicts arise when you compete for the same line change by different branches. 
Try this
https://help.github.com/articles/resolving-a-merge-conflict-on-github/

shivhek25:
The resolve conflicts is deactivated so I followed the instructions on how to resolve the error using cmd. After having done those, the file is still under conflict.

dynamitechetan:
if you are not able to resolve the conflicts then close this PR and send a new one with your changes.

PR #53, opened 2017-06-01
Update countwords.java

MarcHines:
Easier to read. Also, takes into account null and empty string possibilities

### DIFF
--- a/countwords.java
+++ b/countwords.java
@@ -4,30 +4,21 @@ import java.util.Scanner;
  * You enter a string into this program, and it will return how
  * many words were in that particular string
  * 
- * @author Unknown
+ * @author Marcus
  *
  */
-class CountTheWords
-{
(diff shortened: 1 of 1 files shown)


countwords.java
```diff
@@ -4,30 +4,21 @@
  * You enter a string into this program, and it will return how
  * many words were in that particular string
  * 
- * @author Unknown
+ * @author Marcus
  *
  */
-class CountTheWords
```
AnupKumarPanwar: Why removed the class?

MarcHines:
That is an honest mistake. Class should be there.
________________________________

MarcHines:
Also...returning-1 for bad input is weird in this case. Returning 0 instead makes more sense. Another mistake on my part. Sorry :)
________________________________

PR #54, opened 2017-06-01
Update ReverseString.java

MarcHines:
Using recursion to reverse a String is not a good approach


AnupKumarPanwar:
You are directly using the reverse function. But this repo is about the algorithm's implementation in Java and atleast for an independent string reversal program you should not use built-in function here.

PR #70, opened 2017-07-15
changed LinkForLinkedList class name to Node

earngpi:
- I think it's less confusing if we directly call it Node, especially for beginners.
- Also removed displayLink() method, added getValue() instead. Should be more useful when used in other applications.


AnupKumarPanwar:
Please resolve the conflicts

PR #77, opened 2017-08-06
Add files via upload

ParitoshAggarwal:
HashMap implemented. 
Code is working and checked twice to avoid any errors.
Self Explanatory code with handy variable names as are there in actual libraray


ParitoshAggarwal:
Generic Tree added 
Code is working and checked twice to avoid any errors.
Code is Explained briefly.

Issue #65, opened 2017-07-08
Generalize using generics.


KylerSmith:
# Enhancement
## Description
Many of the algorithms, namely sorting, searching and the data structures, are using primitive types for the implementation of the algorithms. This could be generalized to any comparable type  using generics.

varunu28:
Thanks for pointing that out @KylerSmith . @AnupKumarPanwar I would like to work on this issue. Can you please assign it to me

varunu28:
Here is the linear search with generics #80 
Would start working on others and would add more search and sort methods in future

0oo0:
Thanks.

varunu28:
Added generic binary search #82 
Done with the searches. Would be adding generic sorts in future. 
Once both search and sorts are done let us make sure that any further addition of algorithms is in its generic form

varunu28:
Started with creating generic version for sorts
Bubble Sort #83

dynamitechetan:
Can we close this issue?

shubhamtewari:
hey guys, should i alppy generics to the DataStructures?

varunu28:
@shubhamtewari  Can you please point out which data structures, in particular, you want to apply generics?

shubhamtewari:
I can start with stacks, then lists, graphs, queues, and heaps.
Essentially I want to apply generics to all the data structures in the
DataStructures directory.
But stated above are where I'd like to start.

varunu28:
Sure. Go ahead.
You can open a PR once you are done with the changes. Go for refactoring the code rather than starting with a blank slate

PR #87, opened 2017-08-21
Updated QuickSort.java

varunu28:
Updated quicksort to its generic version and changed the filename according to convention

#65


varunu28:
Updated Merge sort to its generic version

#65

PR #97, opened 2017-09-28
Added generic shell sort implementation in Sorts folder

Kriztoper:
Added generic shell sort implementation and tested on Integer array and String array input based on other sorting implementation's input arrays like bubble sort.


dynamitechetan:
please resolve conflicts

Issue #96, opened 2017-09-28
Add more algorithms


AnupKumarPanwar:
This repo has less contributions. Please add more algorithms. 

See the structure and README of this repo for reference
https://github.com/TheAlgorithms/Python

dhinske:
would you be interested in some Set-implementations or maybe Bag? didnt see it in your repo :)

dpunosevac:
I would like to contribute it. I see that you have no shell sort algorithm for Java. I would take that one first. Than later I would continue with few mores, if that is ok.

mpokryva:
Would you be interested in a level-order tree traversal?

dpunosevac:
@mpokryva Sure man. I am up for anything. :)

dynamitechetan:
Thanks. @mpokryva

mpokryva:
Ok, I will take that then.

dpunosevac:
@mpokryva I mistakenly thought you were talking to me. And not asking for yourself. Anyway, you can grab that one, and I hope that @dynamitechetan needs shellsort, so I do that one.

mpokryva:
Ok, sounds good! Sorry for the misunderstanding :)

varunu28:
@dynamitechetan  @AnupKumarPanwar  Would be continue to add more algorithms to this as a part of hacktoberfest
PRs would be maily in Data Structures and other domain

rei2hu:
Please look over pull requests before merging because I noticed the most recent one won't even compile properly and I'm assuming that a lot of people will make PRs soon because of Hacktoberfest (myself included). For example, in PR #101 there are many errors.

Can't compile because there is no type for `len_a` and `len_b`
Can't compile because `calculate_distance` is a non-static method
One of the loops will go on forever because it's declared as `for(int j = 0; i < len_b; j++)` and `i` is not getting updated within the loop.

varunu28:
@rei2hu @dynamitechetan @AnupKumarPanwar 
Also the PRs should be monitored correctly to check for coding style and if an addition to the repo is actually an algorithm or just a solution of an interview question. The long term goal should be to make this repo as a one stop solution for Java algorithms rather than a repository with bunch of random interview problems.

AnupKumarPanwar:
@dynamitechetan @varunu28 @rei2hu 
I think we should create a repo that contains simple English explanation of the algorithms with example and a link to their implementation in different programming languages.

praveenkannan:
[...]

PR #103, opened 2017-09-28
Added level order traversal, and more nodes in main method

mpokryva:
1) Added level order traversal.
2) Added more nodes to example tree, to have a complete BST of height 2.
#103


mpokryva:
#96

mpokryva:
Sure! Will add some soon.

PR #114, opened 2017-09-29
Create AnyBaseToAnyBase

RollandMichael7:
Class for converting a number between any two bases, by going through decimal.


dynamitechetan:
file name does not end in .java

varunu28:
@RollandMichael7 Thanks for the commit. Just a heads up that the code doesn't handles a case where I enter a number of wrong base. We can have a check if the number being input by the user is actually in the correct beginning base
Ex:
Enter number: 1104
Enter beginning base: 2
Enter end base: 10
16

This should prompt the user with an error

We can look for this functionality in future. I will also think about a method to handle this case
Thanks :+1:

RollandMichael7:
Thanks for pointing that out! 

I have an idea about how to handle that. I'll make a pull request by the end of the day :)

PR #117, opened 2017-10-01
maven and unit testing

emergie:
Currently project has no structure, each java file is isolated, results are expected on the terminal.

Proposition introduces unit testing and maven dependency&lifecycle management.

For this PR I refactored BinaryTree.java and wrote 3 simple tests of tree traversal.
If applied further such refactoring should enable introduction of automated testing in the future.

```
mvn clean test
```

```
(...)
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running tree.BinaryTreeTest
Configuring TestNG with: org.apache.maven.surefire.testng.conf.TestNG652Configurator@6ae40994
BinaryTree of [3, 7, 2, 8, 1] inOrder traversal: [1, 2, 3, 7, 8]
BinaryTree of [3, 7, 2, 8, 1] postOrder traversal: [1, 2, 8, 7, 3]
BinaryTree of [3, 7, 2, 8, 1] preOrder traversal: [3, 2, 1, 7, 8]
Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.199 sec

Results :

Tests run: 3, Failures: 0, Errors: 0, Skipped: 0
```


varunu28:
@emergie  Great initiative. Adding tests for algorithms currently present is of higher priority than adding algorithms. Also we can make sure that in future the algorithms added to the repo contain relevant test cases. 

I can collaborate with you to add test cases for the existing algorithms.

@dynamitechetan @AnupKumarPanwar  Lets discuss a structure for this

dynamitechetan:
Yes. adding tests is a good idea. @varunu28 Do you have any suggestion on the structure of this?
Also @emergie  please resolve conflicts. ;D

varunu28:
@dynamitechetan We can convert the complete repo into a java project structure. It will have a source folder comprising of Java files and a test folder with test files. Using a uniform coding style would also help so we can use a common tool such as IntelliJ. This will take some time but would bring a structure to the repo and remove the clutter so that people can directly use this repo as a Java project

dynamitechetan:
I think that the way we have our project right now, it is easier to find the files and look for references. If we convert it in a java style project it will be difficult to find simple algorithms like finding primes.

varunu28:
For improving upon this we can update our README with a ready reference for all the algorithms. So if someone wants to find a single algorithm like the one you mentioned then they can just go through the list provided on README. Otherwise if people want to change something then having a project structure would help as they can build the project from the source using IDEs.

Another approach:
We can also look an option of having a Wiki for listing all our algorithms in all language specific repos.

Otherwise the sister-repo "Algorithms-Explainations" can also help in providing a one stop shop for all the algorithms and language specific repos can be used for adding code for algorithms.

emergie:
@dynamitechetan
conflicts resolved

By the way - changes that introduced conflict brooke `BinaryTree#put` implementation.
After mpokryva changes in `BinaryTree#find` it was not possible to construct more than 1-node tree using `BinaryTree#put`.
https://github.com/TheAlgorithms/Java/commit/e5381585a5329af6a1ad2d0b5ababef11cba4062#diff-ae650ede9a179f475362e199b7c54d1cR68

If there were unit tests + good coverage it would be trivial to detect that regression.

emergie:
My advices

### Project structure/layout

Maven introduced a standard: https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html  
From my experience it is good to follow those conventions.   
Gradle also uses them: https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_project_layout

### Tests structure

It is good to use BDD given/when/then style.

https://martinfowler.com/bliki/GivenWhenThen.html


### Code style

It would be nice to enforce https://google.github.io/styleguide/javaguide.html

Current codebase is a mess.  

For example `AVLTree.java` contains a piece of code that doesn't even compile - programmer decided not to use `{}` after `for` and it ended up like this:

``` java
private void setBalance(Node... nodes) {
    for (Node n : nodes)
        reheight(n);
        n.balance = height(n.right) - height(n.left);
}
```
Indentation clearly shows author's intentions, but tabs don't define a code block in java.

If proper code style was enforced this wouldn't happen:  
https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

--

This repo is intended for beginner programmers.  
IMO aside of teaching the algorithms the code should also follow best practices to help juniors adopt software engineering standards.

varunu28:
@emergie I went through this and found that it is pretty good. As there are no tests on current master and I am sure there are multiple algorithms which would fail on borderline test cases. But pushing the tests on the current master would confuse someone who is new to the repo. What I suggest is to start working on a new branch where we can keep all the code in the main and their test cases in the test folder. Once there are enough test cases to test the algorithms, we can merge it with the master branch.

@dynamitechetan @AnupKumarPanwar  Please let me know what you think about this?

PR #127, opened 2017-10-02
add precision_root_algorithm

dpacmen:
calculate root upto any precision p


dynamitechetan:
can you please also add relevant comments.

PR #134, opened 2017-10-03
Program for calculating Large Factorial Added

Ankit-Dhankhar:
This program used stack instead of traditional data-types thus value of factorial that can be calculated depend on computer power not the range of value supported by data-type in reference to issue #96

### DIFF
--- a/Misc/LargeFactorial.java
+++ b/Misc/LargeFactorial.java
@@ -1,0 +1,32 @@
+import java.util.Scanner;
+import java.util.Stack;
+class LargeFactorial
+{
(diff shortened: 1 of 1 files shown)


dynamitechetan:
please add some relevant comments.

Balance-Breaker:
@dynamitechetan added comments, can you check  the same if any further changes required

varunu28:
@Balance-Breaker Please add structure to the code. Try to keep the main method as clean as possible and put the main logic in a separate function which you are going to call from the main method.

PR #137, opened 2017-10-03
turned some public methods private

daniel-mueller:
Some public methods could be set private. Furthermore the static keyword was missing in two cases.


dynamitechetan:
Thanks.

PR #142, opened 2017-10-03
Palindromic Primes

manmeet-22:
#96


dynamitechetan:
can you add some comments to explain what this code does?

GammaBurst101:
@dynamitechetan , I think the code finds such numbers which are both prime and palindrome between 2 and the input provided by the user. Is there a way I can help?

PR #148, opened 2017-10-04
Added Trie Implementation in Java

udayvig:
Contains functions like add a word, remove a word, search for a particular word, etc.


dynamitechetan:
https://github.com/TheAlgorithms/Java/blob/master/data_structures/Trees/TrieImp.java
Already Present.

PR #163, opened 2017-10-08
Added Comb Sort

Kriztoper:
Added generic comb sort implementation and tested on Integer array and String array input based on other sorting implementation's input arrays like bubble sort.


varunu28:
@Kriztoper  Please add the description comments for CS above the CS function. Currently it shows an error when loaded to an IDE.
Also please add a description for the getNextGap function.

Thanks for the commit. I was able to learn something new :+1:

PR #164, opened 2017-10-08
Add: level order traversal.

KyleScharnhorst:
Prints each level of the bst.
I.e., Root -> its children -> its children's children -> and so on...


varunu28:
Thanks for the commit @KyleScharnhorst

PR #166, opened 2017-10-08
Add: circular buffer implementation.

KyleScharnhorst:
data structure used for producer consumer situations. Allows reading and
writing so long as there is data and the buffer is not full,
respectively.


dynamitechetan:
Thanks

PR #182, opened 2017-10-11
Add Linear Congruential Generator

TobCar:


varunu28:
Thanks for the PR @TobCar Great algorithm

PR #183, opened 2017-10-11
Add validForBase() and min/max base constants

RollandMichael7:
validForBase() ensures that the entered number is valid for the beginning base. Additionally, constant fields for the minimum/maximum bases you want to accept from the user were added; however, max base should not exceed 36 (after which its unclear how to represent digits >36)


varunu28:
Thanks @RollandMichael7

PR #189, opened 2017-10-12
Added Tower of Hanoi recursive algorithm and Longest Common Subsequence (LCS) algorithm

rahul1995:
Tower Of Hanoi algorithm added and LCS algorithm is also added


varunu28:
@rahul1995 Please remove Towers of Hanoi
Its already present in the Misc folder

rahul1995:
@varunu28 Please maintain the same structure in both Python and Java algorithms repo. I found Towers of Hanoi in "Others" folder in python actually so checked only "Others" folder in java

varunu28:
@rahul1995 Thanks for pointing this out. The directory structure has been changed

#196

PR #190, opened 2017-10-13
Add a bag data type

KevinMFinch:
Add a bad data type. Will be used for implementing Digraphs


varunu28:
@KevinMFinch  Please remove the class files

PR #200, opened 2017-10-13
Created new file in Sorts directory, CountingSortIntegers.java

youssefAli11997:
Implemented the Counting Sort Algorithm for sorting integers that are within the range [0:10^8].
It sorts only integer arrays especially positive integers since it uses array elements as indexes in the frequency array.


varunu28:
@youssefAli11997  Thanks for the PR. Can you implement a generic version of same. 
Or I can merge this PR and modify the code to its generic version.

youssefAli11997:
@varunu28 I created a generic version of my counting sort method.

PR #203, opened 2017-10-14
implementation of queues using stacks(both enqueue efficient and dequeue efficient)

pranjal4697:


varunu28:
Kindly keep a single java file for both the stacks and the trie

pranjal4697:
Will keep it in mind for the future codes. Thanks.

PR #206, opened 2017-10-15
Prim's And kruskal's Algorithms

Mansi-Mittal:
Added prim's and kruskal's Algorithms


varunu28:
@Mansi-Mittal Thanks for the PR

PR #230, opened 2017-10-19
I made this easier to understand for newcomers.

karseyr:
The previous version was very complex, and could be hard to follow.  It's my understanding that this repository is supposed to be reference implementations of each algorithm contained.  These should be highly accessible to newbs and pros alike.  This update levels the playing field. 

Feel free to comment on lines to explain their purpose.

### DIFF
--- a/Sorts/BubbleSort.java
+++ b/Sorts/BubbleSort.java
@@ -1,70 +1,36 @@
 /**
  *
- * @author Varun Upadhyay (https://github.com/varunu28)
+ * @author Karsey Renfert (https://github.com/karseyr) 
  *
  */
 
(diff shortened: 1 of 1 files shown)


varunu28:
We already have a well-tested  and generic implementation for Bubble sort and as far as the explanation is concerned, it is present [here](https://github.com/TheAlgorithms/Algorithms-Explainations/blob/master/Sorting%20Algorithms/Bubble%20Sort.md)

If you want to add explanations for proper understanding of any algorithm then you can raise a PR for [Algorithms-Explainations](https://github.com/TheAlgorithms/Algorithms-Explainations)

PR #236, opened 2017-10-20
Update Bubble sort

SYED-AFSAHUL:
Added comments. Made code more readable.


varunu28:
@SYED-AFSAHUL The code is pretty self-explanatory as of now. You don't need to add that a for loop is going to do an iteration or a compare operator is going to compare two values.

You can look towards other options such as machine learning algorithms to contribute to the repo.

PR #266, opened 2017-10-25
Create Dijkstra_shortest_path.java

PAPERPANKS:
- Adding shortest path example
- reference taken from [here](http://www.geeksforgeeks.org/java-program-for-dijkstras-algorithm-with-path-printing/)

In reference to issue #96

### DIFF
--- a/Dynamic Programming/Dijkstra_shortest_path.java
+++ b/Dynamic Programming/Dijkstra_shortest_path.java
@@ -1,0 +1,87 @@
+import java.util.Scanner;
+
+public class Dijkstra
+{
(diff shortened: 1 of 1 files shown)


dynamitechetan:
please check https://github.com/TheAlgorithms/Java/blob/master/Others/Dijkshtra.java
is it similar?

PAPERPANKS:
@dynamitechetan it's little bit similar around `20%` but not the same. also i find mine easier to understand and implement

dynamitechetan:
if changes can be made please update the existing file. please do not make a new file.

PR #269, opened 2017-10-25
Horner’s Method for Polynomial Evaluation

silent-lad:
This algorithm provides a function by which we can evaluate the value of any polynomial at a given X via Horner's method.

### DIFF
--- a/Others/horner.java
+++ b/Others/horner.java
@@ -1,0 +1,32 @@
+// Java program for implementation of Horner Method 
+// for Polynomial Evaluation
+import java.io.*;
+ 
+class HornerPolynomial
+{
+    // Function that returns value of poly[0]x(n-1) +
+    // poly[1]x(n-2) + .. + poly[n-1]
+    static int horner(int poly[], int n, int x)
+    {
+        // Initialize result
+        int result = poly[0];  
+  
+        // Evaluate value of polynomial using Horner's method
+        for (int i=1; i<n; i++)
+            result = result*x + poly[i];
+  
+        return result;
+    }
+     
+    // Driver program
+    public static void main (String[] args) 
+    {
+        // Let us evaluate value of 2x3 - 6x2 + 2x - 1 for x = 3
+        int[] poly = {2, -6, 2, -1};
+        int x = 3;
+        int n = poly.length;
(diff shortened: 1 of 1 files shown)


Others/horner.java
```diff
@@ -0,0 +1,32 @@
+// Java program for implementation of Horner Method 
+// for Polynomial Evaluation
+import java.io.*;
```
christianbender: Please remove this unused header.

Others/horner.java
```diff
@@ -0,0 +1,32 @@
...
+{
+    // Function that returns value of poly[0]x(n-1) +
+    // poly[1]x(n-2) + .. + poly[n-1]
+    static int horner(int poly[], int n, int x)
+    {
+        // Initialize result
+        int result = poly[0];  
+  
+        // Evaluate value of polynomial using Horner's method
+        for (int i=1; i<n; i++)
```
christianbender: Use curly braces for the for-loop.

Others/horner.java
```diff
@@ -0,0 +1,32 @@
+// Java program for implementation of Horner Method 
+// for Polynomial Evaluation
+import java.io.*;
+ 
+class HornerPolynomial
+{
+    // Function that returns value of poly[0]x(n-1) +
+    // poly[1]x(n-2) + .. + poly[n-1]
+    static int horner(int poly[], int n, int x)
```
christianbender: The argument ```n``` is unnecessary. Because you can get the length of the array by ```length``` property.

Others/horner.java
```diff
@@ -0,0 +1,32 @@
...
+        return result;
+    }
+     
+    // Driver program
+    public static void main (String[] args) 
+    {
+        // Let us evaluate value of 2x3 - 6x2 + 2x - 1 for x = 3
+        int[] poly = {2, -6, 2, -1};
+        int x = 3;
+        int n = poly.length;
```
christianbender: unnecessary. see above

Others/horner.java
```diff
@@ -0,0 +1,32 @@
+// Java program for implementation of Horner Method
```
christianbender: Please rename your file according to the class name.

PR #276, opened 2017-10-27
Comments&fixes

KennethNero:
Lots of small quality of code and life fixes for the project. This includes removing package headers (as seems to be the style), cleaning up unused imports, and generally beautifying any clearly out of place code. Theres more left to do, but this was a good start.


dynamitechetan:
thanks

PR #277, opened 2017-10-27
String split using delimiter

Kromzem:
added small algorithm to split string by specified delimiter instead of regex


varunu28:
There is already an effective way to do this in Java using built in libraries. This algorithm is recreating the wheel. Closing it & find the implementation to do this below

String[] strs = s.trim().split("\\s+");

PR #285, opened 2017-10-28
JumpSearch.java

sandeeproy99:
Like Binary Search, Jump Search is a searching algorithm for sorted arrays. The basic idea is to check fewer elements (than linear search) by jumping ahead by fixed steps or skipping some elements in place of searching all elements.


sandeeproy99:
Sir, should i write a Scanner method to take inputs?

christianbender:
@sandeeproy99 
You write a separate method for searching which accepts all type of data.  
Tip: use the data type ```object``` (as generic type)

PR #288, opened 2017-10-28
Fixed files and folders name conventions

freitzzz:


varunu28:
@freitzzz  Thanks for the changes. Can you please open another PR for correctly naming the file i.e. the filename should match the class name in the file. Also if the class name is somewhat ambiguous then rename it to something which is more clear. Thanks

PR #295, opened 2017-10-31
Odd Even Sort

rohitbnag88:


Sorts/OddEvenSorter
```diff
@@ -0,0 +1,52 @@
+import java.io.*;
+
```
christianbender: Please name your file with extension ```.java```

Sorts/OddEvenSorter
```diff
@@ -0,0 +1,52 @@
...
+    {
+        boolean isSorted = false; // Initially array is unsorted
+ 
+        while (!isSorted)
+        {
+            isSorted = true;
+            int temp =0;
+ 
+            // Perform Bubble sort on odd indexed element
+            for (int i=1; i<=n-2; i=i+2)
```
christianbender: You can use the operator ```+=``` instead of ```i=i+2```

Sorts/OddEvenSorter
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    temp = arr[i];
+                    arr[i] = arr[i+1];
+                    arr[i+1] = temp;
+                    isSorted = false;
+                }
+            }
+ 
+            // Perform Bubble sort on even indexed element
+            for (int i=0; i<=n-2; i=i+2)
```
christianbender: You can use the operator ```+=``` instead of ```i=i+2```

Sorts/OddEvenSorter
```diff
@@ -0,0 +1,52 @@
+import java.io.*;
+ 
+class OddEvenSorter
+{
+    public static void oddEvenSort(int arr[], int n)
```
christianbender: You can avoid the argument ```n```. Use the ```length``` attribute of the array.

Sorts/OddEvenSorter
```diff
@@ -0,0 +1,52 @@
...
+ 
+        return;
+    }
+    public static void main (String[] args)
+    {
+        int arr[] = {34, 2, 10, -9};
+        int n = arr.length;
+ 
+        oddEvenSort(arr, n);
+        for (int i=0; i < n; i++)
```
christianbender: You can avoid the for-loop. With  ```import java.util.Arrays;``` and  
```java
System.out.println(Arrays.toString(array));
```
You can print the full array.  
And for the future use curly braces by structures like loops, and if-else!

Sorts/OddEvenSorter
```diff
@@ -0,0 +1,52 @@
...
+                {
+                    temp = arr[i];
+                    arr[i] = arr[i+1];
+                    arr[i+1] = temp;
+                    isSorted = false;
+                }
+            }
+        }
+ 
+        return;
```
christianbender: You can remove the ```return;```

PR #298, opened 2017-10-31
.

NISHITA97:


NISHITA97:
I've made the changes according to the review. Please review them and let me know if I missed out something.

PR #301, opened 2017-11-01
Update Palindrome.java

miguel3010:


varunu28:
@miguel3010  The existing implementation is much more intuitive and gives a better understanding of the palindrome concept. Closing the PR

PR #305, opened 2017-11-12
Extended knapsack

LeeChungWan:
My pull request is an extended Knapsack. Existing knapsack shows only the maximum value. My implementation of knapsack receives input from the user.
1. Enter the number of items and weight of knapsack.
2. Enter the item's value and weight in the order of the number of items entered.
3. Returns the index of the item that makes up the knapsack table and its maximum value.

Please take a look at my code and answer it.


LeeChungWan:
result

This img is result console.

LeeChungWan:
I've made all of your requests.

PR #308, opened 2017-11-13
ClosestPair

YooSeonjae:
The closest pair
- Find the nearest pair of n points
- Applied only in 2D.
- Input Data = {(xi,yi) | 1<=i<=n}


varunu28:
An interview question. Not an algorithm. Closing it

PR #322, opened 2017-11-20
Added Ring Queue(=Circular Queue) 

YooSeonjae:
circular queue or ring queue is a data structure that uses a single, fixed-size buffer as if it were connected end-to-end.

When you show a queue with an array using the Ring Queue, it shows that when you call remove (), there is no need to move a lot because of its free space.

### DIFF
--- a/Data Structures/Queues/RingQueue.java
+++ b/Data Structures/Queues/RingQueue.java
@@ -1,0 +1,94 @@
+public interface Queue {
+
+	public void add(Object object);
+
+	public Object first();
+
+	public Object remove();
+
+	public int size();
+
+}
+public class RingQueue implements Queue {
+	private Object[] objArray = null;
+	private int size;
+	private int front;
+	private int rear;
+
+	public RingQueue(int n) {
+		// Create an array of size n
+		objArray = new Object[n];
+		size = 0;
+		front = 0;
+		rear = 0;
+	}
+
+	@Override
+	public void add(Object obj) {
+		// type obj if there is empty space in the queue
+		// if there is no empty space in the queue throw new IllegalStateException ("The queue is full");
+		if (size == objArray.length)
+			throw new IllegalStateException("The queue is full");
+
+		else {
+			rear = (rear + 1) % objArray.length;
+			objArray[rear] = obj;
+			size++;
+		}
+
+	}
+
+	@Override
+	public Object first() {
+		// return the first element if the queue is not empty
+		// if the queue is empty throw new IllegalStateException ("The queue is empty");
+		if (size == 0)
+			throw new IllegalStateException("The queue is empty");
+
+		return objArray[(front + 1) % objArray.length];
+	}
+
+	@Override
+	public Object remove() {
+		// If the queue is not empty, delete the first element and return
+		// if the queue is empty throw new IllegalStateException ("The queue is empty");
+		Object object = new Object();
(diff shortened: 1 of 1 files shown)


Data Structures/Queues/RingQueue.java
```diff
@@ -0,0 +1,94 @@
...
+	private Object[] objArray = null;
+	private int size;
+	private int front;
+	private int rear;
+
+	public RingQueue(int n) {
+		// Create an array of size n
+		objArray = new Object[n];
+		size = 0;
+		front = 0;
```
christianbender: The problem is that for ```(front+1) % objArray.length``` the pointer ```front``` never reaches the element with index 0 of the array. I would ```front``` set equal to ```-1```.

Data Structures/Queues/RingQueue.java
```diff
@@ -0,0 +1,94 @@
...
+		rear = 0;
```
christianbender: The same thing as above.

Data Structures/Queues/RingQueue.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	@Override
+	public void add(Object obj) {
+		// type obj if there is empty space in the queue
+		// if there is no empty space in the queue throw new IllegalStateException ("The queue is full");
+		if (size == objArray.length)
```
christianbender: Please use curly braces. See here for if-else: http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-142311.html#449

Data Structures/Queues/RingQueue.java
```diff
@@ -0,0 +1,94 @@
...
+			size++;
+		}
+
+	}
+
+	@Override
+	public Object first() {
+		// return the first element if the queue is not empty
+		// if the queue is empty throw new IllegalStateException ("The queue is empty");
+		if (size == 0)
```
christianbender: Same thing. Use curly braces

Data Structures/Queues/RingQueue.java
```diff
@@ -0,0 +1,94 @@
...
+		return objArray[(front + 1) % objArray.length];
+	}
+
+	@Override
+	public Object remove() {
+		// If the queue is not empty, delete the first element and return
+		// if the queue is empty throw new IllegalStateException ("The queue is empty");
+		Object object = new Object();
+
+		if (size == 0)
```
christianbender: Use curly braces, too.

Data Structures/Queues/RingQueue.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	@Override
+	public String toString() {
+		StringBuffer sb = new StringBuffer();
+
+		if (objArray == null)
+			return sb.toString();
+
+		if (front == rear)
```
christianbender: Use curly braces, too.

Data Structures/Queues/RingQueue.java
```diff
@@ -0,0 +1,94 @@
...
+	public int size() {
+		// returns the number of elements stored in the queue
+		return size;
+	}
+
+	@Override
+	public String toString() {
+		StringBuffer sb = new StringBuffer();
+
+		if (objArray == null)
```
christianbender: Use curly braces, too.

PR #327, opened 2017-11-26
add function divide

JeonSeongBae:


JeonSeongBae:
second commit is removing package. Please check again.

PR #328, opened 2017-11-26
Add Edmond's algorithm for maximum flow

flow85:
Edmonds algorithm for maximum flow is an extention from the Ford Fulkerson method for finding the maximum flow in a flow network. It uses breadth-first-search for finding augmenting paths to use a minimal amount of edges.


flow85:
Hi, just wanted to ask for a review and/or critique. Thanks

PR #372, opened 2018-01-27
Added Triple DES algorithm

debck:


debck:
added exception handling plz review it

PR #374, opened 2018-01-31
Fix a logical error and set a method to private

GammaBurst101:
In the code of Others/countwords.java, if the user enter no sentence as input then the code was the giving the number of words to be -1 and also the method used for it was public.
So, I thought to correct it.
Thanks for you time 👍


varunu28:
@GammaBurst101 Thanks for the PR

PR #380, opened 2018-02-17
Add a bitwise conversion for DecimalToBinary.

petersonddb:
This adds a (not pure) bitwise approach, which is more efficient than a conventional approach (it does not need any division operation), to convert from decimal to binary. It also creates an interface to ease the choose between conventional and bitwise methods.
I have a pure bitwise approach in another [branch](https://github.com/petersonddb/Java/blob/dec2bin-purebitwise/Conversions/DecimalToBinary.java) but it is less efficient (because the loops and method calls).

### DIFF
--- a/Conversions/DecimalToBinary.java
+++ b/Conversions/DecimalToBinary.java
@@ -6,27 +6,52 @@ import java.util.Scanner;
  * @author Unknown
  *
  */
-class DecimalToBinary
-{
-	/**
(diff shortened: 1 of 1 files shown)


petersonddb:
@varunu28  Updated.

petersonddb:
@varunu28 Done.

petersonddb:
Take a look.

PR #381, opened 2018-02-21
[modified] merge sort complexities

Shyam2609:
I think complexities should be O(n logn) in every case.
But, if not can I know how?


christianbender:
@Shyam2609  Thanks for contribution

PR #383, opened 2018-02-21
Suffix Tree implementation

shaqed:
Hey, I added a SuffixTree implementation inside DataStructures/Trees

### DIFF
--- a/Data Structures/Trees/SuffixTree.java
+++ b/Data Structures/Trees/SuffixTree.java
@@ -1,0 +1,148 @@
+/**
+ * Authors : Shaked Delarea and Linor Hazan
+ * Implementation of the Suffix Tree, tested with the string "senseses$"
+ * */
+
+import java.util.Hashtable;
+
+public class SuffixTree {
+
+    public static void main(String[] args) {
+        SuffixTree suffixTree = new SuffixTree("senseses$");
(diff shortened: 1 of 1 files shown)


Data Structures/Trees/SuffixTree.java
```diff
@@ -0,0 +1,148 @@
...
+ * */
+
+import java.util.Hashtable;
+
+public class SuffixTree {
+
+    public static void main(String[] args) {
+        SuffixTree suffixTree = new SuffixTree("senseses$");
+        // Answer is in root, run the debugger to see its children or follow the print below
+        Node root = suffixTree.getRoot();
```
christianbender: What for a role has this line?

Data Structures/Trees/SuffixTree.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private int numOfSameCharacters(String substring, String edge) {
+        int ans = 0;
+        for (int i = 0; i < Math.min(substring.length(), edge.length()); i++) {
+            if (substring.charAt(i) == edge.charAt(i)) {
+                // Characters match, continue looking
+                ans++;
+            } else {
+                break;
```
christianbender: That line can be a problem. For example if the first characters don't match but all others are equal the method don't counts it. Because it breaks the loop.

PR #387, opened 2018-03-10
Circular Queue

hmangukia:


Data Structures/Queues/circularqueue.java
```diff
@@ -0,0 +1,99 @@
+import java.io.*;
+import java.util.*;
+
+class cqueue
```
christianbender: Please respect name conversions. class names begin with uppercase.

Data Structures/Queues/circularqueue.java
```diff
@@ -0,0 +1,99 @@
...
+{
+	int f=-1,r=-1; //here f is front pointer and r is rear pointer.
+	Scanner x=new Scanner(System.in);
+	int queue[]=new int[5]; //maximum size of queue is 5.
+	public void insert()
+	{
+		int e;
+		if((f==0 && r==4)||((r+1)==f))
```
christianbender: With this condition I can only put 4 elements in the queue. Even though it has a capacity of 5 elements.

Data Structures/Queues/circularqueue.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.*;
+
+class cqueue
+{
+	int f=-1,r=-1; //here f is front pointer and r is rear pointer.
+	Scanner x=new Scanner(System.in);
+	int queue[]=new int[5]; //maximum size of queue is 5.
+	public void insert()
+	{
+		int e;
```
christianbender: The local variable ```e``` is unnecessary. You can assign the user given value direct.

Data Structures/Queues/circularqueue.java
```diff
@@ -0,0 +1,99 @@
...
+			f++;
+		}
+		else if(f==4)
+		{
+			queue[f]='\0';
+			f=	0;
+		}
+	}
+}
+class circularqueue
```
christianbender: The same thing as above. Classname begin with upper case

Data Structures/Queues/circularqueue.java
```diff
@@ -0,0 +1,99 @@
...
+	}
+	public void delet()
+	{
+		if(f==-1 && r==-1)
+			System.out.print("\nThe queue is empty.");
+		else if(queue[f]=='\0' && queue[r]=='\0')
+			System.out.print("\nThe queue is empty.");
+		else if(f>=0)
+		{
+			queue[f]='\0';
```
christianbender: This mark is error-prone. What is if the user give a 0 as input? Use a negative value and inform the user that he can give only positive integers.

PR #397, opened 2018-03-29
QuickSortAlgo.java

khongta0932488598:
Change to List<E> for any type of array.
Change value of pivot to value of middle index of unsorted array.


christianbender:
@khongta0932488598  Why change the type from array to list?

khongta0932488598:
To create any type of array.If something was wrong,let me know

PR #411, opened 2018-04-09
Generalizing & Refactoring

nikitap492:
There was too many differences in code between sort classes, so I decided to generalize it.
I've created utility class which removes some boilerplate and the general interface for the most algorithms
Also, I replaced all algorithms in a new directory, like others java projects
Fixed some bugs in the algorithms and did some perfomance improvments


Sorts/src/sort/BogoSort.java
```diff
@@ -0,0 +1,59 @@
...
+
+
+/**
+ *
+ * @author Podshivalov Nikita (https://github.com/nikitap492)
+ *
+ * @see SortAlgorithm
+ *
+ */
+public class BogoSort implements SortAlgorithm{
```
christianbender: Please add space between the interface name and the curly brace.

Sorts/src/sort/BubbleSort.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ *
+ * @author Varun Upadhyay (https://github.com/varunu28)
+ * @author Podshivalov Nikita (https://github.com/nikitap492)
+ *
+ * @see SortAlgorithm
+ */
+
+class BubbleSort implements SortAlgorithm{
```
christianbender: Same thing as above.

nikitap492:
Done

nikitap492:
Added the new algorithms

PR #412, opened 2018-04-09
Update Caesar.java

khalil2535:
changes :- 
*fix documentation
*methods (encode,decode) can take any number of shifts now and better performed.

### DIFF
--- a/ciphers/Caesar.java
+++ b/ciphers/Caesar.java
@@ -34,12 +34,12 @@ public class Caesar {
 //            int current = message.charAt(i); //using char to shift characters because ascii is in-order latin alphabet
             char current = message.charAt(i); // Java law : char + int = char
 
-            if (current >= 'A' && current <= 'Z') {
+            if (IsCapitalLatinLetter(current)) {
 
                 current += shift;
                 encoded += (char) (current > 'Z' ? current - 26 : current); // 26 = number of latin letters
 
-            } else if (current >= 'a' && current <= 'z') {
+            } else if (IsSmallLatinLetter(current)) {
 
                 current += shift;
                 encoded += (char) (current > 'z' ? current - 26 : current); // 26 = number of latin letters
--- a/ciphers/Caesar.java
+++ b/ciphers/Caesar.java
@@ -1,117 +1,110 @@
-/**
-Author : FAHRI YARDIMCI
+package ciphers;
 
-A Java implementation of Caesar Cipher.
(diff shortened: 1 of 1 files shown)


ciphers/Caesar.java
```diff
@@ -1,117 +1,110 @@
...
+            shift -= 26;
+        }
+
+        final int length = message.length();
+        for (int i = 0; i < length; i++) {
+
+//            int current = message.charAt(i); //using char to shift characters because ascii is in-order latin alphabet
+            char current = message.charAt(i); // Java law : char + int = char
+
+            if (current >= 'A' && current <= 'Z') {
```
christianbender: Instead of this complex ```if``` use ```Character.isLetter(c)``` see [here](https://www.tutorialspoint.com/java/character_isletter.htm)

ciphers/Caesar.java
```diff
@@ -1,117 +1,110 @@
...
+        String decoded = "";
+
+        while (shift >= 26) { // 26 = number of latin letters
+            shift -= 26;
+        }
+
+        final int length = encryptedMessage.length();
+        for (int i = 0; i < length; i++) {
+            char current = encryptedMessage.charAt(i);
+            if (current >= 'A' && current <= 'Z') {
```
christianbender: Same thing as above.

ciphers/Caesar.java
```diff
@@ -1,117 +1,110 @@
...
+    public static void main(String[] args) {
+        Scanner input = new Scanner(System.in);
+        System.out.println("Please enter the message (Latin Alphabet)");
+        String message = input.nextLine();
+        System.out.println(message);
+        System.out.println("Please enter the shift number");
+        int shift = input.nextInt() % 26;
+        System.out.println("(E)ncode or (D)ecode ?");
+        char choice = input.next().charAt(0);
+        if (choice == 'E' || choice == 'e') {
```
christianbender: Here you can simpler use a ```switch```

ciphers/Caesar.java
khalil2535: Character.isLetter(c) method return true if the input was 'ф' or 'أ' or '褥' while we don't shift them in this method , it's just for shifting Latin letters .

ciphers/Caesar.java
khalil2535: thanks for advice here :+1:

PR #417, opened 2018-04-13
Added a iterative version of a ternary search algorithm

nikitap492:
Added a iterative version of ternary search algorithm and some minor changes in documentation

### DIFF
--- a/Searches/src/search/IterativeTernarySearch.java
+++ b/Searches/src/search/IterativeTernarySearch.java
@@ -1,0 +1,76 @@
+package search;
+
+import java.util.Arrays;
+import java.util.Random;
+import java.util.stream.Stream;
+
+import static java.lang.String.format;
+
+/**
+ *
+ * A iterative version of a ternary search algorithm
+ * This is better way to implement the ternary search, because a recursive version adds some overhead to a stack.
+ * But in java the compile can transform the recursive version to iterative implicitly,
+ * so there are no much differences between these two algorithms
+ *
+ * Worst-case performance	Θ(log3(N))
+ * Best-case performance	O(1)
+ * Average performance	Θ(log3(N))
+ * Worst-case space complexity	O(1)
+ *
+ *
+ * @author Podshivalov Nikita (https://github.com/nikitap492)
+ * @since 2018-04-13
+ *
+ * @see SearchAlgorithm
+ * @see TernarySearch
+ *
+ */
+
+public class IterativeTernarySearch implements SearchAlgorithm {
+
+
+    @Override
+    public <T extends Comparable<T>> int find(T[] array, T key) {
+        int left = 0;
+        int right = array.length - 1;
+        while (true) {
+            int leftCmp =  array[left].compareTo(key);
+            int rightCmp =  array[right].compareTo(key);
+            if (leftCmp == 0) return left;
+            if (rightCmp == 0) return right;
+
+            int leftThird = left + (right - left) / 3;
+            int rightThird = right - (right - left) /3;
+
+            if (array[leftThird].compareTo(key) <= 0) left = leftThird;
+            else right = rightThird;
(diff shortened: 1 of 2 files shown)


Searches/src/search/IterativeTernarySearch.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+public class IterativeTernarySearch implements SearchAlgorithm {
+
+
+    @Override
+    public <T extends Comparable<T>> int find(T[] array, T key) {
+        int left = 0;
+        int right = array.length - 1;
+        while (true) {
```
christianbender: This while loop is dangerous because if the key not contain in the array, we end up with a endless loop. 
```java
Integer shouldBeFound = 1;
```
End up with a endless loop. 

I have test your code. Change your code:
```java
while (!(left == 0 && right == 2)) {
            int leftCmp =  array[left].compareTo(key);
            int rightCmp =  array[right].compareTo(key);
            if (leftCmp == 0) return left;
            if (rightCmp == 0) return right;

            int leftThird = left + (right - left) / 3;
            int rightThird = right - (right - left) /3;
            if (array[leftThird].compareTo(key) <= 0) left = leftThird;
            else right = rightThird;
}
return -1;
```
The termination condition is ```left = 0``` and ```right=2```. You must catch the return value ```-1``` !

Searches/src/search/IterativeTernarySearch.java
```diff
@@ -0,0 +1,76 @@
...
+            int leftCmp =  array[left].compareTo(key);
+            int rightCmp =  array[right].compareTo(key);
+            if (leftCmp == 0) return left;
+            if (rightCmp == 0) return right;
+
+            int leftThird = left + (right - left) / 3;
+            int rightThird = right - (right - left) /3;
+
+            if (array[leftThird].compareTo(key) <= 0) left = leftThird;
```
christianbender: Please use curly braces for the ```if-else```

Searches/src/search/IterativeTernarySearch.java
```diff
@@ -0,0 +1,76 @@
...
+
+    }
+
+
+    public static void main(String[] args) {
+        //just generate data
+        Random r = new Random();
+        int size = 100;
+        int maxElement = 100000;
+        Integer[] integers = Stream.generate(() -> r.nextInt(maxElement)).limit(size).sorted().toArray(Integer[]::new);
```
christianbender: Please break up the lines. max. 80 columns per line.

Searches/src/search/IterativeTernarySearch.java
```diff
@@ -0,0 +1,76 @@
...
+        Integer shouldBeFound = integers[r.nextInt(size - 1)];
+
+        IterativeTernarySearch search = new IterativeTernarySearch();
+        int atIndex = search.find(integers, shouldBeFound);
+
+        System.out.println(format("Should be found: %d. Found %d at index %d. An array length %d"
+                , shouldBeFound, integers[atIndex], atIndex, size));
+
+        int toCheck = Arrays.binarySearch(integers, shouldBeFound);
+        System.out.println(format("Found by system method at an index: %d. Is equal: %b", toCheck, toCheck == atIndex));
```
christianbender: Break up the line, too.

Searches/src/search/IterativeTernarySearch.java
nikitap492: Thanks, you're right. I didnt notice that at all ) . The loop is infinite if a key doesn't exist in an array. But your condition
`while (!(left == 0 && right == 2))` isn't always true. For instance, if the key is bigger than max element of the array then the condition should be `left == array.length - 3 && right == array.length - 1` . I improved the algorithm, so new termination condition is just `right > left`

PR #422, opened 2018-04-20
Addition of tests for lists

dragid10:
I deleted the duplicate Data Structures package because it was unnecessary. I fixed most classes not having a package by adding a package to the top of the file. I also fixed some minor syntax errors in a few files


dragid10:
Added testing documentation as well as tests themselves to code

PR #429, opened 2018-04-25
 the second array may have residual elements.

Hero22:
This is my first time to create pull request, if I did something wrong, I'll apologize.
The next is why a create pull request.
If "i" reaches "mid" , but "j" doesn't reach "right", the left of "j" should also be add to arr. 
Thanks.


Hero22:
Thanks for replying! I'll think about a better commit message~

PR #485, opened 2018-08-15
Added 4 Dynamic Programming Algorithms

MayV:
1. Maximum size square sub matrix with all 1's in it.
2. Maximum sum sub array by removing at most 1 element.
3. Minimum number of edits required to convert string1 to string2.
4. Minimum difference between sum of two partitions.


MayV:
@varunu28 Updated the files according to the suggestions mentioned above.
Thanks @gabru-md for helping me out in my first pull request.

MayV:
@varunu28 please provide any update which is to be rectified in this pr.

PR #487, opened 2018-08-21
Added bucket sort

rajatsharma2393:
Have added another sorting algorithm, bucket sort.


shellhub:
See https://github.com/TheAlgorithms/Java/blob/master/Sorts/BucketSort.java

PR #500, opened 2018-08-29
add Test for InsertionSort

lq920320:


lq920320:
@varunu28  Only this test or including other tests have print statements in my code?

varunu28:
@lq920320 Remove it from other ones too. I missed it in earlier reviews. We don't need a print statement in the tests. If it fails then Junit assertion generates enough logs for clarification

PR #511, opened 2018-09-03
Added SimplexNoise

RalleYTN:
More specifically: height map generation using Simplex Noise


RalleYTN:
Required changes were made.

PR #519, opened 2018-09-18
Simple refactoring for clarity and conciseness

tristangreeno:


varunu28:
@tristangreeno  We have yet not decided to move towards Java10 standards. So putting a ```var``` for datatypes would make the code confusing for beginners. If you want to make changes, then you can put an effort to make the code more functional style using Java 8 changes

PR #523, opened 2018-10-01
Added dynamic fibonacci algorithm

CraxyTM:
Due to the conversation with @varunu28 I now made a fibonacci algorithm that (hopefully) is dynamic programming

### DIFF
--- a/src/main/java/com/dynamic/Fibonacci.java
+++ b/src/main/java/com/dynamic/Fibonacci.java
@@ -1,0 +1,22 @@
+package src.main.java.com.dynamic;
+
+/**
+ * The Fibonacci sequence is a sequence of numbers in which each number after the first two is the sum of the two preceding ones.
+ * 1, 1, 2, 3, 5, 8, 13, ...
+ */
+public class Fibonacci {
+
+    /**
+     * A recursive method that calculates and returns the nth fibonacci number.
+     *
+     * @return the nth fibonacci number.
+     */
+    public int fibonacci(int n) {
+        if (n <= 1) {
+            return n;
+        }
+
+        return fibonacci(n - 1) + fibonacci(n - 2);
+    }
+
+}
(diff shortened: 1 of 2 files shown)


src/main/java/com/dynamic/Fibonacci.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.HashMap;
+
+/**
+ * The Fibonacci sequence is a sequence of numbers in which each number after the first two is the sum of the two preceding ones.
+ * 1, 1, 2, 3, 5, 8, 13, ...
+ */
+public class Fibonacci {
+
+    private static HashMap<Integer, Integer> fibonacciCache = new HashMap<>();
```
yejjuritesh, line 11: Instead of a HashMap, we can do this with constant space by using two variables.

src/main/java/com/dynamic/Fibonacci.java
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * A recursive method that calculates and returns the nth fibonacci number.
+     *
+     * @return the nth fibonacci number.
+     */
+    public int fibonacci(int n) {
```
yejjuritesh, line 18: Instead of a recursive approach or top-down dynamic approach, A bottom-up approach will do the trick.

src/main/java/com/dynamic/Fibonacci.java
yejjuritesh: Also, An Array or ArrayList can be used instead of HashMap. Because the index of array can be assumed as the position of fibbonaci number :
index of Array- fibbonaci number
0 - 1
1 - 1
2 - 2
3 - 3
4 - 5
5 - 8
So, nth fibonacci number will be present at nth index of array.

ayaankhan98:

PR #529, opened 2018-10-02
Re-structure project to use Maven

camilne:
This PR re-structures the project to use a standard build tool.

Why?
1. It standardizes dependency versions. (Also makes barrier to contribution lower because no more manual dependency management)
2. It standardizes project structure (source files, test files, etc.)
3. It is not IDE specific, so people with different IDEs can import the pom file and get the correct project setup in their IDE.


yanglbme:
#795

PR #533, opened 2018-10-02
Update GCD function

SUSHANT12PRABHU:


adugamayuba:
	//take input    
      Scanner scn = new Scanner(System.in);
       System.out.println("please in put the value of N ");
      int N = scn.nextInt(); //N is the input number

 System.out.println("please in put the value of P ");
      int P = scn.nextInt(); //P is precision value for eg - P is 3 in 2.564 and 5 in 3.80870.
      
      System.out.println(squareRoot(N, P));

varunu28:
@SUSHANT12PRABHU Please don't raise a PR from development to master

PR #549, opened 2018-10-04
Fixes #546

### LINKED ISSUE #546: Compiling error in SinglyLinkedList.java
In https://github.com/TheAlgorithms/Java/blob/77259b73b18a1eede7cf0e3fe75aeb580e9faf24/DataStructures/Lists/SinglyLinkedList.java#L48

The user is calling for empty constructor but there's none provided. 
Results in 

```java
ss.java:48: error: constructor Node in class Node cannot be applied to given types;
        Node newNode = new Node();
                       ^
  required: int
  found: no arguments
  reason: actual and formal argument lists differ in length
ss.java:49: error: cannot find symbol
        newNode.data = data;
               ^
  symbol:   variable data
  location: variable newNode of type Node
```
we don't really need an empty constructor simple solution would be to use the parameterzed constructor already built.

pablohildo:
Fixes #546 
Instantiated newNode with the proper constructor

### DIFF
--- a/DataStructures/Lists/SinglyLinkedList.java
+++ b/DataStructures/Lists/SinglyLinkedList.java
@@ -45,8 +45,7 @@ class SinglyLinkedList{
 
     Node InsertNth(Node head, int data, int position) {
         
-        Node newNode = new Node();
-        newNode.data = data;
+        Node newNode = new Node(data);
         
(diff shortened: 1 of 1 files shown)


shellhub:
closed because of conflicts

PR #552, opened 2018-10-06
Updated singly linked list

taher27:
1. default constructor was missing. 
   so it is now created at line 143.
2. error: use data instead of value.
 that change is also updated now.


taher27:
Node InsertNth(Node head, int data, int position) {
        
        Node newNode = new Node();
        newNode.data = data;             // old code
        newNode.value = data;		// updated code
        
        if (position == 0) {
            newNode.next = head;

public Node(int valuein){
		value = valuein;
	}

taher27:
and Add default constructor

Default constructor
	public Node(){
		value = 0;    
	}

DataStructures/Lists/SinglyLinkedList.java
```diff
@@ -46,7 +48,7 @@ public void insertHead(int x){
     Node InsertNth(Node head, int data, int position) {
         
         Node newNode = new Node();
-        newNode.data = data;
+        newNode.value = data;		// take value
```
taher27: Updated code from variable name ( data to value)

DataStructures/Lists/SinglyLinkedList.java
```diff
@@ -141,6 +143,13 @@ public Node(int valuein){
 		value = valuein;
 	}
 
+
+
+// Default constructor
+	public Node(){
```
taher27: add 1 default constructor.

shellhub:
Closed because of Conflicting

PR #558, opened 2018-10-08
Added an explanation to SkylineProblem

rmakynen:
There was no explanation for the SkylineProblem  algorithm, which caused the user to have to google. Now there user gets a quick overview of what the algorithm does.


Others/Dijkstra.java
```diff
@@ -0,0 +1,178 @@
...
+	public Vertex previous = null;
+	public final Map<Vertex, Integer> neighbours = new HashMap<>();
+ 
+	public Vertex(String name)
+	{
+		this.name = name;
+	}
+ 
+	private void printPath()
+	{
```
yanglbme: The opening brace ` {` and method declarations should be on the same line. Most of the standard source code for the Java API is written in this style. 
eg.
```java
public Vertex(String name) {
    this.name = name;
}
```

yanglbme: Also for the blocks inside a function.
```java
if (this == this.previous) {
    System.out.printf("%s", this.name);
}

PR #570, opened 2018-10-10
FINDING GREATEST OF 10 NUMBERS ENTERED

logan0007:


arfinsunny:
package sunny;
import java.util.Scanner;
public class FindGreatest {
        Scanner input=new Scanner(System.in);
       int [] arr=new int[10];
       for (int i = 0; i < 10; i++) {
           System.out.println("Enter Number "+(i+1)+"=");
            arr[i]=input.nextInt();}
        int max=arr[0];
        for (int i = 0; i < 10; i++) 
            if(arr[i]>max)
                max=arr[i];
        System.out.println("Greatest Number = "+max);
    }
}

PR #573, opened 2018-10-10
Updated Factorial.java for large input

sunilnitdgp:
The code will work if the number is too large.


shellhub:
Please rewrite a single file named FactorialLarge.java to Maths package.

PR #588, opened 2018-10-17
Create RecursiveLinearSearch

MatheusMuriel:


havanagrawal:
Few comments:

 - [ ] File extension needs to be `.java`
 - [ ] The algorithm doesn't look very efficient; for an array of size n, it will make n copies, effectively making it an O(n^2) algorithm for linear search.
 - [ ] Indentation needs to be fixed
 - [ ] Missing test cases

PR #592, opened 2018-10-22
Update README.md

m-rahimy:
The folder "Data Structures" was renamed to "DataStructures" but the links was not. fixed that.


yanglbme:
👍Thx

PR #601, opened 2018-10-27
Need to add a precise

SinanKamil:
We need to close the code with a precise


SinanKamil:
This will make the code work because it is not closed with a precise.

PR #636, opened 2018-11-05
Added interpolation search

LesliePinto89:
Added interpolation search and JUnit test.

### DIFF
--- a/src/main/java/com/search/InterpolationSearch.java
+++ b/src/main/java/com/search/InterpolationSearch.java
@@ -51,12 +51,12 @@ public class InterpolationSearch {
 	 * @param highIndex The last and largest element in the sorted array
 	 * @return The array's current retrieved index position
 	 */
-	private <T> int getPostion(T arr[], T key, int lowIndex, int highIndex) {	
+	private <T> int getPostion(T arr[], T key, int lowIndex, int highIndex) {			
 		String startValueString = arr[lowIndex].toString(); //First convert <T> array element to String
-		int startValueInt = Integer.parseInt(startValueString); //Convert String to int to computate later
+		int startValueInt = Integer.parseInt(startValueString); //Convert String to int to later computate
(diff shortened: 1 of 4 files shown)


src/main/java/com/search/InterpolationSearch.java
```diff
@@ -0,0 +1,54 @@
...
+     * @param array The array to make the interpolation search
+     * @param key The element to be found 
+     * @param start The first and smallest element in the sorted array
+     * @param end The last and largest element in the sorted array
+     * @return The index location of the key
+	 */
+	private <T extends Comparable<T>> int search(T arr[], T key, int start, int end) {
+
+		while (start <= end && key.compareTo(arr[start]) > 0
+				|| key.compareTo(arr[start]) == 0 && key.compareTo(arr[end]) < 0 || key.compareTo(arr[end]) == 0) {
```
lodenrogue: Can we put these conditions in one or two methods with easy to read names?

src/main/java/com/search/InterpolationSearch.java
```diff
@@ -0,0 +1,54 @@
...
+
+			String startValueString = arr[start].toString();
+			int startValueInt = Integer.parseInt(startValueString);
+			String endValueString = arr[end].toString();
+			int endValueInt = Integer.parseInt(endValueString);
+
+			String keyValueString = key.toString();
+			int keyValueInt = Integer.parseInt(keyValueString);
+
+			int position = start + (((end - start) / (endValueInt - startValueInt) * (keyValueInt - startValueInt)));
```
lodenrogue: You can probably put everything from String startValueString up until this line in a method called getPosition(...). That way if we ever need to change the logic for getting the position we don't have to disturb the rest of this code.

LesliePinto89:
I have made some improvements, is this revision better?

PR #643, opened 2018-11-12
Test Codes and Refactoring of SinglyLinkedList

ales-blaze:
I am a little bit worried about the test cases!
So waiting for your suggestions for that.

### DIFF
--- a/DataStructures/Lists/SinglyLinkedList.java
+++ b/DataStructures/Lists/SinglyLinkedList.java
@@ -12,69 +12,129 @@
  * @author Unknown
  *
  */
-class SinglyLinkedList{
-	/**Head refered to the front of the list */
+class SinglyLinkedList {
+	public Node getHead() {
+		return head;
+	}
+
+	/**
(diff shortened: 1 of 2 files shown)


varunu28:
@ales-blaze Thanks. Would look into it later during the day

varunu28:
@ales-blaze We are adding tests to Development branch. Therefore you have to open a PR against Development. I am closing this PR. You can raise another PR

ales-blaze:
Ok @varunu28, but what is your opinion about the test?
Are they good or bad ?

varunu28:
@ales-blaze  You would have to follow a proper naming convention. Please have a look at this [thread](https://stackoverflow.com/questions/155436/unit-test-naming-best-practices)

Also, the tests should be pretty generic. A test ```headMustNullAfterDeletion``` is not generic because the head would not be null after every deletion if the list contains more than 1 nodes. 

Go ahead and raise the PR and I would review it thoroughly.

Issue #644, opened 2018-11-13
Java/DataStructures/Trees/BinaryTree.java has error with put method


Keenan-Ker51:
The binary tree class seems to have an error on line 93, when trying to put more than one int into the tree it throws null pointer exception. I think the reason for this is that it calls the find function, using the value that you would want to add to the tree, so of course the find method will only return a null value, unless you happen to be adding the same value twice.

varunu28:
I think this can be solved by adding a null check for the parent before proceeding to the comparison. 
@Keenan-Ker51  Can you help with a PR?

ales-blaze:
I am interested in doing so, wait for your allowance!

varunu28:
@ales-blaze 
You don't need to wait. You can go ahead and open a PR

ales-blaze:
Wait, isn't there any way through which a particular person gets the authority to perform a particular task,
To avoid work collisions.

varunu28:
@ales-blaze You can just comment in the issue that you are working on it. This would avoid any conflicts.

marcgood96:
Should this issue be closed? It seems like themanchangechina fixed it.

varunu28:
Yes. Closing it

PR #646, opened 2018-11-17
fix the BinaryTree put method, returns the parent when the key isn't exist.

themanchangechina:
I think the Tree class you writed is actually a BinarySearchTree. 
#644


varunu28:
@themanchangechina Can you please squash the commits into one so that it is easier to review the changes

PR #654, opened 2018-12-06
Translate README to Korean Language

SunggyuLee:
Readme with korean language will help Koreans to study


varunu28:
@SunggyuLee You don't need to change the existing README as people are following the repo in the English language. We can open a new branch where we can do the translation.

PR #655, opened 2018-12-10
Update Stacks.java

tosukriti:
Update Stacks.java to fix bug in resize() method


tosukriti:
Thank you for the approval.

PR #666, opened 2018-12-24
Add SortUtils before less and print methods methods.

fthgul:
Hi,
I think  SortUtils adding before less and print methods has been skipped.


yanglbme:
@fthgul See line3 👇
```java
 import static Sorts.SortUtils.*;
```

PR #668, opened 2018-12-29
Issue compiling error in SinglyLinkedList.java fixed

Goku1999:
The problem in compiling this particular problem is because we are calling default constructor when it is not even present in the Node class


varunu28:
@Goku1999  Please don't raise a PR for merging development into master

pandeyarun709:
updating for branch

PR #669, opened 2019-01-02
Added Counting Sort and Cycle Sort

ani03sha:
1. Added algorithm class CountingSort and its corresponding test class CountingSortTest
2. Added algorithm class CycleSort and its corresponding test class CycleSortTest

### DIFF
--- a/src/main/java/src/main/java/com/sorts/CountingSort.java
+++ b/src/main/java/src/main/java/com/sorts/CountingSort.java
@@ -16,13 +16,18 @@ public class CountingSort {
             max = Math.max(max, arr[i]);
         }
 
+        // Finding the minimum element in the array
+        int min = arr[0];
+        for (int i = 1; i < arr.length; i++) {
+            min = Math.min(min, arr[i]);
+        }
+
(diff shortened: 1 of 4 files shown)


src/main/java/src/main/java/com/sorts/CountingSort.java
```diff
@@ -0,0 +1,52 @@
...
+            max = Math.max(max, arr[i]);
+        }
+
+        // Creating the count array - This method will store the count of each element in the unsorted array
+        int[] count = new int[max + 1];
+
+        // This loop will store the count of each element in the array
+        for (int i = 0; i < max; i++) {
+
+            count[arr[i]]++;
```
yanglbme: what will happen if arr[i] less than zero ?

ani03sha: @yanglbme good find! I have fixed the issue by adding the offset for the negative values.

yanglbme: @ani03sha 👍

PR #694, opened 2019-01-27
conversion binary to hexadecimal

arodriguez33:
conversion binary to hexadecimal

### DIFF
--- a/src/main/java/com/conversions/BinaryToHexadecimal.java
+++ b/src/main/java/com/conversions/BinaryToHexadecimal.java
@@ -2,43 +2,51 @@ package src.main.java.com.conversions;
 
 import java.math.BigInteger;
 import java.util.HashMap;
+import java.util.Map;
+
 
 public class BinaryToHexadecimal {
+
+    /**
+     * hm to store hexadecimal codes for binary numbers
+     * within the range: 0000 to 1111 i.e. for decimal numbers 0 to 15
+     */
+   private static Map<Integer, String> hmHexadecimal = new HashMap<>(16);
+
+    static {
+        int i;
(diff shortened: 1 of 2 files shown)


src/main/java/com/conversions/BinaryToHexadecimal.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * This method converts a binary number to
+     * a hexadecimal number.
+     *
+     * @param binary The binary number
+     * @return The hexadecimal number
+     */
+
+    public String binToHex(int binary) {
```
yanglbme: The binary number may have 32bits or 64bits, so it's  unsuitable to use `int` type.

arodriguez33: Resolved

src/test/java/com/conversions/BinaryToHexadecimalTest.java
```diff
@@ -0,0 +1,15 @@
...
+import src.main.java.com.conversions.BinaryToHexadecimal;
+import org.junit.Assert;
+
+public class BinaryToHexadecimalTest {
+
+    @Test
+    public void testBinaryToHexadecimal(){
+        BinaryToHexadecimal binaryToHexadecimal = new BinaryToHexadecimal();
+        Assert.assertEquals("Incorrect Conversion", "2A", binaryToHexadecimal.binToHex(101010));
+        Assert.assertEquals("Incorrect Conversion", "24", binaryToHexadecimal.binToHex(100100));
```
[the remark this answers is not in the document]
arodriguez33: Hi, i had to use BigInteger to resolve that case you commented.

Issue #708, opened 2019-02-20
Fibonacci sequence should start with [0, 1] and not [1, 1]


HeikoAlexanderWeber:
# Summary
The Fibonacci sequence is defined by following statement:
_F(n) = F(n-1) + F(n-2) with F(0) = 0 and F(1) = 1_. For reference, please see [The online encyclopedia of integer sequences](https://oeis.org/A000045) as primary reference, leading to other references with more detailed description.
(Older) books sometimes omit the preceding _0_, but this representation is deprecated.

### fibMemo
When calling the function [Fibonacci.fibMemo](https://github.com/TheAlgorithms/Java/blob/e4a9d38ee39767711784f1d5c21029ce2341846f/Dynamic%20Programming/Fibonacci.java#L30), the result of _F(0)_ = _1_. This is incorrect since _F(0) = 0_. Furthermore this leads to the following sequence for the first few integers: _[1, 1, 1, 2, 3, 5, ...]_ for _input=[0, 1, 2, 3, 4, 5, ...]_.
### fibBotUp
When calling the function [Fibonacci.fibBotUp](https://github.com/TheAlgorithms/Java/blob/e4a9d38ee39767711784f1d5c21029ce2341846f/Dynamic%20Programming/Fibonacci.java#L53), the program throws an exception when trying to get the value for _F(0)_ from the map. Therefore _F(0)_ is not supported by this implementation.

## Task
Fix both implementation(s) to
a) Return correct values for _F(x)_
b) Do not throw exceptions
Acceptance criteria is met when both functions return _[0, 1, 1, 2, 3, 5]_ for _input=[0, 1, 2, 3, 4, 5]_.

ales-blaze:
i will work on it!

HeikoAlexanderWeber:
@ales-blaze I have already fixed the bug because I was really bored. See the attached PR.

wangyuan2019:
I will try it

PR #709, opened 2019-02-26
#708 (bugfix for fibonacci sequence function)

HeikoAlexanderWeber:
o fixed function "fibMemo" according to acceptance criteria in #708
o fixed function "fibBotUp" according to acceptance criteria in #708

### DIFF
--- a/Dynamic Programming/Fibonacci.java
+++ b/Dynamic Programming/Fibonacci.java
@@ -17,8 +17,9 @@ public class Fibonacci {
         BufferedReader br = new BufferedReader(new InputStreamReader(System.in));
         int n = Integer.parseInt(br.readLine());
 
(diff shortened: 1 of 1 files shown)


Dynamic Programming/Fibonacci.java
```diff
@@ -54,10 +54,10 @@ private static int fibBotUp(int n) {
 
         Map<Integer, Integer> fib = new HashMap<>();
 
-        for (int i = 1; i < n + 1; i++) {
+        for (int i = 0; i <= n; i++) {
```
HeikoAlexanderWeber: Since the pattern 'i <= n' was also used in 'fibMemo', I decided that using this pattern instead of 'i < n + 1' is more suitable here. This change is more a style fix than a logic fix.

PR #710, opened 2019-02-28
Adding interface for sorting to use polymorphism features.

AnkitAtBrillio:


AnkitAtBrillio:
This will make sure that all the sorting algorithms classes are overriding the same method with same signatures.
This could be extended to all the classes.

PR #725, opened 2019-03-24
Add binary tree implementation with test

RicardoGuevara:
Binary tree implemented for general type of comparable data

### DIFF
--- a/src/main/java/com/dataStructures/BinaryTree.java
+++ b/src/main/java/com/dataStructures/BinaryTree.java
@@ -1,0 +1,125 @@
+package dataStructures;
+
+/**
+ * Binary tree for general value type, without redundancy
+ * @author RICARDO
+ * @param <T> root data
+ */
+public class BinaryTree<T extends Comparable> {
+    private final T data;
+    private BinaryTree  right,  // the upper binary tree
+                        left;   // the lower binary tree
+
+    public BinaryTree(T data) {
+        this.data = data;
+    }
+
+    @Override
+    public String toString(){
+        return this.data.toString();
+    }
+    
+    /**
+     * inserts a new value in it's correspondant place
+     * @param newDataValue value of the new banary tree to add on this tree
+     */
+    public void insert(T newDataValue){
+        this.insert(new BinaryTree(newDataValue));
+    }
+    
+    /**
+     * inserts a new binary tree in it's correspondant place
+     * @param newData new value to add on this tree 
+     */
+    public void insert(BinaryTree newData){
+        
+        int cpr = newData.data.compareTo(this.data); //new value comparission respect to actual value
+        
+        if (cpr < 0)
+            if (this.left == null)
+                this.setLeft(newData);
+            else
+                this.left.insert(newData);
+        else if (cpr > 0)
+            if (this.right == null)
+                this.setRight(newData);
+            else
+                this.right.insert(newData);
+        else
+            System.out.println("Redundant value, not added");
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/dataStructures/BinaryTree.java
```diff
@@ -0,0 +1,125 @@
...
+        return this;
+    }
+    
+    /**
+     * Checks if the data value exist in the tree
+     * @param data data to be searched
+     * @return true if this tree contains the data value, false if not.
+     */
+    public boolean contains(T data){
+        return !(this.search(data) == null);
```
yanglbme: Use `this.search(data) != null` to make code more readable.

src/main/java/com/dataStructures/BinaryTree.java
```diff
@@ -0,0 +1,125 @@
...
+     * uses recursive black magic to print this tree in console
+     * @param tabCounter prev tabs
+     */
+    private void print(int tabCounter){
+        for (int i = 0; i < tabCounter; i++)
+            System.out.print("\t");
+        
+        System.out.println(this);
+        
+        if (this.left!=null)
```
yanglbme: Format your code, like this ↓
```java
if (this.left != null)
    this.left.print(tabCounter + 1);  // it can't be ++ , pls don't change it
if (this.right != null)
    this.right.print(tabCounter + 1);  // it can't be ++ , pls don't change it
```

RicardoGuevara: ok, thanks!

src/main/java/com/dataStructures/BinaryTree.java
RicardoGuevara: thanks i just dindn't see it

varunu28:
@yanglbme Why was this merged to the development?

There is a signature for the author in the code which we decided not to have and is mentioned on README.

It has improper coding standards such as 
```
boolean result = instance.contains(2)&&instance.contains(11);
String proof = instance.getLeft().toString()+instance.toString()+instance.getRight().toString();
```

**Please make sure that you are reviewing the PR before merging.** 

I will be reverting a few PRs which were merged incorrectly on the Developement branch. We need to keep this branch as clean as possible.

yanglbme:
Hi @varunu28 , I think there is no problem adding a signature to the code because it conforms to the Java coding specification. And you can just make some changes to improve the code.

varunu28:
> Hi @varunu28 , I think there is no problem adding a signature to the code because it conforms to the Java coding specification. And you can just make some changes to improve the code.

The problem arises when someone modifies an existing code and adds another signature on top of the existing signature making the file more verbose.

And coming to coding standards, I am still not sure why it was overlooked and how the changes got approved.

Complete code file has uneven distribution of braces, improper spaces and hacky in-line comments spread throughout the code.

yanglbme:
@varunu28 ╮(╯-╰)╭ All right, it's up to you now.

yanglbme:
My Fault, I didn't review the code carefully. So sorry... @varunu28

varunu28:
@yanglbme I am not sure about what idea did you got from this conversation but my only intention is that we have clean and tested code in the repository. 

Members of the repo have already opened issues about code being buggy. The code in binary search on master was also pointed to me by a user through email.

My only request is to have the code tested on your local before approving the changes. We plan to build a test pipeline for having automated cases in future but till then we have to be extra careful about the quality of code we are adding to the repo.

yanglbme:
@varunu28 OK, I got it.

PR #731, opened 2019-04-08
Added FastPower

DDullahan:
Added FastPower

We may calculate power with loops, but what if the index is too large ?

FastPower aims to calculate quickly in this circumstances with time complexity O(log k), where k is the index.

### DIFF
--- a/src/main/java/com/Others/FastPower.java
+++ b/src/main/java/com/Others/FastPower.java
@@ -1,0 +1,37 @@
+package src.main.java.com.Others;
+
+import java.math.BigInteger;
+
+/**
+ * We may calculate power with loops, but what if the index is too large ?
+ * FastPower aims to calculate quickly in this circumstances with time complexity O(log k),
+ * where k is the index.
+ *
+ * @author DDullahan
+ */
+public class FastPower {
+    public static BigInteger calculate(BigInteger n, BigInteger k, BigInteger mod) {
+        BigInteger ans = BigInteger.ONE;
+        while (!k.equals(BigInteger.ZERO)) {
+            int odd = k.mod(new BigInteger("2")).compareTo(BigInteger.ZERO);
+            if(odd == 1){
+                ans = ans.multiply(n).mod(mod);
+            }
+            k = k.divide(new BigInteger("2"));
+            n = n.multiply(n).mod(mod);
+        }
+        return ans.mod(mod);
+    }
+
+    public static long calculate(long n, long k, long mod) {
+        long ans = 1;
+        while (k != 0) {
+            if (k % 2 == 1) {
+                ans = (ans * n) % mod;
+            }
+            k >>= 1;
+            n = (n * n) % mod;
+        }
+        return ans % mod;
+    }
+}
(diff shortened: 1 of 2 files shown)


src/main/java/com/Others/FastPowerTest.java
```diff
@@ -0,0 +1,40 @@
...
+import org.junit.*;
+
+import java.math.BigInteger;
+
+import static org.junit.Assert.*;
+
+public class FastPowerTest {
+
+    @Test
+    public void test() {
```
havanagrawal: This should have `assert*` statements, not `println`s.

src/main/java/com/Others/FastPower.java
```diff
@@ -0,0 +1,37 @@
...
+ * where k is the index.
+ *
+ * @author DDullahan
+ */
+public class FastPower {
+    public static BigInteger calculate(BigInteger n, BigInteger k, BigInteger mod) {
+        BigInteger ans = BigInteger.ONE;
+        while (!k.equals(BigInteger.ZERO)) {
+            int odd = k.mod(new BigInteger("2")).compareTo(BigInteger.ZERO);
+            if(odd == 1){
```
havanagrawal: The result of [`compareTo`](https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#modPow(java.math.BigInteger,%20java.math.BigInteger)) should only be compared with 0, there is no guarantee that it will be 1, only that it may be negative, 0 or positive.

src/main/java/com/Others/FastPower.java
```diff
@@ -0,0 +1,37 @@
...
+                ans = ans.multiply(n).mod(mod);
+            }
+            k = k.divide(new BigInteger("2"));
```
havanagrawal: Instead of using `new BigInteger("2")` in a loop, we could declare it as a constant outside:

```java
public static final BigInteger TWO = new BigInteger("2");
```

DDullahan:
Thanks for your review @havanagrawal , I have fixed these problems.

src/main/java/com/Others/FastPower.java
```diff
@@ -0,0 +1,37 @@
+package src.main.java.com.Others;
+
+import java.math.BigInteger;
+
+/**
+ * We may calculate power with loops, but what if the index is too large ?
+ * FastPower aims to calculate quickly in this circumstances with time complexity O(log k),
+ * where k is the index.
+ *
+ * @author DDullahan
```
yanglbme: Please do not add a signature `@author xxx` inside the code. See [Contribution Guidelines](https://github.com/TheAlgorithms/Java/tree/Development#contribution-guidelines) for details.

src/main/java/com/Others/FastPowerTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.*;
+
+import java.math.BigInteger;
+
+import static org.junit.Assert.*;
+
+public class FastPowerTest {
+
+    void testLong(long n, long k, long m){
+        long result = FastPower.calculate(n,k,m);
```
yanglbme: Add one single space after `,` , e.g.
```java
long result = FastPower.calculate(n, k, m);
```

src/main/java/com/Others/FastPowerTest.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    void testBigInteger(BigInteger n, BigInteger k, BigInteger m){
+        BigInteger result = FastPower.calculate(n,k,m);
+        assertEquals(result, n.modPow(k,m));
+    }
+
+    @Test
+    public void test() {
+        testLong(2,2,10);
```
yanglbme:
```java
testLong(2,2,10); 
->
testLong(2, 2, 10);
```

PR #732, opened 2019-04-11
Efficiency

markettes:
Just two small things that in case the number was very big could be helpful.

### DIFF
--- a/ciphers/Caesar.java
+++ b/ciphers/Caesar.java
@@ -24,8 +24,8 @@ public class Caesar {
     public static String encode(String message, int shift) {
         String encoded = "";
 
-        while (shift >= 26) { // 26 = number of latin letters
-            shift -= 26;
+        if (shift >= 26) { // 26 = number of latin letters
+            shift %= 26;
         }
 
         final int length = message.length();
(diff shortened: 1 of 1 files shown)


markettes:
I have been testing both ways to do it measuring time. Both are very fast and truly this pull request doesn't fix anything very important, but when the computer is working in other things and interrupts the thread, in your case you have a small delay. On the other hand, my solution makes it an atomic operation.
Thanks.

ciphers/Caesar.java
```diff
@@ -24,8 +24,8 @@
     public static String encode(String message, int shift) {
         String encoded = "";
 
-        while (shift >= 26) { // 26 = number of latin letters
-            shift -= 26;
+        if (shift >= 26) { // 26 = number of latin letters
```
havanagrawal: Good catch! In fact, even this `if` is not necessary, since the modulo operation is no-op if `shift` < 26, i.e.

`shift % 26 == shift` if 0 <= `shift` < 26

markettes: Oh! That's true! Less code then :)

havanagrawal:
LGTM

PR #754, opened 2019-05-14
Added Singleton Design pattern implementation in Java and its test

abhijay94:
The singleton pattern is a design pattern that restricts the instantiation of a class to one "single" instance.
This is useful when exactly one object is needed to coordinate actions across the system. The term comes from the mathematical concept of a singleton.

### DIFF
--- a/src/test/java/com/designpatterns/singletonpattern/SingletonTest.java
+++ b/src/test/java/com/designpatterns/singletonpattern/SingletonTest.java
@@ -23,7 +23,6 @@ public class SingletonTest {
                 try {
                     Singleton singletonInstance = Singleton.getInstance();
                     int singletonInsCode = singletonInstance.hashCode();
-                    System.out.println(singletonInsCode);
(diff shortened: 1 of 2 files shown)


src/test/java/com/designpatterns/singletonpattern/SingletonTest.java
```diff
@@ -0,0 +1,45 @@
...
+        boolean testFailed = false;
+        ExecutorService es = Executors.newCachedThreadPool();
+        // Creates 15 threads and makes all of them access the Singleton class
+        // Saves the hash code of the object in a static list
+        for (int i = 0; i < 15; i++)
+            es.execute(() -> {
+                try {
+                    Singleton singletonInstance = Singleton.getInstance();
+                    int singletonInsCode = singletonInstance.hashCode();
+                    System.out.println(singletonInsCode);
```
yanglbme:
```suggestion
```

PR #763, opened 2019-05-21
Added the prototype pattern and its unit test

abhijay94:
The prototype pattern is used when the type of objects to create is determined by a prototypical instance, which is cloned to produce new objects.


abhijay94:
Updated SimpleNoiseTest.java to use the JUnit library mentioned in the development guidelines for this project instead of the JUnit library from "jupiter". Having two jars for the same purpose didn't make sense.

PR #766, opened 2019-05-22
Cast the second operand of Math.pow to double - SonarQube analysis

abhijay94:
SonarQube showed these calls as probable minor bugs. Since Math.pow() accepts double type operands, the second operand should be type-casted explicitly for safety.

### DIFF
--- a/src/main/java/com/generation/SimplexNoise.java
+++ b/src/main/java/com/generation/SimplexNoise.java
@@ -36,7 +36,7 @@ public class SimplexNoise {
 
             this.octaves[index] = new SimplexNoiseOctave(random.nextInt());
             this.frequencys[index] = Math.pow(2, index);
-            this.amplitudes[index] = Math.pow(persistence, octaveCount - index);
+            this.amplitudes[index] = Math.pow(persistence, (double)octaveCount - index);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/generation/SimplexNoise.java
```diff
@@ -36,7 +36,7 @@ public SimplexNoise(int largestFeature, double persistence, long seed) {
 
             this.octaves[index] = new SimplexNoiseOctave(random.nextInt());
             this.frequencys[index] = Math.pow(2, index);
-            this.amplitudes[index] = Math.pow(persistence, octaveCount - index);
+            this.amplitudes[index] = Math.pow(persistence, (double)octaveCount - index);
```
yanglbme:
```suggestion
            this.amplitudes[index] = Math.pow(persistence, (double) octaveCount - index);
```

src/main/java/com/generation/SimplexNoise.java
```diff
@@ -103,7 +103,7 @@ public double getNoise(int x, int y, int z) {
         for (int index = 0; index < this.octaves.length; index++) {
 
             double frequency = Math.pow(2, index);
-            double amplitude = Math.pow(this.persistance, this.octaves.length - index);
+            double amplitude = Math.pow(this.persistance, (double)this.octaves.length - index);
```
yanglbme:
```suggestion
            double amplitude = Math.pow(this.persistance, (double) this.octaves.length - index);
```

yanglbme:
Hi @abhijay94 , there is one more place need to be resolved.

image

PR #769, opened 2019-05-27
Added adapter pattern and its test case

abhijay94:
An Adapter pattern acts as a connector between two incompatible interfaces that otherwise cannot be connected directly. An Adapter wraps an existing class with a new interface so that it becomes compatible with the client’s interface.


12Manoz:
I have not seen any Design Patterns commits in this repository before it would be great if those were taken a part from the algorithms repo

abhijay94:
@yanglbme What do you think about this? Creational patterns are already merged into the Development branch. I think it will be good to have them, especially for the Java repo.

> It would be nice to avoid design patterns as this repository is solely for algorithms and DS

Issue #733, opened 2019-04-11
BFS Algorithm for adjacency list implementation of graph.


Abhishek-kumar-singh:
BFS for the adjacency matrix is already present I would like to contribute BFS for adjacency list implementation of the graph.
Can this be assigned to me?

workinprz:
Please help

yoyou446:
and DFS, using both recursive and non-recursive code perhaps?

yoyou446:
Honestly, I think the code would be easier to right then the algorithm summary

PR #777, opened 2019-06-09
BFS Traversal on Adjacency List implementation of Graph #733

Abhishek-kumar-singh:
BFS Algorithm for adjacency list implementation of the graph. Issue #733

### DIFF
--- a/DataStructures/Graphs/BFSAdjacencyListGraph.java
+++ b/DataStructures/Graphs/BFSAdjacencyListGraph.java
@@ -1,0 +1,156 @@
+
+
+import java.util.LinkedList;
+import java.util.Queue;
+
+/**
+ * Adjacency List implementation of Graph
+ * @author Abhishek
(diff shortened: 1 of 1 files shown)


DataStructures/Graphs/BFSAdjacencyListGraph.java
```diff
@@ -0,0 +1,156 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Main method.
+	 * 
+	 * @param args command line arguments
+	 */
+	public static void main(String[] args) {
```
prashantdoshi28, line 77: If possible add test class instead of using `main` method. 

If you don't know how to write test class, please refer to [DecimalToAnyBaseTest.java](https://github.com/TheAlgorithms/Java/blob/Development/src/test/java/com/conversions/DecimalToAnyBaseTest.java)

PR #781, opened 2019-06-26
Add Bucket Sort

obelisk0114:
Add bucket sort and its time complexity


obelisk0114:
Bucket sort needs the `numeric` value of every element to determine every element's bucket. Existing interface `SortAlgorithm` and `SortUtils` only extends `Comparable` interface. They can't be applied to bucket sort directly. We might use `unicode` to provide the `numeric` value of every element, but it didn't have relation with the two interface.

PR #787, opened 2019-07-07
Update Armstrong.java

travellers:
So, not only the integer less than 1000 can be checked, all positive integer can be checked.

### DIFF
--- a/Others/Armstrong.java
+++ b/Others/Armstrong.java
@@ -14,7 +14,7 @@ public class Armstrong {
 
     public static void main(String[] args) {
         scan = new Scanner(System.in);
-        int n = inputInt("please enter the number");
+        int n = inputInt("please enter the number: ");
         boolean isArmstrong = checkIfANumberIsAmstrongOrNot(n);
         if (isArmstrong) {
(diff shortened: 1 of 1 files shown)


Others/Armstrong.java
```diff
@@ -32,22 +32,24 @@ public static void main(String[] args) {
...
+        placeNum = numberStr.length();
         temp = number;
-        while (number > 0) {
-            remainder = number % 10;
-            sum = sum + (remainder * remainder * remainder);
-            number = number / 10;
+        while (temp > 0) {
+            remainder = temp % 10;
+            sum += Math.pow(remainder, placeNum);
+            temp = temp / 10;
```
yanglbme:
```suggestion
            temp /= 10;
```

Others/Armstrong.java
```diff
@@ -32,22 +32,24 @@ public static void main(String[] args) {
...
         }
-        if (sum == temp) {
+        if (sum == number) {
```
yanglbme: Suggestion:

Simplify code by using `return sum == number;` here.

PR #790, opened 2019-07-08
Reverse String using recursion

prashantp995:


rjtmahinay:
Hi @prashantp995 
In Line 13 of your reverseString() method if the input is null, input.length will throw NullPointerException. I suggest to use input.isEmpty() instead of input.length <= 0

PR #795, opened 2019-07-19
 feat: Add Maven frameworks support

laingke:
1. Add Maven frameworks support to help it more like a Java project.
2. Refact all the testcase by using JUnit Jupiter (in most cases, we can also call it JUnit 5) instead of JUnit 4(ALL refactor is PASSED on my local machine). I consider it necessary for developers to use a newer test API.
3. Add a resources directory to help test code reading the specified resource.(It shall be a better implement)
4. According to the change above, I update the README by deleting the fourth line from the bottom. Developer can import JUnit automatically.


laingke:
A duplicate work as #529.
Diff from #529:
1. a newer JUnit Jupiter version;
2. a specified resources path of test file;
3. a specified encoding.

laingke:
Duplicate of #529.

Diff from #529:
a newer JUnit Jupiter version;
a specified resources path of test file;
a specified encoding.

PR #796, opened 2019-07-25
Non-recursive heap sort implementation

1170300115:


1170300115:
Make little changes to the original HeapSort, thanks to the original authorPodshivalov Nikita (https://github.com/nikitap492)

Issue #474, opened 2018-07-29
Convert the repository into a Java project 


varunu28:
@christianbender @AnupKumarPanwar @dynamitechetan 

This is something which something which we have discussed in the past and have been raised by multiple users of the repo. The repo can be converted to a complete Java project with correct package structure, JUnit tests and Java language standard checkers. I want that anyone who is adding a new algorithm should add respective tests and follow a set coding standard so that the repo can become more usable for users who can then clone it into a local IDE and work on it.

Similar to [Scala Repo](https://github.com/TheAlgorithms/Scala)

I propose that we create a development branch on which we can start making the changes and replace it with the master once it fulfills the above-mentioned requirement. I would like to take the initiative and start this task. 

Please share your feedback.

burnz248:
Hey Anup. I would like to help with this as well. Let me know what you need help with

MarisaAfuera:
Hi Anup. Great idea. I also can help if you need.

rahul-vishwadhaar:
Hello Anup,

This seems to be great thing. Please let me know if I can contribute.

ahmedsameha1:
I'm in.

habibu:
Also, I am in and happy to contributes

lq920320:
So great, happy to join and contributes.

xiaomabuchicao:
我还以为不能说中文

MayV:
I'd like to help and contribute.

esarella:
Presently working on solving most problems. I'm more than willing to help and contribute.

asdiamond:
I can definitely help and contribute to this.

giannisal:
I'd love to contribute to make this repo a java project.

From what I've seen so far, the development branch is identical to the master in terms of structure and content. I'm used to working with maven in Java, however I saw that the scala repo is based on Travis CI. Would you like me to work on either of them, create the project and test for some of the existing classes?

varunu28:
@giannisal @channel I initially created the development branch. I would move the code to that branch in a project structure by the weekend and would update the README for the test case format. After that we can start accepting contributions to the development branch for the tests
We are using ScalaTest spec in the Java repo

varunu28:
I have updated the development branch and added a instructions to the README. We can start with the migration. 

@christianbender Please make sure every PR is adhering to the points mentioned in README on [Development branch](https://github.com/TheAlgorithms/Java/tree/Development)

lq920320:
[...]

PR #815, opened 2019-08-24
refactor code to match new maven-based project structure

SirWindfield:
This commit adds all necessary files as well as correctly specifies the package names for each file. The project now fully mimics maven's project structure (as indicated as a goal earlier in issue #474).

The goal is to make working on the project easier for others. They can now clone the repo and use the `Open as maven project` that is included in every mainstream IDE/editor.

**Edit: ** All tests pass. Might be a good idea to setup a Travis CI config that will check for tests.


yanglbme:
#795

Issue #816, opened 2019-08-27
algorithms in the others folder are too open ended


furahadamien:
I was going through most of the algorithms in the [others](https://github.com/TheAlgorithms/Java/tree/master/Others) folder, and to me, they appear to be too open-ended and as a result, they fail to tackle most of the edge cases. Most of them are addressing the problem but they are not looking at all sorts of inputs. For example, the [countWords](https://github.com/TheAlgorithms/Java/blob/master/Others/CountWords.java) algorithm doesn't look at instances where we have a sequence of non-alphanumeric characters. Should you consider "#####" a word for example? The other problem is that most of the algorithms do not actually explain what they are doing in the first place.

yanglbme:
@furahadamien thanks for your feedback. Could you create a pull request to fix them? PR is welcome in [TheAlgorithms/Java](https://github.com/TheAlgorithms/Java) repository.

PR #818, opened 2019-08-28
Word count

furahadamien:
added a new method that counts words in a string. As per issue [#816](https://github.com/TheAlgorithms/Java/issues/816#issue-485845732), the method ignores all non-alphanumeric characters when counting words


yanglbme:
@furahadamien 👍

PR #825, opened 2019-09-14
Added CombSortTest.java

jayyusi:
copied over CombSort.java from master branch to Development branch.
Added CombSortTest.java to Development branch

### DIFF
--- a/src/test/java/com/sorts/CombSortTest.java
+++ b/src/test/java/com/sorts/CombSortTest.java
@@ -9,12 +9,10 @@ import static org.junit.jupiter.api.Assertions.*;
 
 class CombSortTest {
 
-    //One global combSort instance variable
     CombSort combSort = new CombSort();
 
-    //Integer Sort Test
     @Test
-    void CombSortIntegerTest(){
+    void testIntegerSort(){
 
         Integer array1[] =  { 9, 5, 4, 6, 0, 100, 91, 40, 140, 500, 4, -1, -19};
         Integer array2[] =  { 9, 5, 4, 6, 0, 100, 91, 40, 140, 500, 4, -1, -19};
@@ -29,9 +27,8 @@ class CombSortTest {
         Assertions.assertArrayEquals(array1,array2);
     }
 
-    //String Sort Test
     @Test
-    void CombSortStringTest(){
+    void testStringSort(){
 
         String array1[] =  { "Ten" , "Nine", "Eight", "Seven", "Six", "Five", "Four", "One", "Zero" };
         String array2[] =  { "Ten" , "Nine", "Eight", "Seven", "Six", "Five", "Four", "One", "Zero" };
@@ -44,12 +41,10 @@ class CombSortTest {
 
         //compare the two results, our CombSort vs. Java's sort
         Assertions.assertArrayEquals(array1,array2);
-
     }
 
-    //Double Sort Test
     @Test
-    void CombSortDoubleTest(){
+    void testDoubleSort(){
 
         Double array1[] =  { 90.0, -10.0, 0.0, -1.0, -23.98, 10.76, 20.92 };
         Double array2[] =  { 90.0, -10.0, 0.0, -1.0, -23.98, 10.76, 20.92 };
@@ -62,12 +57,10 @@ class CombSortTest {
 
         //compare the two results, our CombSort vs. Java's sort
         Assertions.assertArrayEquals(array1,array2);
-
     }
 
-    //Character Sort Test
     @Test
-    void CombSortCharacterTest(){
+    void testCharacterSort(){
 
         Character array1[] =  { 'Z', 'A', 'B', 'z', 'b', 'c' };
         Character array2[] =  { 'Z', 'A', 'B', 'z', 'b', 'c' };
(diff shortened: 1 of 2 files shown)


src/test/java/com/sorts/CombSortTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class CombSortTest {
+
+    //One global combSort instance variable
```
SirWindfield: The comment can be removed, not really needed.

src/test/java/com/sorts/CombSortTest.java
```diff
@@ -0,0 +1,85 @@
...
+    CombSort combSort = new CombSort();
+
+    //Integer Sort Test
```
SirWindfield: Same goes here. The naming scheme of your methods should be `testSomething`. Like `testIntegerSort` and `testStringSort`. Methods in Java start with a lowercase letter. Make sure to correct that everywhere else too.

src/test/java/com/sorts/CombSortTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        //using Java's tested sort algorithm to sort our unsorted array
+        Arrays.sort(array1);
+
+        //sort our unsorted array using our new CombSort algorithm
+        combSort.sort(array2);
+
+        //compare the two results, our CombSort vs. Java's sort
+        Assertions.assertArrayEquals(array1,array2);
+
```
SirWindfield: Remove empty lines at the end of each method.

src/test/java/com/sorts/CombSortTest.java
```diff
@@ -0,0 +1,85 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+class CombSortTest {
+
+    //One global combSort instance variable
+    CombSort combSort = new CombSort();
+
+    //Integer Sort Test
+    @Test
+    void CombSortIntegerTest(){
```
SirWindfield: Method name, see above.

src/test/java/com/sorts/CombSortTest.java
```diff
@@ -0,0 +1,85 @@
...
+        //sort our unsorted array using our new CombSort algorithm
+        combSort.sort(array2);
+
+        //compare the two results, our CombSort vs. Java's sort
+        Assertions.assertArrayEquals(array1,array2);
+    }
+
+    //String Sort Test
+    @Test
+    void CombSortStringTest(){
```
SirWindfield: Method name, see above.

src/test/java/com/sorts/CombSortTest.java
jayyusi: done.

src/test/java/com/sorts/CombSortTest.java
jayyusi: how about combSortIntegerTest ?

src/test/java/com/sorts/CombSortTest.java
jayyusi: done.

src/test/java/com/sorts/CombSortTest.java
SirWindfield: No, just change it to `testStringSort` and `testIntegerSort`. Your class is named `CombSortTest` already, no need to add it to the method names. Alongside that, all test methods from the other classes are named using the ``testXXX` scheme. And to keep consistency it would be better if you could change them too :)

jayyusi: CycleSortTest class does not follow the consistency you mentioned ?!
it follows what I suggested :)
class CycleSortTest {

    @Test
    void cycleSortIntegerTest() {

        CycleSort cycleSort = new CycleSort();

SirWindfield: If you take a look at all the other tests, they follow the convention `testXXX` where XXX gives some details of what is being tested. In your case, it would be `testStringSort` or something similar.

Well, looks like some tests need a rename then :) They shouldn't be named like that though. Even the JUnit site lists the `testXXX` naming scheme as a go-to.

jayyusi: haha, yes now you see what I mean, and now I see what you mean, I just noticed in the /search folder, the tests follow the naming convention of testXXX.

I did the changes, did you want me to also remove all the comments like this one:
//Integer Sort Test

I'm almost done, and thank you for the review and the edit suggestions. Its always good to have another pair of eyes. I admit I totally overlooked the method names.

SirWindfield: Would be nice if you could remove any comments that don't really add value, yes. If they are needed to explain _why_ a certain thing is done, let them in. But something like `// Integer Sort test` is not really needed.

No problem :)

jayyusi: alright, done.
I made a commit to the same pull request. 
Do you know when will pull requests be merged with the development branch ?

SirWindfield: It usually takes some time...

ayaankhan98:

PR #861, opened 2019-09-24
fix memory leak

shellhub:


DataStructures/Lists/SinglyLinkedList.java
```diff
@@ -61,7 +61,9 @@ public void deleteHead() {
             throw new RuntimeException("The list is empty!");
         }
 
+        Node destory = head;
```
Naibeck: Typo on destory, I believe you're aiming for destroy

shellhub: Thanks!

Issue #784, opened 2019-07-01
Looks like BFS and DFS is missing 


darshantumkur:
https://github.com/TheAlgorithms/Java/blob/master/DataStructures/Graphs/DFS.java

Sasi-007:
The code doesn't exist their. Can i Take this issue.

swatiprajapati08:
BFS and DFS in JAVA #1019

anilabha:
I just send a marge request on Graph Algorithms 
1.Bellman ford algorithm
2.BFS
3.DFS
4.Prims algorithm
5.Dijkstra algorithm

PR #932, opened 2019-10-01
BFS traversal added

vishnoitanuj:
Issue #784

### DIFF
--- a/DataStructures/Graphs/BFS.java
+++ b/DataStructures/Graphs/BFS.java
@@ -1,0 +1,47 @@
+import java.util.*;
+
+public class BFSs {
+
+	public static void main(String[] args) {
+		Scanner s = new Scanner(System.in);
+		int V = s.nextInt();
+		int E = s.nextInt();
+
+		int edges[][]=new int[V][V];
+        for(int i=0;i<E;i++){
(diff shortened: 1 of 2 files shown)


DataStructures/Graphs/BFS.java
```diff
@@ -0,0 +1,47 @@
...
+			}
+		}
+    }
+    
+    private static void printBFS(int edges[][],int sv,boolean visited[]){
+        Queue<Integer> q=new LinkedList<>();
+        q.add(sv);
+        int n=edges.length;
+        visited[sv]=true;
+        String s=new String();
```
Amanda1232, line 35: Our static analysis engine detects that your pull request includes a change that **method invokes inefficient new String() constructor**. According to [the best practice](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dm-method-invokes-inefficient-new-string-constructor-dm-string-void-ctor), creating a new java.lang.String object using the no-argument constructor **wastes memory** because the object so created will be functionally indistinguishable from the empty string constant "".  Java guarantees that identical string constants will be represented by the same String object.  **Therefore, you should just use the empty string constant directly.**

Issue #560, opened 2018-10-08
Other/Dijkshtra.java has some problems. I would like to recommend it's removal.


rmakynen:
Other/Dijkshtra.java has some problems:
- Hard-coded magic numbers (1000000l)
- initialization of a very large 2-dimensional array when it is not needed when using another implementation. - No description was given on the input format the algorithm was expecting.
- Poor commenting.
- Name of the algorithm is wrong. Should be "Dijkstra"

I have created a new pull request #559 that has new code for the algorithm. It has none of the above mentioned problems from what I can tell.

If the previously mentioned pull gets accepted, then I would recommend that the old version "Dijkshtra.java" should be removed.

prashantdoshi28:
Hey @vishnoitanuj , I have requested some changes, go through my review for those.

PR #933, opened 2019-10-01
Djikstra optimized

vishnoitanuj:
As of errors stated in #560 . The issue related solved with code optimization

### DIFF
--- a/DataStructures/Graphs/BFS.java
+++ b/DataStructures/Graphs/BFS.java
@@ -1,0 +1,47 @@
+import java.util.*;
+
+public class BFSs {
+
+	public static void main(String[] args) {
+		Scanner s = new Scanner(System.in);
+		int V = s.nextInt();
+		int E = s.nextInt();
+
+		int edges[][]=new int[V][V];
+        for(int i=0;i<E;i++){
+            int fv=s.nextInt();
+            int sv=s.nextInt();
+            edges[fv][sv]=1;
+            edges[sv][fv]=1;
+        }
+        printBFS(edges);
+	}
+    
+    private static void printBFS(int edges[][]){
+        boolean visited[] = new boolean[edges.length];
+        // printBFS(edges,0,visited);
+		for(int i=0;i<edges.length;i++) {
+			if(!visited[i]) {
+				printBFS(edges,i,visited);
+			}
+		}
+    }
+    
+    private static void printBFS(int edges[][],int sv,boolean visited[]){
+        Queue<Integer> q=new LinkedList<>();
+        q.add(sv);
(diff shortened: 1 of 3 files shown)


DataStructures/Graphs/BFS.java
```diff
@@ -0,0 +1,47 @@
...
+			}
+		}
+    }
+    
+    private static void printBFS(int edges[][],int sv,boolean visited[]){
+        Queue<Integer> q=new LinkedList<>();
+        q.add(sv);
+        int n=edges.length;
+        visited[sv]=true;
+        String s=new String();
```
Amanda1232, line 35: Our static analysis engine detects that your pull request includes a change that **method invokes inefficient new String() constructor**. According to [the best practice](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dm-method-invokes-inefficient-new-string-constructor-dm-string-void-ctor), creating a new java.lang.String object using the no-argument constructor **wastes memory** because the object so created will be functionally indistinguishable from the empty string constant "".  Java guarantees that identical string constants will be represented by the same String object.  **Therefore, you should just use the empty string constant directly.**

DataStructures/Graphs/DFS.java
```diff
@@ -0,0 +1,46 @@
+import java.util.*;
+
+public class DFS {
```
prashantdoshi28, line 3: Please make separate PRs for unrelated codes.

DataStructures/Graphs/BFS.java
```diff
@@ -0,0 +1,47 @@
+import java.util.*;
+
+public class BFSs {
```
prashantdoshi28, line 3: Please make separate PRs for unrelated codes.

Others/Dijkstra.java
```diff
@@ -1,192 +1,52 @@
...
+	public static int findMinVertex(int a[],boolean visited[]) {
+		int min=-1;
+		for(int i=0;i<a.length;i++) {
+			if(!visited[i] && (min==-1 || a[i]<a[min]))
+				min=i;
+		}
+		return min;
+	}
+
+	public static void main(String[] args) {
```
prashantdoshi28, line 37: Try not to use this approach of having 
```java
public static void main(String[] args) {...}
```
call your actual method. 

Instead create your class while keeping in mind of how others will use this i.e. by creating objects and calling your actual methods. So make your methods 
```java
public <return-type> methodName()
```

Please make this changes.

prashantdoshi28 (changes requested):
I have requested some changes, please go through my comments.

PR #943, opened 2019-10-02
corrected prime check

debarghya472:


shellhub:
previous version was better solution. So I close this PR

PR #964, opened 2019-10-04
simplified program

yatin9988:
removed uneccessary conditions.


Spikatrix:
Your code is flawed as it returns the function will return `true` for `n <= 1`

PR #972, opened 2019-10-05
Added Feistel cipher

Cornul11:


Cornul11:
Could the maintainers take a look?

PR #975, opened 2019-10-06
Update Factorial.java

NeelP2121:
decreased line of code by 3


Spikatrix:
Please test the code before making such changes. 

Your modified code is flawed as it will begin recursion decrementing `n` by 1 in every step and when `n` becomes negative, the `ArithmeticException` will be thrown.
 
Also note that recursion consumes stack space and decreases the readability of the code when compared to iteration. This might result in a Stack Overflow for large values of `n`. You shouldn't try to reduce the number of lines of code by sacrificing readability.

PR #979, opened 2019-10-06
update

shellhub:
* previous version can't delete first element when only one element exists (bug)
* this version is more readable and convenient(such as `delete` `insert`)

Thanks ^_^


shellhub:
@yanglbme

Issue #638, opened 2018-11-08
Double Ended queue D.S is not implemented


ales-blaze:
And i would like to implement it , Any suggestions?

haruntuncay:
Hello, if this is still needed, I would like to contribute and implement it.
Also, by double ended queue, do you mean a deque ?
Please let me know if I can be assigned to this.
Thanks.

pgupta1795:
Hi,Please let me know if this issue can be assigned and implemented?

Nuthi-Sriram:
I would like to work on this

Nuthi-Sriram:
I have submitted a pull request for deque hope you like the code

iamlatin:
FEDERALGRANTS.COM  nada activo

iamlatin:
api

prashantdoshi28:
Hey @Nuthi-Sriram , I have reviewed your PR, please go through my findings in my review. Let's make sure that this issue can be closed in few days.

PR #980, opened 2019-10-06
Added deque program

Nuthi-Sriram:
#638 Implementation of double-ended queue

### DIFF
--- a/DataStructures/Queues/Deque.java
+++ b/DataStructures/Queues/Deque.java
@@ -1,0 +1,163 @@
+/**Implementation of double ended queue*/
+
+import java.util.Scanner;
+
+public class Deque {
+	private int maxSize;
+	private int f, r, a[];
+
+	public Deque(int S) {
+		maxSize = S;
+		a = new int[S];
+		f = r = -1;
+	}
+
+	int deleteRear() {
+		if (f == -1)
+			return -1;
+		else if (f == r) {
+			int x = a[r];
+			f = r = -1;
+			return x;
+		} else {
+			int x = a[r];
+			r--;
+			return x;
+		}
+	}
+
+	int insertFront(int x) {
+		if (f == 0)
+			return 0;
+		else if (f == -1 || r == -1) {
+			f++;
+			r++;
+			a[f] = x;
+			return x;
+		} else {
+			f--;
+			a[f] = x;
+			return x;
+		}
+	}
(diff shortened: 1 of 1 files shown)


DataStructures/Queues/Deque.java
```diff
@@ -0,0 +1,163 @@
+/**Implementation of double ended queue*/
+
+import java.util.Scanner;
+
+public class Deque {
+	private int maxSize;
+	private int f, r, a[];
```
ales-blaze: Variables name should reflect what it is not , instead of using f , r ,a you should use first , rear , queue or something of your choice but descriptive in nature

DataStructures/Queues/Deque.java
```diff
@@ -0,0 +1,163 @@
...
+			return -1;
+		else {
+			for (int i = f; i <= r; i++)
+				System.out.println(a[i]);
+			return 0;
+		}
+
+	}
+
+	public static void main(String[] args) {
```
ales-blaze: Instead of main function , you should have written test cases , which will be far better than using main function

DataStructures/Queues/Deque.java
Nuthi-Sriram: Added descriptive names

DataStructures/Queues/Deque.java
```diff
@@ -0,0 +1,163 @@
...
+			System.out.println("Enter 1 to perform insertRear");
+			System.out.println("Enter 2 to perform deleteFront");
+			System.out.println("Enter 3 to perform deleteRear");
+			System.out.println("Enter 4 to perform insertFront");
+			System.out
+					.println("Enter 5 to display all the elements present in the double ended queue");
+
+			choice = sc.nextInt();
+			switch (choice) {
+			case 1: {
```
prashantdoshi28, line 104: You can omit `{}` braces in case statement, like below : 

```diff
- case 1: {...} 
+ case 1: ...
```

prashantdoshi28 (changes requested):
I have requested some changes, please go through my comments for those. 

Also add checks to your PR which ensures that this code matches or fulfils required coding standards of repo.

Nuthi-Sriram:
I have made the requested changes and opened a pull request https://github.com/TheAlgorithms/Java/pull/2036

PR #995, opened 2019-10-08
toString

shellhub:
* traverse queue using `toString()` method


shellhub:
Please review! @yanglbme 🙂

PR #1004, opened 2019-10-10
Add Power Of Two in Math

shubham-ahuja13:


shellhub:
see: https://github.com/TheAlgorithms/Java/blob/master/Maths/PowerOfTwoOrNot.java

PR #1012, opened 2019-10-11
Three files added

mashkarharis:
i have added three methods.Those,
1.odd_composite     (int)----->(Whether it odd composite or not)
2.Cylinder                 (radius,height)-----------> (volume,area)
3.Circle                     (radius)----------------->(are,perimeter)
I think this will be helpful.
Thank you !


mashkarharis:
Thank you

PR #1116, opened 2019-10-23
added factorial using recursion

arpitgupta1906:


shellhub:
File existed! so closed

PR #1194, opened 2019-12-27
chain of responsibility pattern

ali4j:
Two commits are pushed.

1. CoR pattern sample is added.
2. A class is moved to its specific package which I mistakenly forgot to put it in its own one in my last commit.


ali4j:
There is a problem with the tests in this PR, I close it and make a new one containing a new set of tests.

PR #1199, opened 2020-01-04
Add simple gradle workflow for Development

myronrotter:
Adds simple github on push gradle workflow for Development branch.


sbatten1969:
@myronrotter :  Nice work.  I tested on my local machine, and this is a positive improvement on the default gradle workflow, especially the unit test executions.

PR #1260, opened 2020-03-25
Checker-Annotation

guy34code:


Others/Dijkstra.java
```diff
@@ -123,7 +125,7 @@ public Graph(Edge[] edges) {
     /**
      * Runs dijkstra using a specified source vertex
      */
-    public void dijkstra(String startName) {
+    public void dijkstra(@NonNull String startName) {
```
Maxi17: You don't need to add `@NonNull` because it is the default qualifier.

Others/Dijkstra.java
```diff
@@ -104,7 +106,7 @@ public String toString() {
     /**
      * Builds a graph from a set of edges
      */
-    public Graph(Edge[] edges) {
+    public Graph(@Positive Edge[] edges) {
```
Maxi17: The `@Positive` annotation is not in the NullnessChecker  hierarchy. You mixed the NullnessChecker and IndexChecker qualifiers in this case study. It is not wrong to annotate a file using 2 type systems, but it is not the usual way a case study is done.

Also, the annotation is incorrect. What you meant is that the array `edges` contains objects of type `Edge`, which can't be `@Positive`, because it is meaningless. Only types like `int`, which represent a number can be `@Positive`, `@NonNegative` and some other IndexChecker qualifiers.

Others/Dijkstra.java
```diff
@@ -95,7 +97,7 @@ public int compareTo(Vertex other) {
             return Integer.compare(dist, other.dist);
         }
 
-        @Override
+        @Override
```
Maxi17: Avoid making changes that add no value to the PR.

Others/Dijkstra.java
```diff
@@ -56,7 +58,7 @@ public static void main(String[] args) {
         public final String v1, v2;
         public final int dist;
 
-        public Edge(String v1, String v2, int dist) {
+        public Edge(@NonNull String v1, @NonNull String v2, @NonNegative int dist) {
```
Maxi17: `@NonNull is not needed, see the explanation below.

Others/Dijkstra.java
```diff
@@ -56,7 +58,7 @@ public static void main(String[] args) {
         public final String v1, v2;
         public final int dist;
```
Maxi17: I expected a `@NonNegative` annotation here, too. This field is initialized in the constructor with a non-negative value.

Others/Dijkstra.java
guy34code: Sorry, I'll keep in mind next time.

Others/Dijkstra.java
guy34code: Done!

Others/Dijkstra.java
guy34code: Done!

Others/Dijkstra.java
guy34code: Okay, Done!

PR #1261, opened 2020-03-26
DecimalToHexadecimal and BFS

iheb1196:
I changed the DecimalToHexadecimal Class so it become more useful where the user after implementing the method can used it in a concrete probleme where he has to to change a given number from decimal to Hexadecimal


ayaankhan98:

PR #1266, opened 2020-04-06
solved problem of lowercase letters

aniket3991:
solved problem with lowercase letters

### DIFF
--- a/Conversions/AnyBaseToAnyBase.java
+++ b/Conversions/AnyBaseToAnyBase.java
@@ -23,12 +23,31 @@ public class AnyBaseToAnyBase {
 
     public static void main(String[] args) {
         Scanner in = new Scanner(System.in);
-        String n;
-        int b1, b2;
+        String n ;
+	    
+        int b1, b2,ifzero = 0;
         while (true) {
             try {
                 System.out.print("Enter number: ");
                 n = in.next();
+		    
+		    
+		char[] digits = n.toCharArray();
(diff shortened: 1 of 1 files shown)


drabbit75:
it seems not to solve lowercase letter problem . do you want to solve "n=0" bug ?

drabbit75:
ready to close with the doubt

aniket3991:
@drabbit75  can you give me an input where you find problems in lower case letters

drabbit75:
> @drabbit75 can you give me an input where you find problems in lower case letters

the codes seems not to solve "lower case problem", what  bugs do you want to fix

aniket3991:
actually when you wanted to covert a more than base 10 numbers, you have to give the values of "abcdef" in capital letters otherwise code was not working so i fixed that

drabbit75:
in the codes ：
 {'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'A', 'B', 'C', 'D', 'E',
                'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V',
                'W', 'X', 'Y', 'Z'};" // digitsForBase contains all the valid digits for the base given".
this limits the input and I think it is not a bug and not neccesary to fix personally.

aniket3991:
So you can say that i expanded the input limit

mamadaliev (approved):

PR #1270, opened 2020-04-10
Add Harmonic Series

quackduck:
Has two methods

`harmonic()` - Calculates the partial sum of the Harmonic Series to the `n`th number in the series.
`revharmonic()` - Calculates the value of `n` that will give you the value you inputted or more if you feed it into `harmonic()`.

This is my first ever contribution to Open Source Software. Thank you and have a great day!

### DIFF
--- a/Maths/Harmonic.java
+++ b/Maths/Harmonic.java
@@ -1,3 +1,4 @@
+package Maths;
 
 /**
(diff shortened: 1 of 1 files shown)


quackduck:
Jeez

Issue #1308, opened 2020-05-09
Missing Algorithms


joshiujjawal22:
- Bucket Sort
- Knuth's Morris Pratt (KMP) algorithm 
- Lower and Upper bound using binary search
- 3 sum Algorithm
can i add those?

deadshotsb:
Yes, sure. Please do not wait for permission. Just go ahead and submit pull requests for review.

PR #1316, opened 2020-05-18
Missing Codes

joshiujjawal22:
I have added some missing codes:
- Array rotation
- 3 sum (to find triplets in efficient manner)
Pull request related to #1308  .
Further, I will add some more codes related to binary search(Lower/upper bound).

### DIFF
--- a/Others/3 sum.java
+++ b/Others/3 sum.java
@@ -1,0 +1,48 @@
+package Others;
+
+import java.util.Scanner;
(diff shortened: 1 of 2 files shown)


deadshotsb:
Provide some examples cases at the top using multi-line comment, it will help any user have a clear idea of what your code does

joshiujjawal22:
I have added some examples cases.You may check.

PR #1323, opened 2020-05-22
Added Missing Algorithms

prateekKrOraon:
String matching algorithms
- Rabin-Karp algorithm
- String matching using Finite Automata


deadshotsb:
@prateekKrOraon Good work

PR #1330, opened 2020-05-28
Corrected ShellSort.java

himalayanDrifter:
I have corrected the shell sort and it is working fine for all the test cases but the previous code which i deleted looks very much professional then the one I wrote.


shellhub:
I think previous version is better. Thanks :)

PR #1338, opened 2020-06-01
Minimum sum partition

swatiprajapati08:
Given an array, the task is to divide it into two sets S1 and S2 such that the absolute difference between their sums is minimum.


DynamicProgramming/Minimum sum partition
```diff
@@ -0,0 +1,64 @@
+// Partition a set into two subsets such that the difference of subset sums is minimum
+
```
deadshotsb, line 2: Kindly provide some sample test cases

swatiprajapati08: Added.

PR #1339, opened 2020-06-02
added RR scheduling with process queue with list

badallogical:
i have implemented the RR scheduling with the linkedlist as process queue including any arrival time possible for process.


badallogical:
tell me after review my code whether you want to implement it whole using linkedlist, i just use ArrayList for process queue to hold process arrive at the same time while cpu is busy.

PR #1350, opened 2020-06-12
A-Star Algorithm Included

MarcosVillacanas:
A-Star is one of the most popular graph path-search algorithms.
Due to the fact that it uses a heuristic function to decide between the available paths, it is considered to be a smart algorithm.
I have included an example with the graph drawn, the output formatted, and comments to explain the code.

I hope you accept me, let me know if I have to change anything :)

### DIFF
--- a/DataStructures/Graphs/A_Star.java
+++ b/DataStructures/Graphs/A_Star.java
@@ -1,3 +1,7 @@
+/*
+	Time Complexity = O(E), where E is equal to the number of edges
+*/
+
 package A_Star;
 
 import java.util.*;
--- a/DataStructures/Graphs/A_Star.java
+++ b/DataStructures/Graphs/A_Star.java
@@ -27,49 +27,6 @@ public class A_Star {
             this.graph[edge.getFrom()].add(new Edge(edge.getFrom(), edge.getTo(), edge.getWeight()));
             this.graph[edge.getTo()].add(new Edge(edge.getTo(), edge.getFrom(), edge.getWeight()));
         }
-
-        private void initializeGraph() {
-            this.addEdge(new Edge(0, 19, 75));  this.addEdge(new Edge(0, 15, 140));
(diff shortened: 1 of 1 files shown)


DataStructures/Graphs/A_Star.java
```diff
@@ -0,0 +1,162 @@
...
+
+        private ArrayList<Edge> getNeighbours(int from) { return this.graph[from]; }
+
+        //Graph is bidirectional, for just one direction remove second instruction of this method.
+        private void addEdge (Edge edge) {
+            this.graph[edge.getFrom()].add(new Edge(edge.getFrom(), edge.getTo(), edge.getWeight()));
+            this.graph[edge.getTo()].add(new Edge(edge.getTo(), edge.getFrom(), edge.getWeight()));
+        }
+
+        private void initializeGraph() {
```
luislizarraga, line 31: I think that this initialization should be outside of the class, in the `main` function, as if I wanted to use the A* class as a package inside a program, I wouldn't be able to add Edges to a graph

DataStructures/Graphs/A_Star.java
```diff
@@ -0,0 +1,162 @@
...
+                System.out.println("Optimal path: " + this.path.toString() +
+                        ", distance: " + this.distance);
+            else
+                System.out.println("There is no path available to connect the points");
+        }
+    }
+
+    public static void main(String[] args) {
+        //heuristic function optimistic values
+        int[] heuristic = {366, 0, 160, 242, 161, 178, 77, 151, 226,
```
luislizarraga, line 122: What does this array represent?

MarcosVillacanas: The estimated values calculated to reach point x from point y. This algorithm needs a collection of values like this to make decisions. It's up to the user the way in which those values are calculated.

DataStructures/Graphs/A_Star.java
MarcosVillacanas: I have fixed that, thanks for your time

PR #1351, opened 2020-06-17
Ford Fulkerson Algorithm

oldMarcosVillacanas:
Code is explained from top to bottom, surrounded by six different examples that can be selected from the main procedure. Please if you have any questions or doubts ask them, I will answer you as soon as possible.

### DIFF
--- a/DataStructures/Graphs/Ford_Fulkerson_Max_Flow.java
+++ b/DataStructures/Graphs/Ford_Fulkerson_Max_Flow.java
@@ -1,0 +1,135 @@
+package Ford_Fulkerson_Max_Flow;
+
+import java.util.*;
+
+public class Ford_Fulkerson_Max_Flow {
+
+    private static int graphNodes;
+    private static int source;
+    private static int sink;
+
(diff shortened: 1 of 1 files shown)


DataStructures/Graphs/Ford_Fulkerson_Max_Flow.java
```diff
@@ -0,0 +1,137 @@
...
+
+    private static ArrayList<Integer> pickExample(int choice) {
+        //from, to, capacity, null (to visually separate edges)
+        switch (choice) {
+            case 0: {
+                //From 0 (source) to 10 (sink) should return 7
+                source = 0;
+                sink = 10;
+                graphNodes = 11;
+                return new ArrayList<>(Arrays.asList(0, 1, 7, null, 0, 2, 2, null, 0, 3, 1, null, 1, 4, 2,
```
luislizarraga, line 80: Why do you add a null value in the `i+3` position?

MarcosVillacanas: Because with that null value, the code is easier to read.
Graph edges are set in groups of three values, from, to and weight. 
Null values separate those edges, and programmers don't lose time trying to understand how many edges are and which values conform them.

PR #1355, opened 2020-06-30
Linked Hash Map Example

suranasourabh:


DataStructures/HashMap/Hashing/LinkedHashMap.java
```diff
@@ -0,0 +1,35 @@
+    import java.util.*;
```
deadshotsb, line 1: @suranasourabh Kindly provide a description and a sample case for your repository

PR #1363, opened 2020-07-11
format code

shellhub:


shellhub:
@yanglbme please review. thanks

Others/TowerOfHanoi.java
```diff
@@ -5,16 +5,13 @@
 class TowerOfHanoi {
     public static void shift(int n, String startPole, String intermediatePole, String endPole) {
         // if n becomes zero the program returns thus ending the loop.
-        if (n == 0) {
```
AnupKumarPanwar, line 8: Why are you removing the base case?

shellhub: recursion function will be terminated when `if (n != 0 )` failed

PR #1369, opened 2020-07-25
Added Count Digit program

Utsav1999:
Added a program that will **count the number of digits** of a user given number.

### DIFF
--- a/Maths/CountDigit.java
+++ b/Maths/CountDigit.java
@@ -1,5 +1,5 @@
 import java.util.*;
-import java.lang.*;
+package Maths;
 // count the number of digits in a number
 class CountDigit {
 	public static void main(String args[]) {
--- a/Maths/CountDigit.java
+++ b/Maths/CountDigit.java
@@ -1,0 +1,13 @@
+import java.util.*;
+import java.lang.*;
+// count the number of digits in a number
+class CountDigit{
+	public static void main(String args[]){
+		Scanner sc = new Scanner(System.in);
+		System.out.print("Enter the number: ");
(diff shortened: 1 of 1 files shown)


Maths/CountDigit.java
```diff
@@ -0,0 +1,13 @@
+import java.util.*;
+import java.lang.*;
+// count the number of digits in a number
+class CountDigit{
+	public static void main(String args[]){
+		Scanner sc = new Scanner(System.in);
+		System.out.print("Enter the number: ");
+		int number = sc.nextInt();
+		int digits = 0;
+		digits = (int)Math.floor(Math.log10(number) + 1);
```
deadshotsb, line 10: @Utsav1999  Its a good thinking but it has some limitations as it does not work for negetive numbers and zero paticularly because the log function is defined in (0,inf)

Utsav1999: Then I will add two if conditions, one for checking whether it is 0 or not and the other one for checking the negative number.

deadshotsb: You can try absolute fn and an if else in case of 0

Utsav1999: @deadshotsb I updated my code.

Maths/CountDigit.java
```diff
@@ -0,0 +1,20 @@
+import java.util.*;
```
StepfenShawn, line 1: Add `package Maths;` here.

Utsav1999: @StepfenShawn I updated my program with `package Maths;`

PR #1372, opened 2020-07-27
Added Java Program for Armstrong Number

Utsav1999:
I added a program to check  whether a number is **an Armstrong Number or not**.


rjtmahinay:
Hi @Utsav1999 this program already exist in https://github.com/TheAlgorithms/Java/blob/master/Others/Armstrong.java

Utsav1999:
Thank you @rjtmahinay for highlighting this. But this problem should be added to **Math** section.

PR #1382, opened 2020-08-01
Add HashMapLinearProbing.java and MainLinearProbing.java

### LINKED ISSUE #1380: Add Hashmap implementation with open addressing
benefits: faster than chaining when the load factor is low 

 could also add make bucket array larger functionality when load factor > .7

rbshealy:
closes #1380 
The class HashMapLinearProbing implements a hash map with open addressing functionality, covering delete, find, insert, and display.
The class MainLinearProbing is a driver that allows testing of the HashMapLinearProbing class.

### DIFF
--- a/DataStructures/HashMap/Hashing/HashMapLinearProbing.java
+++ b/DataStructures/HashMap/Hashing/HashMapLinearProbing.java
@@ -1,0 +1,163 @@
+package DataStructures.HashMap.Hashing;
+
+/**
+ * This class is an implementation of a hash table using linear probing
+ *
+ */
+class HashMapLinearProbing {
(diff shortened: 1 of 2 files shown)


DataStructures/HashMap/Hashing/HashMapLinearProbing.java
```diff
@@ -0,0 +1,163 @@
+package DataStructures.HashMap.Hashing;
+
+/**
+ * This class is an implementation of a hash table using linear probing
+ *
+ */
+class HashMapLinearProbing {
```
StepfenShawn, line 7: use `public class HashMapLinearProbing` :)

PR #1401, opened 2020-08-08
Created test class for Abecedarian class

NicklausKhaw:
This is a junit test for Abecedarian class


shellhub:
Please test with in main function

PR #1402, opened 2020-08-08
Test class for Palindrome class

StevenWong0121:
A test class for Palindrome class which is testing the method of ReverseString, FirstWay and isPalindrome


shellhub:
We just write test in single file which contains `main()`  .

Issue #1378, opened 2020-07-31
Add clearAll and index based insertion in DoublyLinkedList and fixed bugs


Zineeddine998:
#1378

Zineeddine998:
#1379

rnitish:
#1410

PR #1410, opened 2020-08-11
Added insert element by index and clearList to DoublyLinkedList.java

rnitish:
I have added two methods:

1. insertElementByIndex - add an element to specified position.
2. clearList - clears the DoublyLinkedList.

Fixed some errors

closes issue #1378

### DIFF
--- a/DataStructures/Lists/DoublyLinkedList.java
+++ b/DataStructures/Lists/DoublyLinkedList.java
@@ -25,11 +25,17 @@ public class DoublyLinkedList {
     private Link tail;
 
     /**
+     * Size refers to the number of elements present in the list
+     */
+    private int size;
+
+    /**
      * Default Constructor
      */
     public DoublyLinkedList() {
         head = null;
         tail = null;
+        size = 0;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


DataStructures/Lists/DoublyLinkedList.java
```diff
@@ -43,6 +49,7 @@ public DoublyLinkedList(int[] array) {
         for (int i : array) {
             insertTail(i);
         }
+        size = array.length;
```
rbshealy, line 52: If size > 0 before this function is called, size != array.length

rbshealy: use size += array.length

rnitish: Thank you @rbshealy  for reviewing, I will make the required changes.

rbshealy: Yes this is correct, didn't realize this is the constructor. Great work!

rnitish: Thank you

PR #1417, opened 2020-08-16
Add PythagoreanTriple.java and Add assert statements

rbshealy:
PythagoreanTriple.java is a function that determines whether its inputs, a,b,c are a Pythagorean triple.
returns true if a^2 + b^2 = c^2 and all are positive non-zero values, false otherwise

Adding assert statements to AbsoluteMax, AbsoluteMin, AbsoluteValue, MinValue, and MaxValue helps for testing

### DIFF
--- a/Maths/AbsoluteMax.java
+++ b/Maths/AbsoluteMax.java
@@ -10,10 +10,10 @@ import java.util.Arrays;
  */
 public class AbsoluteMax {
     public static void main(String[] args) {
-    	int[] testnums = new int[]{-2, 0, 16};
+    	int[] testnums = {-2, 0, 16};
     	assert absMax(testnums) == 16;
     	
-        int[] numbers = new int[]{3, -10, -2};
+        int[] numbers = {3, -10, -2};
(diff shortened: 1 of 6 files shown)


Maths/AbsoluteMax.java
```diff
@@ -10,6 +10,9 @@
  */
 public class AbsoluteMax {
     public static void main(String[] args) {
+    	int[] testnums = new int[]{-2, 0, 16};
```
StepfenShawn, line 13:
```java
int[] testnums = {-2, 0, 16};
```

Maths/AbsoluteMin.java
```diff
@@ -10,6 +10,9 @@
  */
 public class AbsoluteMin {
     public static void main(String[] args) {
+    	int[] testnums = new int[]{4, 0, 16};
```
StepfenShawn, line 13:
```java
int[] testnums = {4, 0, 16};
```

weXiao-a:
Thank you so much!

PR #1423, opened 2020-08-18
update average.java

drabbit75:
1. add int type array
2. The equality of floating-point numbers should be judged by the absolute value of the difference less than a small number


Maths/Average.java
```diff
@@ -4,10 +4,13 @@
...
+    private static final double SMALL_VALUE = 0.00001f;
     public static void main(String[] args) {
-        assert average(new double[]{3, 6, 9, 12, 15, 18, 21}) == 12;
-        assert average(new double[]{5, 10, 15, 20, 25, 30, 35}) == 20;
-        assert average(new double[]{1, 2, 3, 4, 5, 6, 7, 8}) == 4.5;
+        assert Math.abs(average(new double[]{3, 6, 9, 12, 15, 18, 21}) - 12) < SMALL_VALUE;
+        assert Math.abs(average(new double[]{5, 10, 15, 20, 25, 30, 35}) - 20) < SMALL_VALUE;
+        assert Math.abs(average(new double[]{1, 2, 3, 4, 5, 6, 7, 8}) - 4.5) < SMALL_VALUE;
+        int[] array = {2, 4, 10};
+        assert average(array) == 5;
```
StepfenShawn, line 13:
```java
 assert average(new int[]{2, 4, 10}) == 5;
```

drabbit75: it is also ok

PR #1424, opened 2020-08-19
Create Range.java

gijsh21:
For finding the range of an array of numbers


drabbit75:
why do you not use template class,

gijsh21:
As in a generic class? I don't think they work for primitives, at least I can't seem to get them to work.

drabbit75:
en, I just make  a mistake

PR #1495, opened 2020-09-30
Create QuickSelect.java

sharduln30:
* [x] Add an algorithm
* [ ] Fix a bug or typo in an existing algorithm?
Added a new problem on Sorting named "QuickSelect.java"


sharduln30:
Hi! Please review the problem and revert me if there is any issue.

PR #1500, opened 2020-09-30
Improved Sorting Example BubbleSort.java

saurab2019:
Hi,
I think for explaining sorting algorithm , if we have the array in decreasing order intially and then we arranged in increasing order is more useful in understanding the time complexity of different algorithms

### DIFF
--- a/Sorts/BubbleSort.java
+++ b/Sorts/BubbleSort.java
@@ -37,11 +37,11 @@ class BubbleSort implements SortAlgorithm {
     public static void main(String[] args) {
 
         // Integer Input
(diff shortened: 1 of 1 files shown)


saurab2019:
thanks for explaining these algorithms

Saby702 (approved):

deadshotsb:
I have labelled your pr as invalid due to two reasons:-

1. Minor text edits are not entertained during hacktoberfest

2. In bubble sort algo irrespective of the order of input pattern the algo will compare each and every adjascent pairs, so there is no such worst and best case it's always O(n2)

PR #1727, opened 2020-10-10
Generate no.

alix30:
Factors of a Positive Integer

### **Describe your change:**

* [ ] Add an algorithm?
* [ ] Fix a bug or typo in an existing algorithm?
* [ ] Documentation change?

### **Checklist:**

* [ ] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
* [ ] This pull request is all my own work -- I have not plagiarized.
* [ ] I know that pull requests will not be merged if they fail the automated tests.
* [ ] This PR only changes one algorithm file.  To ease review, please open separate PRs for separate algorithms.
* [ ] All new Java files are placed inside an existing directory.
* [ ] All filenames are in all uppercase characters with no spaces or dashes.
* [ ] All functions and variable names follow Java naming conventions.
* [ ] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
* [ ] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.

### DIFF
--- a/Factors of a Positive Integer
+++ b/Factors of a Positive Integer
@@ -1,0 +1,20 @@
+public class Main {
+
+  public static void main(String[] args) {
+
+    // positive number
+    int number = 60;
+
+    System.out.print("Factors of " + number + " are: ");
+
+    // loop runs from 1 to 60
+    for (int i = 1; i <= number; ++i) {
+
+      // if number is divided by i
+      // i is the factor
+      if (number % i == 0) {
+        System.out.print(i + " ");
+      }
+    }
+  }
+}


SethiPandi:
This loop does not run from 1 to 60, it runs from 2 to 61. You probably meant i++, not ++i.

Also you don't have to bother iterating i == 1, and i == X because 1 and X will always be factors.

I think this would be better if this only looped until half of X and instead and printed both i and the result of the division?

For example: X is 60
Already print 1, and 60.
X is 60, so loop from 2, to 30.
If i % number == 0, print both i and number / i.

The logic might not be perfect since it was only in my head, but something along those lines should work and require significantly less iterations.

PR #1864, opened 2020-10-18
Adding a CheckSubstrings.java file to new_branch

AruneshNaha:
### **Describe your change:**

* [ ] Check substring in a string

### DIFF
--- a/strings/CheckSubstrings.java
+++ b/strings/CheckSubstrings.java
@@ -1,0 +1,30 @@
+package strings;
+
+import java.util.Arrays;
+
+/**
+ * This program checks if one string is the substring of the other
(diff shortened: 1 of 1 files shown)


SethiPandi:
I don't think this is a particularly useful addition to the repository.

And I don't think this is even considered an "algorithm" really, it's just two standard Java methods, toLowerCase and contains. 

The method names are misleading, a better name would be `containsIgnoreCase`. (Like what `equalsIgnoreCase` is to `equals`.)
There is also no need to initialize the boolean if you'll assign it a new value right after.

The entire method could've been written as:

```java
return a.toLowerCase().contains(b.toLowerCase());
```

PR #1910, opened 2020-10-22
Added Subset of Sum oneof Dynamic programming

Ash515:
Given a set of non-negative integers, and a value sum, determine if there is a subset of the given set with sum equal to given sum.
Example:

Input: set[] = {3, 34, 4, 12, 5, 2}, sum = 9
Output: True  
There is a subset (4, 5) with sum 9.

Input: set[] = {3, 34, 4, 12, 5, 2}, sum = 30
Output: False
There is no subset that add up to 30.

### DIFF
--- a/DynamicProgramming/Subset of sum.java
+++ b/DynamicProgramming/Subset of sum.java
@@ -1,0 +1,44 @@
+// A recursive solution for subset sum 
+// problem 
+class GFG { 
+
+	// Returns true if there is a subset 
+	// of set[] with sum equal to given sum 
+	static boolean isSubsetSum(int set[], 
+							int n, int sum) 
+	{ 
+		// Base Cases 
+		if (sum == 0) 
+			return true; 
+		if (n == 0) 
+			return false; 
+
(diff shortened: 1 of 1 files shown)


DynamicProgramming/Subset of sum.java
```diff
@@ -0,0 +1,44 @@
...
+			|| isSubsetSum(set, n - 1, sum - set[n - 1]); 
+	} 
+
+	/* Driver program to test above function */
+	public static void main(String args[]) 
+	{ 
+		int set[] = { 3, 34, 4, 12, 5, 2 }; 
+		int sum = 9; 
+		int n = set.length; 
+		if (isSubsetSum(set, n, sum) == true)
```
prashantdoshi28, line 35: I think if you can perform following change, then it'd be more appropriate : 

```diff
- if(isSubsetSum(set, n, sum) == true)
+ if(isSubsetSum(set, n, sum)) 
```

As return type of `isSubsetSum()` is `boolean` , you can omit equals check.

PR #2017, opened 2020-11-04
Create FizzBuzz.java

coliner300:
Added FizzBuzz to the sort section

### DIFF
--- a/Sorts/FizzBuzz.java
+++ b/Sorts/FizzBuzz.java
@@ -1,0 +1,30 @@
+import java.util.*;
+class FizzBuzz
+{
+    public static void main(String args[])
+    { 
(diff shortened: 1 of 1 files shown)


SethFalco:
* This isn't a sorting algorithm.
* There is no need to have a `public static void main`, instead it'd be better to implement it as a generic function. For example, you could implement multiple methods and use parameters. This would make it much easier to unit test.
  * `String getFizzBuzz(int x)` // Returns Fizz, Buzz, or FizzBuzz
  * `String[] fizzBuzzForRange(int max)` // Loop 0 to `max` and call `getFizzBuzz(int)`

It might be better to make the PR to the Development branch since it has a better structure and place for unit tests.

PR #2022, opened 2020-11-05
ReverseWord & ReverseWordsTest

loveshdongre:
### **Describe your change:**

- [x] Add an algorithm?
- [ ] Documentation change?

### **Checklist:**

- [x] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.

### DIFF
--- a/src/main/java/com/string/ReverseWords.java
+++ b/src/main/java/com/string/ReverseWords.java
@@ -7,8 +7,8 @@ public class ReverseWords {
    * @param s the string to convert
    * @return the {@code String}, converted to a string with reveresed words.
    */
-  
-  public static String returnReverseWords(String s) {
+
+  public String returnReverseWords(String s) {
     StringBuilder sb = new StringBuilder();
     StringBuilder word = new StringBuilder();
 
--- a/src/main/java/com/string/ReverseWords.java
+++ b/src/main/java/com/string/ReverseWords.java
@@ -1,4 +1,4 @@
-package strings;
+package com.strings;
 
 public class ReverseWords {
   /**
--- a/src/main/java/com/string/ReverseWords.java
+++ b/src/main/java/com/string/ReverseWords.java
@@ -1,0 +1,29 @@
+package strings;
+
+public class ReverseWords {
+  /**
+   * Converts all of the words in this {@code String} to reversed words
+   *
+   * @param s the string to convert
+   * @return the {@code String}, converted to a string with reveresed words.
+   */
+  public static String returnReverseWords(String s) {
+    StringBuilder sb = new StringBuilder();
+    StringBuilder word = new StringBuilder();
+
+    for(int i = 0; i < s.length(); i++) {
+        char c = s.charAt(i);
+        if(c == ' ') {
+          
+            sb.append(word);
+            sb.append(" ");
+            word.setLength(0);
+            continue;
+        }
+        word.insert(0, c);
+    }
+    sb.append(word);
+
+    return sb.toString();
+  }
+}
--- a/src/test/java/com/string/ReverseWordsTest.java
+++ b/src/test/java/com/string/ReverseWordsTest.java
@@ -1,0 +1,14 @@
+package com.string;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class ReveresWordsTest {
+
+    @Test
+    void testAlphabetical() {
+        ReverseWords reverseWords = new ReverseWords();
+        Assertions.assertEquals(true, returnReverseWords("this is my car"), "siht si ym rac");
+        Assertions.assertEquals(true, returnReverseWords("ABC 123"), "CBA 321");
+    }
+}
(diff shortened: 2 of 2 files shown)


src/test/java/com/string/ReverseWordsTest.java
```diff
@@ -0,0 +1,13 @@
+package com.string;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class ReveresWordsTest extends ReverseWords {
```
prashantdoshi28, line 6: This is not proper way to write tests. 

If you don't know how a test is to be written, please check [DecimalToAnyBaseTestjava](https://github.com/TheAlgorithms/Java/blob/Development/src/test/java/com/conversions/DecimalToAnyBaseTest.java). 

As you can see, it does not extend its source class. Instead it initializes its object with help of constructor. 

Please do needful.

prashantdoshi28 (changes requested):

src/main/java/com/string/ReverseWords.java
```diff
@@ -0,0 +1,29 @@
+package strings;
+
+public class ReverseWords {
+  /**
+   * Converts all of the words in this {@code String} to reversed words
+   *
+   * @param s the string to convert
+   * @return the {@code String}, converted to a string with reveresed words.
+   */
+  public static String isReverseWords(String s) {
```
prashantdoshi28, line 10: Method name is kind of not in sync with what it does, i.e. 

```java
 public static String isReverseWords(String s) {...} // Suggests that this will return boolean value. 

 public static String returnReverseWords(String s) {...} // Suggests that this will return reversed string value. 
```

**NOTE** : This is just an observation or suggestion.

loveshdongre: I renamed it `isReverseWords` to `returnRevereseWords`

src/test/java/com/string/ReverseWordsTest.java
loveshdongre: I removed the import line and inheritance from `ReverseWordsTest` and initialized an object of `ReverseWords` in it.

src/test/java/com/string/ReverseWordsTest.java
```diff
@@ -0,0 +1,15 @@
+package com.string;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class ReveresWordsTest {
+
+    @Test
+    void testAlphabetical() {
+        ReverseWords reverseWords = new ReverseWords();
```
prashantdoshi28, line 10: Good to see that you implemented test structure properly. But I'm not sure that this test will be passing, as you used `returnReverseWords()` directly without providing proper info to compiler about where this method will be present. Also in your case, as you have written 
```java
public static String returnReverseWords(String s) {...}
``` 
, you actually don't need an object to access static method. 
You can do following in test class
```java
Assertions.assertEquals(true, ReverseWords.returnReverseWords("this is my car"), "siht si ym rac");
```

Or you could remove `static` from method declaration and use object to access method like 
```java
Assertions.assertEquals(true, reverseWords.returnReverseWords("this is my car"), "siht si ym rac");
```

Please check these and do needful.

prashantdoshi28 (changes requested):

src/test/java/com/string/ReverseWordsTest.java
loveshdongre: I changed `testAlphabetical` to `testReverseWords`
and removed `static` keyword from `returnReverseWords`

After pushing these changes I am not able to pass 1 check which is `Java gradle CI / Test algorithms (macos-latest, 1.8) (pull_request)`. I don't know how to work with that.

prashantdoshi28: Hi @loveshdongre , I'm not sure of the issue, I will let you know if I find anything. If you find anything, please update over here so anyone else checking this issue can focus on something else.

loveshdongre: Yes sure

prashantdoshi28: Hey @ayaankhan98 , if you have some spare time, please check this issue 

> After pushing these changes I am not able to pass 1 check which is `Java gradle CI / Test algorithms (macos-latest, 1.8) (pull_request)`. I don't know how to work with that. Need help!

I'm also looking into but as I'm unable to view logs, I can't pinpoint exact error. But seeing console logs, I think issue is related to OpenGL with macOS (although only jdk8 is failing and not jdk11, so I can be wrong).  Thanks :)

loveshdongre: I changed `package strings;` to `package com.strings;` and now everything works. my bad.

prashantdoshi28: Yeah, I missed that too, good work.

loveshdongre: Should I close the pull request and wait for it to get merged?

ayaankhan98 (approved):

JackZeled0n:
I don't understand how you use the assertEquals. At the moment when I run the test java throw an exception because in your method return a string and then you compare with a boolean

PR #2039, opened 2020-11-20
Matrix Transpose algorithm Solved

Rajat-jain29:
Solved in Proper way

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,10 @@ gradle.properties
 .vscode
 
 *.log
+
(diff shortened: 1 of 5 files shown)


Rajat-jain29:
How to Solve Code formatter/format(pull_request) checks??

prashantdoshi28 (changes requested):
Please remove unnecessary files from PR. PR should only have files that actually depict your work in this case `*.java` files.

PR #2046, opened 2020-11-24
Implementation of dijkstra algorithm.

abdorah:
I want to add my own implementation of Dijkstra algorithm.
This implementation will be able to give a shortest path to any vertex of a graph by knowing only the source of the path.
The way it works is by giving a HashMap of "previous" which are basically all the visited vertexes.


abdorah:
I have implemented the code in this forked repo in the following link:
https://github.com/abdorah/Java

PR #2062, opened 2020-12-18
Patterns

ayushkumar0208:
Added 9 Patterns.


SethFalco:
What are these "Patterns"? Where do they come from? What's the purpose of them?

1. An algorithm doesn't need a `public static void main` it can run in a method and have unit tests instead. There's no need to make them interactive.
2. It's unclear what algorithm you're implementing if any, it'd be better if you could explain and add docs with links to what you're doing and what it's solving.

It could be nice to try to the PR to the Development branch instead since this is better structured, and has a place for unit tests.

PR #2180, opened 2021-04-13
Development test failure

mk-94:
### **Describe your change:**

- [ ] Add an algorithm?
- [X ] Fix a bug or typo in an existing algorithm?
- [X ] Documentation change?

#### References

Correct the test in UpperTest.java and delete comment. Test-cases are working now. 

### **Checklist:**

- [X] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [X] This pull request is all my own work -- I have not plagiarized.
- [X] I know that pull requests will not be merged if they fail the automated tests.
- [X] All new Java files are placed inside an existing directory.
- [X] All filenames are in all uppercase characters with no spaces or dashes.
- [X] All functions and variable names follow Java naming conventions.
- [X] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.
- [X] If this pull request resolves one or more open issues then the commit message contains `Fixes: #{$ISSUE_NO}`.


src/test/java/com/string/UpperTest.java
```diff
@@ -8,8 +8,8 @@
     @Test
     void testUpper() {
         Assertions.assertEquals(toUpperCase("abc"), ("abc").toUpperCase(), "The strings are equals");
-        //Assertions fail for functional reasons
-        Assertions.assertEquals(toUpperCase("abc"), "abc", "The strings are not equals");
+
+        Assertions.assertNotEquals(toUpperCase("abc"), "abc", "The strings are not equals");
```
nishantc1527, line 12: I think it would be useful if there would be a test something like:

```
Assertions.assertEquals(toUpperCase("abc"), "ABC", "The strings are equal");
```

Just to check if the string is correctly capitalized.

src/test/java/com/string/UpperTest.java
```diff
@@ -8,8 +8,8 @@
     @Test
     void testUpper() {
         Assertions.assertEquals(toUpperCase("abc"), ("abc").toUpperCase(), "The strings are equals");
```
nishantc1527, line 10: Instead of "The strings are equals" I think it should be "The strings are equal"

nishantc1527 (changes requested):

src/test/java/com/string/UpperTest.java
mk-94: I'll change it an make a new pull request. :)

mk-94:
I need some more details why the test did not pass. 
com.string.UpperTest > testUpper() FAILED
    org.opentest4j.AssertionFailedError at UpperTest.java:12
That does not make sense i think. Locally everything worked fine.

nishantc1527:
I'm also wondering why the tests don't work. I'm looking at it right now.

src/test/java/com/string/UpperTest.java
```diff
@@ -7,9 +7,9 @@
 
     @Test
     void testUpper() {
-        Assertions.assertEquals(toUpperCase("abc"), ("abc").toUpperCase(), "The strings are equals");
-        //Assertions fail for functional reasons
-        Assertions.assertEquals(toUpperCase("abc"), "abc", "The strings are not equals");
+        Assertions.assertEquals( "ABC",toUpperCase("abc"), "The strings are equal");
```
nishantc1527, line 10: I guess you can revert this assertion and we can figure it out later.

nishantc1527 (changes requested):

PR #2181, opened 2021-04-13
Formula to find median in Binary Search changed

Souravjyoti:
### **Describe your change:**

Finding median using the previous formula may carry the median value out of range. Therefore, updated the formula with a better one.

### **Checklist:**

### DIFF
--- a/Searches/BinarySearch.java
+++ b/Searches/BinarySearch.java
@@ -45,7 +45,8 @@ class BinarySearch implements SearchAlgorithm {
     if (right < left) return -1; // this means that the key not found
 
     // find median
-    int median = (left + right) >>> 1;
+    //int median = (left + right) >>> 1;
+    int median = left + ((right-left) >>> 1)
     int comp = key.compareTo(array[median]);
 
     if (comp == 0) {
--- a/Searches/BinarySearch.java
+++ b/Searches/BinarySearch.java
@@ -45,7 +45,7 @@ class BinarySearch implements SearchAlgorithm {
     if (right < left) return -1; // this means that the key not found
 
     // find median
-    int median = left + ((right-left) >>> 1)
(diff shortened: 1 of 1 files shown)


Searches/BinarySearch.java
```diff
@@ -45,7 +45,8 @@
     if (right < left) return -1; // this means that the key not found
 
     // find median
-    int median = (left + right) >>> 1;
+    //int median = (left + right) >>> 1;
+    int median = left + ((right-left) >>> 1)
```
shellhub, line 49:
```suggestion
    int median = left + ((right-left) >>> 1)
```

Searches/BinarySearch.java
```diff
@@ -45,7 +45,7 @@
     if (right < left) return -1; // this means that the key not found
 
     // find median
-    int median = (left + right) >>> 1;
+    int median = left + ((right-left) >>> 1)
```
shellhub, line 48:
```suggestion
    int median = left + ((right-left) >>> 1);
```

shellhub (approved):
Thanks for your contribution.

Searches/BinarySearch.java
```diff
@@ -45,7 +45,7 @@
     if (right < left) return -1; // this means that the key not found
 
     // find median
-    int median = (left + right) >>> 1;
+    int median = left + ((right-left) >>> 1);
```
nirvanarsc, line 48: Actually this change is not necessary. Above expression is already using a [logical shift](https://en.wikipedia.org/wiki/Logical_shift) (three `>>>`) which will preserve the sign bit if a potential overflow happens when summing (left + right). For languages like C++ and others that only support [arithmetic shift](https://en.wikipedia.org/wiki/Arithmetic_shift) (two `>>`) above workaround is necessary.

TL;DR original code works fine

PR #2182, opened 2021-04-14
BinarySearch: clearer median computation

allentiak:
">>> 1" is simply too obscure...

### **Describe your change:**

- [ ] Add an algorithm?
- [x] Fix a bug or typo in an existing algorithm?
- [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/Searches/BinarySearch.java
+++ b/Searches/BinarySearch.java
@@ -45,7 +45,7 @@ class BinarySearch implements SearchAlgorithm {
     if (right < left) return -1; // this means that the key not found
 
     // find median
-    int median = left + ((right - left) / 2);
+    int median = (left + right) >>> 1;
     int comp = key.compareTo(array[median]);
 
     if (comp == 0) {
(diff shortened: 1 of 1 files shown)


Searches/BinarySearch.java
```diff
@@ -45,7 +45,7 @@
     if (right < left) return -1; // this means that the key not found
 
     // find median
-    int median = left + ((right - left) >>> 1);
+    int median = left + ((right - left) / 2);
```
yanglbme, line 48:
```suggestion
    int median = (left + right) >>> 1;
```

yanglbme (changes requested):

Searches/BinarySearch.java
allentiak: What about `int median = (left + right) / 2;`?
After all, the whole issue here is the `>>> 1` part...

shellhub: When `left = Integer.MAX_VALUE - 1`, `right = Integer.MAX_VALUE`,  `int median = (left + right) / 2` will not work exactly.

shellhub (approved):

Issue #2212, opened 2021-05-21
Add Password Validator


esparev:
Validates if the entered password has at least one special character, one number, doesn't have spaces and has more than 5 but less than 20 characters in length.

### Case 1 (Invalid Password):
![image](https://user-images.githubusercontent.com/59494364/119076135-a991f200-b9b7-11eb-819b-66296875d391.png)

### Case 2 (Valid Password):
![image](https://user-images.githubusercontent.com/59494364/119076251-d940fa00-b9b7-11eb-9c5a-122522fcd700.png)

anishsaraogi:
I would like to work on this issue.

domonwa1:
Ok

ddiaaz:
Would these changes be on PasswordGen.java?

ddiaaz:
Would you like these changes on PasswordGen.java?

esparev:
Sure!, why not?

ddiaaz:
@esparev great. For the second case, are we checking if two characters are out of place and correcting it for the user? In other words, does the order of the characters matter?

esparev:
No, the order of the characters don't matter

Chprudhvi12:
Hi @esparev and @ddiaaz ,
Is this issue resolved?

PR #2215, opened 2021-05-22
2212

anishsaraogi:
Issue #2212 fixed and Password Validator added.

### DIFF
--- a/PasswordValidator.java
+++ b/PasswordValidator.java
@@ -4,41 +4,51 @@ class PasswordValidator
 {
   public static void main(String args[])
   {
-    PasswordValidator pv=new PasswordValidator();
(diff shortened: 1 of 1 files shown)


esparev:
Technically, it's not wrong, and the code wasn't broken, but it isn't working the way it's supposed to with the changes you made.

esparev:
I suggest you look again at the code that I have and I strongly recommend that you give the variables a more explicit and explanatory name.

anishsaraogi:
It's probably the special characters because as I saw, your code includes more special characters than I put. I will add it asap.

esparev:
Appreciate the attention, thank you @anishsaraogi

PR #2377, opened 2021-10-01
Create MergeSort.java

Shubham56-droid:
I have added a merge sort in java if you need any changes to it pls tell me I will do it.


siriak:
We already have a merge sort in the proper place, you either haven't even checked for it in sorts or tried to deceive reviewers by putting it in a wrong directory

PR #2395, opened 2021-10-01
Create CountWords

CuriousRadhika:
Java program to count number of words in the string input by the user

### DIFF
--- a/Strings/CountWords
+++ b/Strings/CountWords
@@ -1,0 +1,40 @@
+import java.util.Scanner;
+class CountWords
+{
+    public static void main(String[]args)
+    {
+        Scanner sc = new Scanner(System.in);
+        System.out.println("Enter any sentence");
+        String s = sc.nextLine();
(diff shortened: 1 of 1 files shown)


ritweekraj2802:
Respected sir , please assign me this problem.

Algorithm:
1.)Initialize count with 1 as if there are no spaces in the string, then there will be one word in the String.
2).Check if you encounter any space.
3).Once you find the space, check it next character. If it is not space then we found a word in the String.Increment the count variable.
4).Once you reach end of String, count variable will hold number of words in the String.

code:

public class CountNumberOfWordsInStringMain {
 
        String str = "please assign me this problem";
 
        int count = 1;
 
        for (int i = 0; i < str.length() - 1; i++)
        {
            if ((str.charAt(i) == ' ') && (str.charAt(i + 1) != ' '))
            {
                count++;
            }
        }
        System.out.println("Number of words in a string : " + count);
    }
}

PR #2460, opened 2021-10-03
Print all possible words from phone digits

ritweekraj2802:
### **Describe your change:**

Algorithm:  

1 )Map the number with its string of probable alphabets, i.e 2 with “abc”, 3 with “def” etc.
2 )Create a recursive function which takes the following parameters, output string, number array, current index, and length of number array
3)If the current index is equal to the length of the number array then print the output string.
4 )Extract the string at digit[current_index] from the Map, where the digit is the input number array.
5 )Run a loop to traverse the string from start to end
6 )For every index again call the recursive function with the output string concatenated with the ith character of the string and the current_index + 1.
#### References

https://stackoverflow.com/questions/2344496/how-can-i-print-out-all-possible-letter-combinations-a-given-phone-number-can-re/24484148

### **Checklist:**

### DIFF
--- a/Others/Implementing_auto_completing_features_using_trie.java
+++ b/Others/Implementing_auto_completing_features_using_trie.java
@@ -1,0 +1,183 @@
+package Others;
+
+
+// Java Program to implement Auto-Complete 
+// Feature using Trie
+
+import java.util.*;
+import java.io.*;
+import java.lang.*;
+
+class Trieac {
+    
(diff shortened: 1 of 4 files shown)


ritweekraj2802:
Please accept my PR
Thank You

siriak:
We don't add interview questions, only well-known problems, thanks!

PR #2496, opened 2021-10-05
added ML algo written from scratch

par1hsharma:
added ML algo written from scratch


siriak:
Because we welcome machine learning algorithms, but this is JAVA repository and you wrote it in PYTHON. Those are different languages

PR #2503, opened 2021-10-06
Create Parallel_vs_Sequential _Sort_of_N_Arrays.java

ShubhanshuJha:
-> Sorting N Arrays using Sequential vs Multi-Threading approach together with Arrays.sort().
This approach will be taking very little time than a normal sorting approach for a 2-D array, as each array will be sorted separately (parallelly) on a thread.
There are many alternate ways of sorting the arrays in the 2D array, but, this one is good if you want to implement your sorting algorithm on each array of a 2D array, and to achieve that just replace the Arrays.sort() with your algorithm initializing method.

So, please do add this file because this would be helpful for many people out there.

#This file covers the objective of:
-> Sorting 2D arrays
-> Parallel sorting approach
-> Testing the algorithm for extensively large arrays by just putting the size of the array
-> Displaying the time consumed for both sequential sorting approach and parallel sorting approach
-> Easily modifiable code

### DIFF
--- a/Sorts/Sort_N_Array.java
+++ b/Sorts/Sort_N_Array.java
@@ -1,0 +1,92 @@
+##Sorting N Arrays using Multi-Threading approach and Arrays.sort()
+This approach will be take very less time than a normal sorting approach for a 2-D array.
+There are many alternate ways of sorting the arrays in the 2D-array, but, this one is good if you want to implement your own sorting algorithm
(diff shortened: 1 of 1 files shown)


ShubhanshuJha:
I've submitted an approach on how to do sorting of N arrays using Sequential vs Multi-Threading approach using Arrays.sort() method. There are various concepts implemented in the code that would be helpful for many people, so please add the file.

PR #2585, opened 2021-10-12
Vertical Order Traversal in a tree added

Pramil01:
### **In DataStructure folder under tree subcategory vertical order traversal added.**
Vertical order traversal in a tree prints the tree from left to right and top to bottom in vertical columns , the codes not only implements it but also explains the process and what data structure is used for what purpose .


Pramil01:
The changes asked are done .

PR #2663, opened 2021-10-17
Create decode-ways.java

dev7omprakash:
To decode an encoded message, all the digits must be grouped then mapped back into letters using the reverse of the mapping above (there may be multiple ways). For example, "11106" can be mapped into:

"AAJF" with the grouping (1 1 10 6)
"KJF" with the grouping (11 10 6)
Note that the grouping (1 11 06) is invalid because "06" cannot be mapped into 'F' since "6" is different from "06".

Given a string s containing only digits, return the number of ways to decode it.

Example 1:

Input: s = "12"
Output: 2
Explanation: "12" could be decoded as "AB" (1 2) or "L" (12).

### **Describe your change:**

- [ ] Add an algorithm?
- [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/DynamicProgramming/decode-ways.java
+++ b/DynamicProgramming/decode-ways.java
@@ -1,0 +1,43 @@
+class Solution {
+    public int numDecodings(String s) {
+      
+        int dp[]=new int[s.length()+1];
+        Arrays.fill(dp,-1);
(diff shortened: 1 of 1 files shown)


siriak:
We don't add leetcode problems, thanks!

PR #2697, opened 2021-10-20
Create ReverseArray.java

fahamabbasi:


ramdcoder:
`
import java.util.Arrays;
public class ReverseWithoutArray
{
   public int[] arrayReverse(int[] arrInput)
   {
      if(arrInput == null || arrInput.length <= 1)
      {
         System.out.println("Please enter valid array");
      }
      for(int a = 0; a < arrInput.length / 2; a++)
      {
         int temp = arrInput[a];
         arrInput[a] = arrInput[arrInput.length - 1 - a];
         arrInput[arrInput.length - 1 - a] = temp;
      }
      return arrInput;
   }
   public static void main(String[] args)
   {
      ReverseWithoutArray obj = new ReverseWithoutArray();
      int[] arr = {2, 4, 6, 8, 10, 12, 14, 16};
      System.out.println("Given array: " + Arrays.toString(arr));
      System.out.println("Reversed array: " + Arrays.toString(obj.arrayReverse(arr)));
   }
}
`

PR #2770, opened 2021-10-30
Matrix chain memorization added

lapy113:
### **Describe your change:**

Added Matrix chain Multiplication using memorization technique


lapy113:
Found Little error

PR #2773, opened 2021-10-30
Create EqualRectangles.java

sharduln30:
You are given 4n sticks, the length of the i-th stick is ai.

You have to create n rectangles, each rectangle will consist of exactly 4 sticks from the given set. The rectangle consists of four sides, opposite sides should have equal length and all angles in it should be right. Note that each stick can be used in only one rectangle. Each stick should be used as a side, you cannot break the stick or use it not to the full length.

You want to all rectangles to have equal area. The area of the rectangle with sides a and b is a⋅b.

Your task is to say if it is possible to create exactly n rectangles of equal area or not.

You have to answer q independent queries.

Input
The first line of the input contains one integer q (1≤q≤500) — the number of queries. Then q queries follow.

The first line of the query contains one integer n (1≤n≤100) — the number of rectangles.

The second line of the query contains 4n integers a1,a2,…,a4n (1≤ai≤104), where ai is the length of the i-th stick.

Output
For each query print the answer to it. If it is impossible to create exactly n rectangles of equal area using given sticks, print "NO". Otherwise print "YES".

Example
input
5
1
1 1 10 10
2
10 5 2 10 1 1 2 5
2
10 5 1 10 5 1 1 1
2
1 1 1 1 1 1 1 1
1
10000 10000 10000 10000
output
YES
YES
NO
YES
YES

### **Describe your change:**

- [ ] Add an algorithm?
- [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/Maths/EqualRectangles.java
+++ b/Maths/EqualRectangles.java
@@ -1,0 +1,60 @@
+// You are given 4n sticks, the length of the i-th stick is ai.
+
+// You have to create n rectangles, each rectangle will consist of exactly 4 sticks from the given set. The rectangle consists of four sides, opposite sides should have equal length and all angles in it should be right. Note that each stick can be used in only one rectangle. Each stick should be used as a side, you cannot break the stick or use it not to the full length.
+
+// You want to all rectangles to have equal area. The area of the rectangle with sides a and b is a⋅b.
+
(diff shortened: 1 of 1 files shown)


siriak:
We don't add problems from Codeforces and similar websites, thanks!

---
**Sema Reaction:** :hammer_and_wrench: This code needs a fix

PR #2792, opened 2021-10-31
Bottom view for tree

lapy113:
### **Describe your change:**

Added bottom view  for tree in java


siriak (changes requested):
We already matrix the multiplication problem https://github.com/TheAlgorithms/Java/blob/5f59cd85d64ee074e13c5b14366f3df00c3227ff/DynamicProgramming/MatrixChainMultiplication.java

---
**Sema Reaction:** :hammer_and_wrench: This code needs a fix

PR #2795, opened 2021-10-31
Add changes to non-java semantics and to some loops

ivashchenko5:
### **Describe your change:**

- [+] Fix a bug or typo in an existing algorithm?

### **Checklist:**

### DIFF
--- a/Sorts/CircleSort.java
+++ b/Sorts/CircleSort.java
@@ -9,7 +9,6 @@ public class CircleSort implements SortAlgorithm {
     @Override
     public <T extends Comparable<T>> T[] sort(T[] array) {
         int n = array.length;
-        while(doSort(array, 0, n - 1));
         return array;
     }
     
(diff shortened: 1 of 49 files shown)


Sorts/CircleSort.java
```diff
@@ -9,7 +9,6 @@
     @Override
     public <T extends Comparable<T>> T[] sort(T[] array) {
         int n = array.length;
-        while(doSort(array, 0, n - 1));
```
siriak, line 12: What's up with that?

---
**Sema Reaction:** :question: I have a question

q
```diff
@@ -0,0 +1,2508 @@
+* [31m9e36810[m -[33m (HEAD -> master, origin/master, origin/HEAD)[m Add surface area of Cone and Hemisphere (#2791) [32m(5 hours ago) [1;34m<Vɪɴᴀʏᴀᴋ Pᴀɴᴅᴇʏ>[m
```
siriak, line 1: Please remove that

siriak (changes requested):
@yanglbme @aitorfi what do you think? Should we merge this PR?

---
**Sema Reaction:** :question: I have a question

Sorts/CircleSort.java
aitorfi: That's not right, you need a loop that calls doSort.

aitorfi (commented):
The loop in Sorts/CircleSort can't be removed and I don't think it's necessary to make all these variables final, everything else looks good to me.

PR #2824, opened 2021-11-16
Adds factor class in math folder

omarshall02:
Created a factor class in the math folder


siriak:
Hi, the naive factoring algorithm is quite simple, we don't need to include it, thanks!

PR #2842, opened 2021-11-24
fix bug of A_Star.Graph constructor case in jdk17

R-juzi:
fix bug of A_Star.Graph constructor case in jdk17


siriak (approved):
LGTM, thanks!

PR #2847, opened 2021-11-27
refactoring : UnionFind.java

namookk:
### **Describe your change:**

- [ ] Add an algorithm?
- [ ] Documentation change?

### **Checklist:**

### DIFF
--- a/src/main/java/com/thealgorithms/searches/UnionFind.java
+++ b/src/main/java/com/thealgorithms/searches/UnionFind.java
@@ -4,49 +4,29 @@ import java.util.*;
 
 public class UnionFind {
 
-    private int[] p;
-    private int[] r;
+    //private int[] p;
+    //private int[] r;
+    private int[] unf;
 
     public UnionFind(int n) {
-        p = new int[n];
-        r = new int[n];
-
-        for (int i = 0; i < n; i++) {
-            p[i] = i;
-        }
+        unf = new int[n+1];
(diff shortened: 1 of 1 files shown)


yanglbme:
I think the following code would be better. It's the union-find template.

```java
public class UnionFind {
    private int[] p;

    public UnionFind(int n) {
        p = new int[n];
        for (int i = 0; i < n; i++) {
            p[i] = i;
        }
    }

    public int find(int x) {
        if (p[x] != x) {
            p[x] = find(p[x]);
        }
        return p[x];
    }

    public void union(int a, int b) {
        p[find(a)] = find(b);
    }

    public int count() {
        int cnt = 0;
        for (int i = 0; i < p.length; ++i) {
            if (i == find(i)) {
                ++cnt;
            }
        }
        return cnt;
    }

    @Override
    public String toString() {
        return "p " + Arrays.toString(p);
    }

    // Tests
        UnionFind uf = new UnionFind(5);
        uf.union(1, 2);

        // true
        System.out.println(uf.find(1) == uf.find(2));

        // 4
        System.out.println(uf.count());

        uf.union(3, 4);

        // false
        System.out.println(uf.find(1) == uf.find(3));

        // true
        System.out.println(uf.find(3) == uf.find(4));

        // 3
        System.out.println(uf.count());
    }
}
```

siriak (changes requested):
Please use PascalCase for method naming and avoid using contracted variable names in favor of full versions

PR #2909, opened 2022-01-18
Added sort linked list

mikeo9911:
### **Describe your change:**

- [x] Add an algorithm?


siriak (changes requested):
Thanks for contributing, there're some sorts for linked lists already implemented here https://github.com/TheAlgorithms/Java/blob/857c4aafb27f2d1a3db41dd393c90e07a9d656b4/src/main/java/com/thealgorithms/sorts/LinkList_Sort.java Maybe you could improve them? It's not worth duplicating algorithms

PR #2953, opened 2022-02-21
feat: Add coverage measurement for MaximumMinimumWindow.calculateMaxO…

chubretovic:
Manual coverage measurement for the function calculateMaxOfMin() of the class MaximumMinimumWindow.


siriak:
Thanks for the suggestion, but I'd prefer something more automated

PR #3194, opened 2022-07-16
SubSet(SkienaForm) including process solution , is a solution and con…

Jyo561:
I added  backtracking algorithm for SubSet which is in sequence with the way of Skiena Book containing Construct Candidate, is a solution and process solution

### DIFF
--- a/src/main/java/com/thealgorithms/backtracking/subset(Skienaform).java
+++ b/src/main/java/com/thealgorithms/backtracking/subset(Skienaform).java
@@ -1,0 +1,46 @@
+package com.thealgorithms.backtracking;
+
+class Subset{
+	static void Backtracking(int a[],int k,int n,int input[]){
+		int c[]=new int[n];
+		int ncan=0;
+		if(is_a_solution(n,k)){
+			process_solution(a,n,k,input);
+		}
+		else{
+			k++;
+			ncan=construct_candidate(c,k,n);
+			for(int i=0;i<ncan;i++) {
+				a[k-1]=c[i];
+				Backtracking(a,k,n,input);
+			}
+		}
+	} 
+	static int construct_candidate(int c[],int k,int n){
+		c[0]=0;
+		c[1]=1;
+		return 2;
+	}
+	static boolean is_a_solution(int k,int n){
+		if(k-1==(n-1)){
+			return true;
+		}
+		else{
+			return false;
+		}
+	}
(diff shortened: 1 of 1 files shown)


src/main/java/com/thealgorithms/backtracking/subset(Skienaform).java
```diff
@@ -0,0 +1,46 @@
...
+			}
+		}
+		System.out.println();
+	}
+	public static void main(String[] args) {
+		int input[]={2,4,3,1,1};
+		int n=input.length;
+		int a[]=new int[n];
+		Backtracking(a,0,n,input);
+	}
```
siriak, line 45: Please replace the `main()` with automated tests

src/main/java/com/thealgorithms/backtracking/subset(Skienaform).java
```diff
@@ -0,0 +1,46 @@
...
+		c[1]=1;
+		return 2;
+	}
+	static boolean is_a_solution(int k,int n){
+		if(k-1==(n-1)){
+			return true;
+		}
+		else{
+			return false;
+		}
```
siriak, line 30:
```suggestion
		return k-1==(n-1);
```

src/main/java/com/thealgorithms/backtracking/subset(Skienaform).java
```diff
@@ -0,0 +1,46 @@
+package com.thealgorithms.backtracking;
+
+class Subset{
+	static void Backtracking(int a[],int k,int n,int input[]){
```
siriak, line 4: Please give descriptive names to the parameters instead of `a`, `k`, `n`, and `input`. And the function name needs to be changed as well.

siriak (changes requested):
The file name cannot contain braces

src/main/java/com/thealgorithms/backtracking/subset(Skienaform).java
Jyo561: But the function names are same as of skiena pseudocode which I was trying to replicate

siriak: If I want to use this function, I wouldn't know what `Subset.Backtracking` does and what are `a, k, n, input` that it accepts. Function signature must be self-explanatory.

siddhant2002:
Can I work under this issue?

PR #3220, opened 2022-08-17
Modified the code for Seive of Eratosthenes

Shashwat79802:
Simplified the code, and added the documentation where I improvised

The following changes were made :

1. The previous code used .fill method of the Arrays class which uses an internal for loop to initialize an array with the provided value, but this code doesn't need to initialize with any value as it works with the default value which is initialized to the array.
2. Previous code used 2 arrays: an array of type Type named numbers and an array of type int named prime. But this code uses only one array of type boolean named notPrime.
3. Some more in-built methods were used like .stream, .filter, and .count which are not needed in this code.

I hope that the points I've listed helps to understand my code.


siriak:
I like the previous implementation more because it does what it's expected to do, but you have moved the filtration of primes to `main()` and accepting a buffer and modifying it is confusing

PR #3223, opened 2022-08-18
added understandable comments and Deque.

hirentimbadiya:
hi! i have added **easy to understand comments** and changed the stack into Deque which uses **ArrayDeque** internally which is dynamically and **more preferrable than stack** , thanks!!

### DIFF
--- a/src/main/java/com/thealgorithms/datastructures/stacks/InfixToPostfix.java
+++ b/src/main/java/com/thealgorithms/datastructures/stacks/InfixToPostfix.java
@@ -1,6 +1,6 @@
 package com.thealgorithms.datastructures.stacks;
 
-import java.util.Stack;
+import java.util.Deque;
 
 public class InfixToPostfix {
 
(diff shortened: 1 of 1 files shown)


mohitchakraverty:
Why did you replace stack with deque? Infix to postfix is a typical application of stack, as the working fits the definition of stack wherein the operator and parenthesis are added to the stack, maintaining their precedence, and then pop out in the order following LIFO and thus forming a postfix expression.

hirentimbadiya:
> Why did you replace stack with deque? Infix to postfix is a typical application of stack, as the working fits the definition of stack wherein the operator and parenthesis are added to the stack, maintaining their precedence, and then pop out in the order following LIFO and thus forming a postfix expression.

you are right but using Stack internally uses the array-indexes while ArrayDeque & Deque are interface which are faster and preffered to use than stack and we can do all the operations of stack using ArrayDeque that's why i thought we can change it , although it won't affect the final result.

siriak (commented):
The algorithm calls for stack interface and using Stack class reflects that better. Using IStack interface would be even better (Stack and Deque would both implement it), but it doesn't exist as of now.

PR #3240, opened 2022-08-30
Binary Search in 2D Array

Shashwat79802:
Added the program to search for the target element using **Binary Search** in the provided **2D array**. The program is modified but uses the same approach as it is built on top of binary search.


siriak (changes requested):
The code looks good but lacks some JUnit tests. Could you add those?

Shashwat79802:
Added JUnit test for the code.

siriak (approved):
Looks good, thanks!

PR #3246, opened 2022-09-02
Adding the program of Magic number and it's JUnit test

Shashwat79802:
Added a new directory named **bitmanipulation** inside the main and test directory and added the program to find the **Nth Magic Number** and its **JUnit test**. If the new directory get's merged, I have more programs on bit manipulation to put into the directory.

### DIFF
--- a/src/main/java/com/thealgorithms/bitmanipulation/MagicNumber.java
+++ b/src/main/java/com/thealgorithms/bitmanipulation/MagicNumber.java
@@ -1,0 +1,30 @@
+package com.thealgorithms.bitmanipulation;
+
+/*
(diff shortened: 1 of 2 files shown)


siriak (changes requested):
The code looks good, but could you include a reference to where more info on magic numbers can be found? A Wikipedia or OEIS page would be ideal

Shashwat79802:
I couldn't find any reference to Magic Number, there are direct code samples only available. But I guess, the provided explanation should be helpful.

siriak:
We only include solutions to well-known problems and this one seems to be just a programming exercise. Therefore, it's unlikely to be included despite the code and tests looking good. Thanks for your time!

Issue #3245, opened 2022-09-01
Feature for "LinkedList" to find the middle node.


Tarunbot:
as there is no predefined implementation for finding of middle node in linked list so i thought if this feature will be added it would be a plus.so allow me to do so .Thank you.

parth-verma7:
@Tarunbot I would like to work upon this issue. Can u please assign it to me?

dev-singh2:
@siriak Is the issue still unsolved? if so, I would be glad to work on it.

chandraPrakash23:
I can solve this task using three methods with time complexity of O(N), so you can assign this task to me. Thank You

shahabrar7746:
hey, is there anyone working on this issue. if not let me work on this. as this looks a perfect start for my Open Source contribution.

shahabrar7746:
heyy, @Tarunbot i have made a PR for this issue. if you dont mind can you please review it and merge it to master
https://github.com/TheAlgorithms/Java/pull/3849

shahabrar7746:
@siriak can you assign this issue to me.

shahabrar7746:
@siriak check my PR https://github.com/TheAlgorithms/Java/pull/3862

aayan14:
Hi @siriak 
If this issue is still unsolved can you assign it to me?

PR #3249, opened 2022-09-05
#3245 Added functionality to find the middle node of singly linked list

Tarunbot:
#3245

so here i actually implemented the functionality using slow and fast pointer to find the middle node  so how it will work:-
So basically this function will return the middle node of the singly linked list.
if Linked list size is odd----->it will return the middle node
else if Linked List size is even ---->it will return the first node from the two middle nodes.


siriak (changes requested):
Please add JUnit tests to check that it works

Abhishekverma2002 (approved):
Time complexity = O(n)
Space Complexity= O(n)

PR #3252, opened 2022-09-05
Binary Search in Infinite Array

Shashwat79802:


siriak (changes requested):
Is it https://en.wikipedia.org/wiki/Exponential_search? If so, please change naming accordingly, add JUnit tests, remove `main` and reuse binary search as a subroutine instead of duplicating it here, thanks!

PR #3262, opened 2022-09-10
Added file to calculate set bits in a number

PratyushSingh07:
I have used Brian Kernighans algorithm to count set bits of a number . I have add this file in others folder and the test file in the corresponding folder in test folder.


siriak (approved):
Looks good, thanks!

PR #3300, opened 2022-09-28
sort a LinkedList in Java

megha-vishwakarma:
I have used the Collections.sort() method for sorting elements stored in the LinkedList class.


siriak:
We don't include examples of usage, thanks!

PR #3373, opened 2022-10-03
Added Binary tree traversal and practice problems

CoderLovely08:
Added Tree traversal algorithms
- Iterative
- 1. Depth first search
- 2. Breadth First Search


CoderLovely08:
Will submit again after rectifying changes

PR #3424, opened 2022-10-06
Request to add Trapping Rain Water in misc.

AnuragKumar007:
It's my 1st PR. I want to add Classic question "**Trapping Rainwater**" in misc folder. I've tried my best to keep the code neat & clean. Also used comments to make the code easy to understand for all.
Request to add a **Classic Question**.
### **Describe your change:**

- [x] Add an algorithm?
- [ ] Documentation change?

### **Checklist:**

- [x] All new algorithms have a URL in its comments that points to Wikipedia or other similar explanation.

### DIFF
--- a/src/main/java/com/thealgorithms/misc/TrappingRainwater.java
+++ b/src/main/java/com/thealgorithms/misc/TrappingRainwater.java
@@ -14,7 +14,7 @@ import java.util.*;
  */
 public class TrappingRainwater {
 
-    public static long trappedRainWater(int height[]){
+    public static long trappedRainwater(int height[]){
         int n = height.length;
         int waterLevel =0;
         int trappedWater =0;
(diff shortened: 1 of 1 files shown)


gkgaurav31:
In line 39, 

`trappedWater += waterLevel - height[i]; //width =1`

It will have negative value (which is not possible for water trapped) if height[i] is more than both leftMax[i] and rightMax[i].

We can fix that in the following way:

```
//If the currentHeight is more than the max on left and right both, take 0.
trappedWater += Math.max(0, Math.min(leftMax[i],rightMax[i])-height[i]);
```

AnuragKumar007:
> 

Thanks! but can you tell me the test case too? 
I think the code is correct. Let's take an array - {3,0,5,2,0,2} 
it should give output - 5 and it's giving

gkgaurav31:
Just noticed, the way you've generated the prefix/suffix array takes care of this edge case. Should be okay.

siriak (changes requested):
The code looks good, could you please add some JUnit tests? (see https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms)

PR #3555, opened 2022-10-15
Added Sub Array program in java

sdeep1998:
#Added Sub Array program in java

- Added Sub array program to find the equal sum of elements without changing the order of digits.


sdeep1998 (commented):
Taking the odd length of subarrays and only doing the sum of those.

siriak:
And a reference to problem statement

PR #3572, opened 2022-10-17
Added Radix Sort in Java Programming

pareshapraj:
### **Radix Sort in Java Programming**

Radix sort is [a sorting algorithm](https://www.programiz.com/dsa/sorting-algorithm) that sorts the elements by first grouping the individual digits of the same place value. Then, sort the elements according to their increasing/decreasing order.

Suppose, we have an array of 8 elements. First, we will sort elements based on the value of the unit place. Then, we will sort elements based on the value of the tenth place. This process goes on until the last significant place.

### DIFF
--- a/src/test/java/com/thealgorithms/sorts/RadixSort.java
+++ b/src/test/java/com/thealgorithms/sorts/RadixSort.java
@@ -1,0 +1,66 @@
+// Radix Sort in Java Programming
+
+import java.util.Arrays;
+
+class RadixSort {
+
+  // Using counting sort to sort the elements in the basis of significant places
+  void countingSort(int array[], int size, int place) {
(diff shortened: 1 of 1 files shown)


debasishbsws:
@pareshapraj  `RadixSort.java` is already present inside `src/main/java/com/thealgorithms/sorts/RadixSort.java`. you have put it inside the test package where it does not belong.

debasishbsws:
But you can write test cases for Radix Sort inside `src/test/java/com/thealgorithms/sorts/` make the file name `RadixSortTest.java` and put tests inside it. you can refer to other files present in that package for reference.

siriak (changes requested):
Please replace this with JUnit tests

PR #3645, opened 2022-10-24
Optimized Pangram.java code to O(1)

TaranjeetSinghKalsi:
optimized code from O(n) to O(1) and added comments


debasishbsws:
Here the code is not Optimized, because `String.indexOf(char c)` takes `O(n)` in the best case. [stackoverflow](https://stackoverflow.com/questions/3562919/what-is-the-cost-complexity-of-a-string-indexof-function-call).
  
You can put this as another way to solve the problem. Make another Method and put this approach there.

TaranjeetSinghKalsi:
> Here the code is not Optimized, because `String.indexOf(char c)` takes `O(n)` in the best case. [stackoverflow](https://stackoverflow.com/questions/3562919/what-is-the-cost-complexity-of-a-string-indexof-function-call).
> 
> You can put this as another way to solve the problem. Make another Method and put this approach there.

ok

PR #3717, opened 2022-10-28
Added functions in Perimeter.java

TaranjeetSinghKalsi:
Added functions perimeterTriangle() and perimeterTrapezoid() in Perimeter.java and corrected functions names in PascalCase and removed main()
Added JUnit Tests for new functions and renamed perimterTest.java to PerimeterTest.java


src/main/java/com/thealgorithms/maths/Perimeter.java
```diff
@@ -1,42 +1,62 @@
...
+     * Calculate the Perimeter of polygon
+     * 
      * @parameter number of sides.
+     * @parameter length of side.
      * @return Perimeter of given polygon
      */
-    public static float perimeter_polygon(int n, float side) {
-        float perimeter = n * side;
-        return perimeter;
+    public static float perimeterPolygon(int n, float side) {
```
debasishbsws, line 13: Polygon doesn't have to have equal length sides. It may be different. Make two functions of same name one for equal length sides Polygon, another for variable length (take variable length argument as parameters)

src/test/java/com/thealgorithms/maths/perimeterTest.java
```diff
@@ -1,40 +0,0 @@
-package com.thealgorithms.maths;
```
debasishbsws, line 1: Remove this file

debasishbsws (changes requested):

src/main/java/com/thealgorithms/maths/Perimeter.java
TaranjeetSinghKalsi: > 

ok, then should I also remove perimeterTriangle() and perimeterTrapezoid() and hence replace them with variable-length polygon perimeter calculator function ?

src/test/java/com/thealgorithms/maths/perimeterTest.java
TaranjeetSinghKalsi: > Remove this file

Already removed from my side

src/main/java/com/thealgorithms/maths/Perimeter.java
debasishbsws: Yes, I think, you are right. rather than a triangle, we could have an Equilateral triangle or a polygon that has all the sides of equal length. Just put some comments about what the function is representing

src/test/java/com/thealgorithms/maths/perimeterTest.java
debasishbsws: you should run 
```sh
git rm perimeterTest.java
```

TaranjeetSinghKalsi: ok thanks

> you should run
> 
> ```shell
> git rm perimeterTest.java
> ```

PR #3763, opened 2022-11-10
Fixing changes

PraveenNanda124:


siriak:
In this repo we use braces for ifs with single-line bodies

PR #3865, opened 2023-01-30
Create JobSequncingalgo.java

Vedeesh6:


debasishbsws (changes requested):
Change the file name to JobSequencing.java

debasishbsws:
It should fix the build fail

Vedeesh6:
> It should fix the build fail
Sorry, didn't pay heed to build errors


> Change the file name to JobSequencing.java
Yeah, I've done it now.

Vedeesh6:
> It should fix the build fail


Yeah, I've done it now. Thanks.

siriak (changes requested):
What scheduling algorithm is that? Some algorithms are already implemented 
image

Vedeesh6:
> What scheduling algorithm is that? Some algorithms are already implemented 

Yeah, I did see them before applying for this pull request, and even though the second one looks very similar by name, the fact is that it isn't. Both of these algorithms in the image are for the processes to organize them according to their burst time.
But the algorithm which I've proposed is one of the most widely known problems in dynamic programming.
This problem states that you have a business running and you've been provided with some tasks that have their own deadlines and profits such that they all can't be fulfilled in the given time, so obviously you need to choose and choose the best order to gain maximum profit.
Now one approach is that take jobs with max profit and try to maximize their number but that's what a greedy approach is, and we all know that it can't offer the best solution every time. So what to do?
That's what this algorithm is all about.
For more info you can visit:
https://www.scaler.com/topics/job-sequencing-with-deadlines/ 
or just google "Job Sequencing Problem".

siriak (changes requested):
Got it, it's different. Could you remove logging to stdout and add JUnit tests for the algorithm then?

Vedeesh6:
> Got it, it's different. Could you remove logging to stdout and add JUnit tests for the algorithm then?

I tried to do that but it was way too hard because non of the primitive datatypes are used in the base, a new class argument "Job" is introduced to produce the desired result and the output is such that it feels interactive and is easy to understand even for a beginner. You should try to run the code once to gain a better insight.

Sample input dataset =>
Enter number of jobs -> 8

For Job 1 ->

Enter the profit and deadline (space separated) ->60 2

For Job 2 ->

Enter the profit and deadline (space separated) ->90 6

For Job 3 ->

Enter the profit and deadline (space separated) ->30 3

For Job 4 ->

Enter the profit and deadline (space separated) ->90 4

For Job 5 ->

Enter the profit and deadline (space separated) ->30 1

For Job 6 ->

Enter the profit and deadline (space separated) ->60 4

For Job 7 ->

Enter the profit and deadline (space separated) ->70 1

For Job 8 ->

Enter the profit and deadline (space separated) ->100 6

In the output spot that the job on day 3rd is exempted to include both the jobs of the 4th day because they give more profit, this shows a better approach for job selection.

siriak (commented):
We implement algorithms in reusable format so that they could be used from client applications if needed. This also makes algorithm contracts clear. Having tests is required for all new algorithms

Issue #3922, opened 2023-03-14
[BUG] DIRECTORY.md is not properly formatted.


cclauss:
### Description

[`DIRECTORY.md`](../blob/master/DIRECTORY.md) looks really ugly especially at the end.

Perhaps the Python code run in our GitHub Action [directory_workflow.yml](../blob/master/.github/workflows/update_directory.yml#L25) should be replaced by Java code so that contributors to this repo are better able to maintain it.

### Steps to reproduce

1. Click on [`DIRECTORY.md`](../blob/master/DIRECTORY.md) and look.

### Excepted behavior

Users want a readable directory.

siriak:
There are two options:
1. Fix the existing script to format properly
2. Rewrite it in Java so that it formats properly and can be maintained in the future

PR #4124, opened 2023-03-26
fix DIRECTORY.md is not properly formatted. #3922

JiaZombie:
- [ ] ~All filenames are in PascalCase.~
- [ ] ~All functions and variable names follow Java naming conventions.~
- [ ] ~All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.~

### DIFF
--- a/.github/workflows/update_directory.yml
+++ b/.github/workflows/update_directory.yml
@@ -1,5 +1,7 @@
 # This GitHub Action updates the DIRECTORY.md file (if needed) when doing a git push or pull_request
 name: Update Directory
+permissions:
+  contents: write
 on:
   push:
     paths:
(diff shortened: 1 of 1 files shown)


siriak (commented):
Could you make it run on every push and pull request? I don't think running it an extra time should be a problem.
image

siriak:
I meant to change it to `on: ["push", "pull_request"]`, but it's not that important. There's a bigger problem with permissions , which means that the directory file is not updated as the workflow runs:
image
Do you know what the problem might be and how can I grant those permissions?

siriak (approved):
Looks good, thanks!

siriak (approved):
Thanks, will see how it works!

PR #4130, opened 2023-03-28
Optimised SieveOfEratosthenes algorithm

sanjug-sonowal:

/* Respected Maintainer of this Repository 
I hope this message finds you well. I am writing to propose an optimized implementation of the Sieve of Eratosthenes algorithm in Java.

I have made the following improvements to the existing implementation:

Removed unnecessary comments and improved code readability.
Simplified the implementation by removing redundant code.
Reduced the loop iterations by checking only odd numbers after 2.

I believe these changes will result in a more efficient and concise implementation of the algorithm.

Thank you for considering my proposal. I would be happy to address any questions or concerns you may have.
*/

import java.util.ArrayList;
import java.util.Arrays;

public class SieveOfEratosthenes {

    /**
     * @param n The number till which we have to check for prime Prints all the
     * prime numbers till n. Should be more than 1.
     * @return ArrayList of all prime numbers between 0 to n
     */
    public static ArrayList<Integer> findPrimesTill(int n) {
        // Create boolean array where index is number and value is flag - is that number a prime or not.
        boolean[] numbers = new boolean[n + 1];

        // Start with assumption that all odd numbers are primes.
        Arrays.fill(numbers, true);
        numbers[0] = numbers[1] = false;
        numbers[2] = true;

        double cap = Math.sqrt(n);
        // Main algorithm: mark all numbers which are multiples of some other primes as not prime
        for (int i = 3; i <= cap; i += 2) {
            if (numbers[i]) {
                for (int j = i * i; j <= n; j += 2 * i) {
                    numbers[j] = false;
                }
            }
        }

        // Write all primes to result list
        ArrayList<Integer> primes = new ArrayList<>();
        primes.add(2);
        for (int i = 3; i <= n; i += 2) {
            if (numbers[i]) {
                primes.add(i);
            }
        }

        return primes;
    }

        int n = 100;
        System.out.println("Searching for all primes from zero to " + n);
        ArrayList<Integer> primes = findPrimesTill(n);
        System.out.println("Found: " + primes);
    }
}

### DIFF
--- a/src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
+++ b/src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
@@ -1,82 +1,48 @@
-package com.thealgorithms.others;
-
+import java.util.ArrayList;
 import java.util.Arrays;
 
-/**
- * Sieve of Eratosthenes is an ancient algorithm for finding all prime numbers
- * up to any given limit. It does so by iteratively marking as composite (i.e.,
- * not prime) the multiples of each prime, starting with the first prime number,
- * 2. The multiples of a given prime are generated as a sequence of numbers
- * starting from that prime, with constant difference between them that is equal
- * to that prime. This is the sieve's key distinction from using trial division
- * to sequentially test each candidate number for divisibility by each prime.
- * Once all the multiples of each discovered prime have been marked as
- * composites, the remaining unmarked numbers are primes.
- * <p>
- * Poetry about Sieve of Eratosthenes:
- * <p>
- * <i>Sift the Two's and Sift the Three's:</i></p>
- * <p>
- * <i>The Sieve of Eratosthenes.</i></p>
- * <p>
- * <i>When the multiples sublime,</i></p>
- * <p>
- * <i>The numbers that remain are Prime.</i></p>
- *
- * @see <a href="https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes">Wiki</a>
- */
 public class SieveOfEratosthenes {
 
     /**
      * @param n The number till which we have to check for prime Prints all the
      * prime numbers till n. Should be more than 1.
-     * @return array of all prime numbers between 0 to n
+     * @return ArrayList of all prime numbers between 0 to n
      */
-    public static int[] findPrimesTill(int n) {
-        // Create array where index is number and value is flag - is that number a prime or not.
-        // size of array is n + 1 cause in Java array indexes starts with 0
-        Type[] numbers = new Type[n + 1];
+    public static ArrayList<Integer> findPrimesTill(int n) {
+        // Create boolean array where index is number and value is flag - is that number a prime or not.
+        boolean[] numbers = new boolean[n + 1];
 
-        // Start with assumption that all numbers except 0 and 1 are primes.
-        Arrays.fill(numbers, Type.PRIME);
-        numbers[0] = numbers[1] = Type.NOT_PRIME;
+        // Start with assumption that all odd numbers are primes.
+        Arrays.fill(numbers, true);
+        numbers[0] = numbers[1] = false;
+        numbers[2] = true;
 
         double cap = Math.sqrt(n);
-        // Main algorithm: mark all numbers which are multiples of some other values as not prime
-        for (int i = 2; i <= cap; i++) {
-            if (numbers[i] == Type.PRIME) {
-                for (int j = 2; i * j <= n; j++) {
-                    numbers[i * j] = Type.NOT_PRIME;
+        // Main algorithm: mark all numbers which are multiples of some other primes as not prime
+        for (int i = 3; i <= cap; i += 2) {
+            if (numbers[i]) {
+                for (int j = i * i; j <= n; j += 2 * i) {
+                    numbers[j] = false;
                 }
             }
         }
 
-        //Write all primes to result array
-        int primesCount = (int) Arrays
-                .stream(numbers)
-                .filter(element -> element == Type.PRIME)
-                .count();
-        int[] primes = new int[primesCount];
-
-        int primeIndex = 0;
-        for (int i = 0; i < n + 1; i++) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
```diff
@@ -1,82 +1,48 @@
...
- * <p>
- * <i>Sift the Two's and Sift the Three's:</i></p>
- * <p>
- * <i>The Sieve of Eratosthenes.</i></p>
- * <p>
- * <i>When the multiples sublime,</i></p>
- * <p>
- * <i>The numbers that remain are Prime.</i></p>
- *
- * @see <a href="https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes">Wiki</a>
```
siriak, line 26: Please leave a short explanation and a reference to Wikipedia

src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
```diff
@@ -1,82 +1,48 @@
...
- */
 public class SieveOfEratosthenes {
 
     /**
      * @param n The number till which we have to check for prime Prints all the
      * prime numbers till n. Should be more than 1.
-     * @return array of all prime numbers between 0 to n
+     * @return ArrayList of all prime numbers between 0 to n
```
siriak, line 9: Why do you return ArrayList instead of array?

src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
```diff
@@ -1,82 +1,48 @@
...
      */
-    public static int[] findPrimesTill(int n) {
-        // Create array where index is number and value is flag - is that number a prime or not.
-        // size of array is n + 1 cause in Java array indexes starts with 0
-        Type[] numbers = new Type[n + 1];
+    public static ArrayList<Integer> findPrimesTill(int n) {
```
siriak, line 11: What if n is negative?

src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
```diff
@@ -1,82 +1,48 @@
...
-    }
-
     public static void main(String[] args) {
         int n = 100;
         System.out.println("Searching for all primes from zero to " + n);
-        int[] primes = findPrimesTill(n);
-        System.out.println("Found: " + Arrays.toString(primes));
+        ArrayList<Integer> primes = findPrimesTill(n);
+        System.out.println("Found: " + primes);
     }
```
siriak, line 47: Please remove `main` and add JUnit tests if they are missing

src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
```diff
@@ -1,82 +1,48 @@
...
-     * @return array of all prime numbers between 0 to n
+     * @return ArrayList of all prime numbers between 0 to n
      */
-    public static int[] findPrimesTill(int n) {
-        // Create array where index is number and value is flag - is that number a prime or not.
-        // size of array is n + 1 cause in Java array indexes starts with 0
-        Type[] numbers = new Type[n + 1];
+    public static ArrayList<Integer> findPrimesTill(int n) {
+        // Create boolean array where index is number and value is flag - is that number a prime or not.
+        boolean[] numbers = new boolean[n + 1];
```
siriak, line 13: Can `numbers` be renames to show what it holds? Like `isPrime`?

siriak (changes requested):

src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
sanjug-sonowal: Arrays in Java have a fixed size, so if we were to return an array from the method, we would need to specify its size before calling the method. This can be inconvenient in some cases. ArrayLists, on the other hand, can grow dynamically as needed, so they are more flexible.

src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
sanjug-sonowal: If n is negative, the current implementation would throw an ArrayIndexOutOfBoundsException when creating the boolean array. To handle this case, we can add a check at the beginning of the method to throw an IllegalArgumentException with a message explaining that the limit must be at least 2.

src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
siriak: Why do we need to specify the array size before calling the method? We can figure out how many elements in array we need inside the method. The biggest difference between them is that ArrayList can be mutated easier than array and we don't want that because it's only used to return data. So using array is better in this case. You can use ArrayList inside the method when accummulating primes, but return an array.

src/main/java/com/thealgorithms/others/SieveOfEratosthenes.java
siriak: Yes please add validation

Aman254:
hey @sanjug-sonowal @siriak I have More better Optimized Approach Here's It 


public class Seive {
        int n = 40;
        boolean[] primes = new boolean[n+1];
        sieve(n, primes);
    }

    // false in array means number is prime
    static void sieve(int n, boolean[] primes) {
        for (int i = 2; i*i <= n; i++) {
            if (!primes[i]) {
                for (int j = i*2; j <= n; j+=i) {
                    primes[j] = true;
                }
            }
        }

        for (int i = 2; i <= n; i++) {
            if (!primes[i]) {
                System.out.print(i + " ");
            }
        }
    }
}

Aman254:
@siriak Just 26 -30 Lines and No Repetition and Better complexity

siriak:
@Aman254 please open a PR with the changes and JUnit tests for them and I'll review

Aman254:
@siriak Sure !

PR #4136, opened 2023-03-31
If array is already empty.

R-Alothaim:
Added if condition to check whether the array is already empty or no, if yes return null.


siriak (commented):
The proposed behavior is incorrect because an error must be thrown if there's no element at the specified index

PR #4142, opened 2023-04-04
NullPointer Exception is handled.

Akshith121:


siriak (changes requested):
The code looks good, could you please add some JUnit tests to prove the fix is effective? (see https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms)

Akshith121:
> The code looks good, could you please add some JUnit tests to prove the fix is effective? (see https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms)

I have made some minor changes to reverseList() method and added some JUnit tests to prove that code works fine in any testcase

src/main/java/com/thealgorithms/datastructures/lists/SinglyLinkedList.java
```diff
@@ -122,20 +122,25 @@ public void swapNodes(int valueFirst, int valueSecond) {
...
+    public Node reverseList(Node node) {
+        //When node is null, then "next" will cause an error as we are accesing node.next, which will give a NullPointer Exception
+        if(node==null){
+            return node;
         }
-        Node prev = null, curr = node, next;
+        Node prev = null;
+        Node curr = node;
+        Node next=curr.next;
         while (curr != null) {
```
siriak, line 133: Could you add a check for curr.next  != null here instead of the nested if?

src/main/java/com/thealgorithms/datastructures/lists/SinglyLinkedList.java
```diff
@@ -122,20 +122,25 @@ public void swapNodes(int valueFirst, int valueSecond) {
...
      */
-    Node reverseList(Node node) {
-        Node prevNode = head;
-        while (prevNode.next != node) {
-            prevNode = prevNode.next;
+    public Node reverseList(Node node) {
+        //When node is null, then "next" will cause an error as we are accesing node.next, which will give a NullPointer Exception
+        if(node==null){
+            return node;
         }
```
siriak, line 129: This check is redundant, this case is handled by the loop below already

siriak (changes requested):

src/main/java/com/thealgorithms/datastructures/lists/SinglyLinkedList.java
Akshith121: without checking whether curr is null , if we check for curr.next it might cause NullPointer Exception , so I will check for    
 curr != null and put next = curr.next inside the while loop

siriak (approved):
Looks good, thanks!

PR #4175, opened 2023-04-30
Added another Optimized solution for Anagrams check.

manojCode94:

/*Respected Maintainer of this Repository, I hope this message finds you well. I am writing to propose another optimized solution for checking if two strings are anagrams or not in Java.
*/

**Approach**
```
boolean isAnagram(String s, String t) {
    if (s.length() != t.length()) {
        return false;
    }
    int[] freq = new int[26];
    for (int i = 0; i < s.length(); i++) {
      int pos1 = s.charAt(i) - 'a';
      int pos2 = s.charAt(i) - 'a';
      freq[pos1]++;  freq[pos2]--;
    }
    for (int i = 0; i < 26; i++) {
        if (freq[i] != 0) {
            return false;
        }
    }
    return true;
}
```

- If the lengths of the two strings are different, they cannot be anagrams. Return false.
- Initialize an array of size 26, where each element corresponds to the frequency of a character.
- Iterate through both strings, incrementing the frequency of each character in the first string and decrementing the frequency of each character in the second string.
- Iterate through the frequency array and check if all the elements are zero. If so, the two strings are anagrams of each other. Return true.
- If any element in the frequency array is non-zero, the two strings are not anagrams of each other. Return false.

**Complexity**

- The time complexity of this approach is O(n), where n is the length of the input strings.
- The space complexity is O(1), as we are using a fixed-size array. This approach is faster than using a hashmap, as it avoids the overhead of creating and iterating through a hashmap.


siriak (changes requested):
The code looks good, could you please add some JUnit tests? (see https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms)

manojCode94:
Added JUnit Test as per suggested for Approach5 in AnagramsTest.java Build successfully.

siriak (approved):
Looks good, thanks!

manojCode94:
Thanks @siriak

PR #4200, opened 2023-05-27
Octal to binary Converter 

BamaCharanChhandogi:

**Octal To Binary Code**


src/main/java/com/thealgorithms/conversions/OctalToBinary.java
```diff
@@ -0,0 +1,39 @@
+package com.thealgorithms.conversions;
+
+import java.util.Scanner;
+public class OctalToBinary {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.print("Enter an octal number: ");
+        int octalNumber = scanner.nextInt();
```
itamar124812, line 8: This number is by default decimal right? (because java int works with decimal).
You need to get the number as a string.

src/main/java/com/thealgorithms/conversions/OctalToBinary.java
```diff
@@ -0,0 +1,39 @@
...
+
+        // Convert octal to binary
+        long binaryNumber = convertOctalToBinary(octalNumber);
+
+        System.out.println("Binary equivalent: " + binaryNumber);
+    }
+
+    public static long convertOctalToBinary(int octalNumber) {
+        int decimalNumber = 0;
```
itamar124812, line 17: There is a better way to convert from octal to binary (you can convert directly from octal to binary without converting to decimal in the middle).
Every digit in octal can immediately converted to 3 digits in binary. (because 2^3 = 8)

src/main/java/com/thealgorithms/conversions/OctalToBinary.java
BamaCharanChhandogi: Check new PR

PR #4216, opened 2023-06-19
tests: add tests for `EulersFunction`

vil02:
This PR:
- adds tests for [`EulersFunction`](https://github.com/TheAlgorithms/Java/blob/e29a22770ba4b89be32df07448868379682d50f0/src/main/java/com/thealgorithms/others/EulersFunction.java#L9),
- updates `EulersFunction.getEuler` to throw an `IllegalArgumentException` for non-positive inputs, 
- makes `EulersFunction` a proper _utility_ class,
- updates the documentation.

### DIFF
--- a/src/main/java/com/thealgorithms/others/EulersFunction.java
+++ b/src/main/java/com/thealgorithms/others/EulersFunction.java
@@ -1,10 +1,7 @@
 package com.thealgorithms.others;
 
 /**
- * You can read more about Euler's totient function
- *
- * <p>
- * See https://en.wikipedia.org/wiki/Euler%27s_totient_function
+ * @brief utility class for <a href="https://en.wikipedia.org/wiki/Euler%27s_totient_function">Euler's totient function</a>
  */
 final public class EulersFunction {
     private EulersFunction() {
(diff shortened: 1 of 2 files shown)


src/main/java/com/thealgorithms/others/EulersFunction.java
```diff
@@ -1,17 +1,27 @@
...
- * You can read more about Euler's totient function
- *
- * <p>
- * See https://en.wikipedia.org/wiki/Euler%27s_totient_function
+ * @brief utility class for <a href="https://en.wikipedia.org/wiki/Euler%27s_totient_function">Euler's totient function</a>
  */
-public class EulersFunction {
+final public class EulersFunction {
+    private EulersFunction() {
+    }
```
debasishbsws, line 8: Why we are restricted from making an object of this class?

vil02: The class `EulersFunction` has no non-`static` members. Therefore, in my opinion, having objects of that class does not make too much sense. That is why I decided to make the default constructor `private`.

Also thanks to this, `jacoco` shows _100%_ code coverage (because it is not complaining about not creating objects of the class `EulersFunction`).

siriak (approved):
Looks good, thanks!

PR #4226, opened 2023-07-01
Java Soptless formatter implemented

### LINKED ISSUE #4224: [FEATURE REQUEST] Java Code Formatter

The current Clang format linter definitely does its job to identify linting errors in the code.
But I think it will be better if we could have a formatter that we can integrate with CI jobs to auto-formate code if possible.
Therefore, It will be better to have a Java code formatter.

### Issue details

Currently, contributors have to format the code manually. though it can a auto fixable

debasishbsws:
Fix #4224


debasishbsws:
It seems like it is caching previous results that it should not. Let me see what I can do.

yanglbme:
This code change is too big, I think it is better to continue using spaces as the indentation method.

siriak:
Yes, please change back to spaces. They are displayed consistently on GitHub

PR #4230, opened 2023-07-06
DFS and BFS search algorithms  - Resolve #4229

### LINKED ISSUE #4229: [FEATURE REQUEST] DFS and BFS  search algorithms are lack of effective tests and Node class should be decoupled from DFS

- Write DFS and BFS  search algorithms effective tests.
- Write a new Node class in structures  - its should be decoupled from the DFS class.
- Some code refactoring.
-  Support of multiple data types

### Issue details

BreadthFirstSearch.java
- add algorithm tests
- refactor
DepthFirstSearch.java

- write a new test file 
- refactor - mention that this is a recursive solution (in the future add an iterative solution as well)

### Additional Information

_No response_

snirt:

Resolve #4229


pom.xml
```diff
@@ -37,8 +37,10 @@
         <dependency>
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
-            <version>${assertj.version}</version>
+            <version>3.21.0</version>
```
debasishbsws, line 40: assertj version we are using is `3.23.1` why we need to change that?

src/main/java/com/thealgorithms/datastructures/Node.java
debasishbsws, line 1: Can't we use the already written [Graphs.java](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/graphs/Graphs.java) or [GenericTree](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/trees/GenericTree.java) data structure. and search among that?

snirt: a tree could be represented both ways:

- graph that defined by edges and vertices - can have more methods and features
- root node that points to another nodes - a simpler, straight-forward way to describe a tree

in my opinion, this way is better for BFS / DFS because it explicitly defines the parent-child relationship between nodes.

pom.xml
debasishbsws: `assertThat` was already deprecated in Junit assertion. And they have removed it in higher version. you can use other methods to do the same. [Junit Assertions Docs](https://junit.org/junit5/docs/5.0.1/api/org/junit/jupiter/api/Assertions.html)

but if you require to use the assertThat it is available on Assertj. You need to import
`import static org.assertj.core.api.Assertions.assertThat;` to use.

Do not downgrade the `assertj.version`

debasishbsws (changes requested):
Also Check the linting error

pom.xml
snirt: I had some problems with this package right after opened the project, i tried to change some configurations to make it to be compiled, it resolved when I changed the version to 3.21.0.
I'm still not sure how to resolve this issue in a better way, trying to resolve it right now. 
Screenshot from 2023-07-09 22-53-25
anyway, Its may not should be a part of this PR

snirt:
Regarding the linter format, I suggest to configure it to allow the left-side format in the attached image:
Screenshot from 2023-07-10 13-32-18
This format helps with readability

src/main/java/com/thealgorithms/datastructures/Node.java
debasishbsws: DFS and BSF are graph traversal algorithms, though we learn and understand them via Tree as it is easier that way but don't you think
A graph traversal algorithm should be Grneric to all graphs even trees. Or we can have both implementations there.

siriak: I agree that searches should work on graphs and not trees as there is absolutely nothing specific to trees in them

siriak:
> Regarding the linter format, I suggest to configure it to allow the left-side format in the attached image: Screenshot from 2023-07-10 13-32-18 This format helps with readability

I agree with you, could you do it here or in a separate PR if it involves a lot of changes?

src/main/java/com/thealgorithms/datastructures/Node.java
snirt: maybe we should separate them to two sections - `Graphs` and `Trees`, Although tree is a kind of graph, since both are common valid representations.

debasishbsws: I think just Graph is fine. as it will be more generic. As Trees are Graph. So the same algo can work on both

siriak (approved):
Looks good, thanks!

siriak:
@debasishbsws could you review again?

debasishbsws (approved):
Looks good

Issue #4240, opened 2023-07-14
[FEATURE REQUEST]Add "Burst Balloons" DP Problem


Shiva953:
### What would you like to Propose?

The "burst balloons" problem, which involves determining the optimal order of bursting balloons to maximize the total coins obtained by considering adjacent balloons and applying Dynamic Programming.

### Issue details

You are given n balloons, indexed from 0 to `n - 1`. Each balloon is painted with a number on it represented by an array nums. You are asked to burst all the balloons.

If you burst the `i`th balloon, you will get `nums[i - 1] * nums[i] * nums[i + 1]` coins. If `i - 1` or `i + 1` goes out of bounds of the array, then treat it as if there is a balloon with a 1 painted on it.

Return the maximum coins you can collect by bursting the balloons wisely.

### Additional Information

An example for the same would be : 
Input - nums = [3,1,5,8]
Output - 167
Explanation:
nums = [3,1,5,8] --> [3,5,8] --> [3,8] --> [8] --> []
coins =  3*1*5    +   3*5*8   +  1*3*8  + 1*8*1 = 167

PR #4241, opened 2023-07-14
Added Balloon Burst DP Problem

Shiva953:
For #4240 


siriak:
Could you add a link to Wikipedia or a different source (other than websites with coding problems)? We only add well-known algorithms to avoid infinite growth of the codebase.

siriak (commented):
Thanks for contributing this problem, but we don't add problems from the type of website you referenced. As I see, it's only a programming challenge in applying dynamic programming. As such, there are already more than enough examples of this method in the codebase. Thanks anyway!

PR #4209, opened 2023-05-31
refactor: generalise `NthUglyNumber`

vil02:
This PR:
- refactors the [`NthUglyNumber`](https://github.com/TheAlgorithms/Java/blob/4bbc4bd69f2a68963ba733df744c6ea356aba109/src/main/java/com/thealgorithms/maths/NthUglyNumber.java#L18),
- adds tests.

- [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [x] This pull request is all my own work -- I have not plagiarized it.
- [x] All filenames are in PascalCase.
- [x] All functions and variable names follow Java naming conventions.
- [x] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.

PR #4248, opened 2023-07-22
style: basic linting of `NthUglyNumber`

vil02:
This PR removes some redundant import (i.e. `java.lang.IllegalArgumentException`) and does some other _basic linting_.

- [x] ~~All filenames are in PascalCase.~~
- [x] ~~All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.~~

### DIFF
--- a/src/main/java/com/thealgorithms/maths/NthUglyNumber.java
+++ b/src/main/java/com/thealgorithms/maths/NthUglyNumber.java
@@ -21,7 +21,7 @@ public class NthUglyNumber {
 
     /**
      * @brief initialized the object allowing to compute ugly numbers with given base
-     * @param baseNumbers the given base of ugly numbers
+     * @param inBaseNumbers the given base of ugly numbers
      * @exception IllegalArgumentException baseNumber is empty
      */
     NthUglyNumber(final int[] inBaseNumbers) {
--- a/src/main/java/com/thealgorithms/maths/NthUglyNumber.java
+++ b/src/main/java/com/thealgorithms/maths/NthUglyNumber.java
@@ -1,6 +1,5 @@
 package com.thealgorithms.maths;
 
-import java.lang.IllegalArgumentException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


src/main/java/com/thealgorithms/maths/NthUglyNumber.java
```diff
@@ -16,22 +15,22 @@
...
     /**
      * @brief initialized the object allowing to compute ugly numbers with given base
-     * @param baseNumbers the given base of ugly numbers
+     * @param inBaseNumbers the given base of ugly numbers
      * @exception IllegalArgumentException baseNumber is empty
      */
-    NthUglyNumber(int[] baseNumbers) {
-        if (baseNumbers.length == 0) {
+    NthUglyNumber(final int[] inBaseNumbers) {
+        if (inBaseNumbers.length == 0) {
```
siriak, line 28: What's the point of renaming it? I think `baseNumbers` was adequate

vil02: `checkstyle` reports this as [coding/HiddenField error](https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/HiddenFieldCheck.html).

Originally (#4209) I thought that the logic here is _simple enough_ to use the same name. Now I realized that `checkstyle` complains about this. In order to avoid discussions like _I think that the situation here is simple enough to use the same name_, I would suggest to follow the suggestion from `checkstyle` (discussions should be about important stuff like design decisions etc. not about things, which could be automated).

nth_ugly_number

siriak: We don't have `checkstyle` integrated into our CI pipeline, so there's no need to adhere to its style. Having a name such as `inBaseNumbers` doesn't make sense (the `in` part). The prefix doesn't mean anything, so should be removed. I don't think it's worth changing otherwise totally legible code for the sake of pleasing a formatter we don't even officially use.

vil02: Cf. 01108390d9d8704e63b7be754d45a6e536108091.

siriak (approved):
Looks good, thanks!

PR #4265, opened 2023-07-27
Updating NonRepeatingTwoElements.java for -loop optimization

alisyedj2:
I have just replaced for-loop with for-each which saves runtime in accessing elements of array.

- [ X] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [ X] This pull request is all my own work -- I have not plagiarized it.
- [ X] All filenames are in PascalCase.
- [ X] All functions and variable names follow Java naming conventions.
- [ X] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.

### DIFF
--- a/src/main/java/com/thealgorithms/maths/NonRepeatingElement.java
+++ b/src/main/java/com/thealgorithms/maths/NonRepeatingElement.java
@@ -33,8 +33,9 @@ public class NonRepeatingElement {
         }
 
         // Find XOR of the 2 non repeating elements
-        for (i = 0; i < n; i++) {
-            res ^= arr[i];
+        //using for each is good than looping and accessing
+        for (int j : arr) {
+            res ^= j;
         }
 
         // Finding the rightmost set bit


src/main/java/com/thealgorithms/maths/NonRepeatingElement.java
```diff
@@ -33,8 +33,9 @@ public static void main(String[] args) {
         }
 
         // Find XOR of the 2 non repeating elements
-        for (i = 0; i < n; i++) {
-            res ^= arr[i];
+        //using for each is good than looping and accessing
```
vil02, line 36:
```suggestion
```

vil02 (changes requested):
You have identified a file, which definitely needs some clean-up - that is already very good. Before you do any serious changes, please start with adding a file `NonRepeatingTwoElementsTest.java` to [`./src/test/java/com/thealgorithms/maths`](https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms/maths) containing basic tests. You will quickly realise that the current implantation is _untestable_. In order to make it testable, please _abstract_/move the core functionality to a separate method, for example `public static Set<Integer> find(final int[] arr)` (feel free to use other return type).

I am sure that the maintainers will be more than happy to merge your changes as soon as you will bring this file to some acceptable form.

src/main/java/com/thealgorithms/maths/NonRepeatingElement.java
siriak: No need in this comment

siriak (changes requested):
The code looks good, could you please add some JUnit tests? (see https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms). We require covering all changes with tests

PR #4266, opened 2023-07-27
Factorial refactoring, added unit tests

albina-astr:

### DIFF
--- a/src/main/java/com/thealgorithms/maths/Factorial.java
+++ b/src/main/java/com/thealgorithms/maths/Factorial.java
@@ -9,7 +9,7 @@ public class Factorial {
      */
     public static long factorial(int n) {
         if (n < 0) {
-            throw new IllegalArgumentException("Input number cannot be negative!");
+            throw new IllegalArgumentException("Input number cannot be negative");
         }
         long factorial = 1;
         for (int i = 1; i <= n; ++i) {
--- a/src/main/java/com/thealgorithms/maths/Factorial.java
+++ b/src/main/java/com/thealgorithms/maths/Factorial.java
@@ -1,15 +1,6 @@
 package com.thealgorithms.maths;
 
 public class Factorial {
-
-    /* Driver Code */
-    public static void main(String[] args) {
-        assert factorial(0) == 1;
-        assert factorial(1) == 1;
-        assert factorial(5) == 120;
-        assert factorial(10) == 3628800;
-    }
-
     /**
      * Calculate factorial N using iteration
      *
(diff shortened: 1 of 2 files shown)


src/main/java/com/thealgorithms/maths/Factorial.java
```diff
@@ -1,15 +1,6 @@
 package com.thealgorithms.maths;
 
 public class Factorial {
```
vil02, line 3: I think it does not make sense to have any objects of the class `Factorial` - one can make a _proper_ utility  class of it. What do you think about the following change:
```suggestion
public final class Factorial {
    private Factorial() {
    }
```

src/test/java/com/thealgorithms/maths/FactorialTest.java
```diff
@@ -5,9 +5,19 @@
 import org.junit.jupiter.api.Test;
 
 public class FactorialTest {
+    private static final String EXCEPTION_MESSAGE = "Input number cannot be negative!";
```
siriak, line 8:
```suggestion
    private static final String EXCEPTION_MESSAGE = "Input number cannot be negative";
```
I wouldn't be so emotional in the exception message :)

src/main/java/com/thealgorithms/maths/Factorial.java
```diff
@@ -18,11 +9,12 @@ public static void main(String[] args) {
      */
     public static long factorial(int n) {
         if (n < 0) {
-            throw new IllegalArgumentException("number is negative");
+            throw new IllegalArgumentException("Input number cannot be negative!");
```
siriak, line 12:
```suggestion
            throw new IllegalArgumentException("Input number cannot be negative");
```

src/main/java/com/thealgorithms/maths/Factorial.java
albina-astr: Thanks for the suggestion. But I think it is not necessary at all.

src/test/java/com/thealgorithms/maths/FactorialTest.java
albina-astr: Ok, no emotions next time.

src/main/java/com/thealgorithms/maths/Factorial.java
siriak: @vil02  While the usage of final itself is good, mostly we don't use it at the moment. I think we can make a transition everywhere at once to keep code consistent

siriak (approved):

src/main/java/com/thealgorithms/maths/Factorial.java
vil02: @siriak The main part of this suggestion was to explicitly define the default constructor of `Factorial` as private. The `final` is just a nice small addition.

siriak: I see now, I missed this at first. Anyway, feel free to raise a PR with the changes you propose, and we can continue discussion there

vil02: cf. #4267.

Issue #4338, opened 2023-08-29
[FEATURE REQUEST] Adding geometric mean in math


lukasb1b:
### What would you like to Propose?

I did not see a geometric mean function in maths. 
I added a test feature in the attachment I obviuously would change it according to the rules of the repository if assigned to me. 

### Issue details

Missing geometric mean function in maths src/main/java/com/thealgorithms/maths

### Additional Information

public class test {
    public static void main(String[] args) {
        int[] test = {3,4};
        System.out.println(geometric_mean(test));
    }

    public static double geometric_mean(int[] num) {
        int result=0;
        for(int i:num){
            result+=i*i;
        }

        return Math.pow(result,1.0/ num.length);
    }
}

punit324:
Hi @lukasb1b ,
The method you have proposed for geometric mean is seems to be wrong.
Correct formula is,
GM = (x₁ · x₂ · ... · xₙ)1/n
So,
`public static double geometric_mean(int[] num) {
    int result=1;
    for(int i:num){
        result*=i;
    }

    return Math.pow(result,1.0/ num.length);
}`

lukasb1b:
Yeah your right @punit324  i need to multiply. 
Your solution is wrong too you need multiply it by i^2 so its:
public static double geometric_mean(int[] num) {
    int result=1;
    for(int i:num){
        result*=i*i;
    }
    return Math.pow(result,1.0/ num.length);
}

lukasb1b:
Yeah your right @punit324

punit324:
@lukasb1b
Changes are merged into master. You can now close this issue as done.

raisahab9125:
For the code you have to apply the formula in a right manner the appering
code showing the syntax error.
Check out the formula of arithmetic mean.

raisahab9125:
public static void main(String[] args) {
int[] test = {3,4};
System.out.println(geometric_mean(test));
}

public static double geometric_mean(int[] num0) {
    int result=0;
    for(int i*num0){
        result+=i*i;
    }

    return Math.pow(result,1.1/ num0.length);
}
}







The code is ready to move.

lukasb1b:
We already discussed this issue and its fixed and the right algorithm is merged

punit324:
@raisahab9125 I've used formulas from wiki.
Here's the links.
[Arithmetic Mean](https://en.wikipedia.org/wiki/Arithmetic_mean)
[...]

PR #4339, opened 2023-08-30
Added different types of Mean

punit324:
Changes for feature request #4338 

- [x]  This pull request is all my own work -- I have not plagiarized it.
- [x]  All functions and variable names follow Java naming conventions.
- [x]  All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.


src/main/java/com/thealgorithms/maths/Mean.java
```diff
@@ -0,0 +1,54 @@
+package com.thealgorithms.maths;
+
+/**
+ * https://en.wikipedia.org/wiki/Mean
+ * <p>
+ * by: Punit Patel
+ */
+public class Mean {
```
vil02, line 8:
```suggestion
public final class Mean {
    private Mean() {
    }
```
This makes this class a _proper utility class_.

src/main/java/com/thealgorithms/maths/Mean.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Arithmetic / Pythagorean Mean = (x₁ + x₂ + ... +  xₙ) / n
+     */
+    public static double arithmeticMean(double[] numbers) {
+        double sum = 0;
+        if (numbers.length == 0) {
+            return sum;
```
vil02, line 16: Returning `0` might be very misleading in such situation. Please note that some of the tests need to be updated.
```suggestion
            return Double.NAN;
```

src/main/java/com/thealgorithms/maths/Mean.java
```diff
@@ -0,0 +1,54 @@
...
+        return sum / numbers.length;
+    }
+
+    /**
+     * Geometric Mean =  (x₁ · x₂ · ... · xₙ) ^ 1/n
+     */
+    public static double geometricMean(double[] numbers) {
+        double product = 1;
+        if (numbers.length == 0) {
+            return product;
```
vil02, line 31:
```suggestion
            return Double.NAN;
```

src/main/java/com/thealgorithms/maths/Mean.java
```diff
@@ -0,0 +1,54 @@
...
+ * https://en.wikipedia.org/wiki/Mean
+ * <p>
+ * by: Punit Patel
+ */
+public class Mean {
+
+    /**
+     * Arithmetic / Pythagorean Mean = (x₁ + x₂ + ... +  xₙ) / n
+     */
+    public static double arithmeticMean(double[] numbers) {
```
vil02, line 13:
```suggestion
    public static double arithmeticMean(final double[] numbers) {
```

src/main/java/com/thealgorithms/maths/Mean.java
```diff
@@ -0,0 +1,54 @@
...
+            sum += num;
+        }
+
+        return sum / numbers.length;
+    }
+
+    /**
+     * Geometric Mean =  (x₁ · x₂ · ... · xₙ) ^ 1/n
+     */
+    public static double geometricMean(double[] numbers) {
```
vil02, line 28:
```suggestion
    public static double geometricMean(final double[] numbers) {
```

src/main/java/com/thealgorithms/maths/Mean.java
```diff
@@ -0,0 +1,54 @@
...
+            product *= num;
+        }
+
+        return Math.pow(product, 1.0 / numbers.length);
+    }
+
+    /**
+     * Harmonic Mean = n / (1/x₁ + 1/x₂ + ... +  1/xₙ)
+     */
+    public static double harmonicMean(double[] numbers) {
```
vil02, line 43:
```suggestion
    public static double harmonicMean(final double[] numbers) {
```

src/main/java/com/thealgorithms/maths/Mean.java
```diff
@@ -0,0 +1,54 @@
...
+        double sum = 0;
+        if (numbers.length == 0) {
+            return sum;
```
vil02, line 46:
```suggestion
            return Double.NAN;
```

vil02 (changes requested):
General remarks:
- treating the cases of empty input: I suggested to return [`Double.NAN`](https://docs.oracle.com/javase/8/docs/api/java/lang/Double.html#NaN). Other possibilities to consider: throw an [`IllegalArgumentException`](https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html) or use [`Optional`](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) as a return type.
- naming: what do you think to rename this class to `Means` and/or shorten the method names to `arithmetic`, `geometric`, `harmonic`?

- input type: does it make sense to expect a more general input than `double[]`? Not only in terms of the type, but also to allow more containers (any _Iterable_?).

- implementation: what do you think about using `Stream.reduce()` like approach, instead of having explicit loops? For example, `arithmeticMean` could be rewritten as:

    ```java
    public static double arithmeticMean(final double[] numbers) {
        if (numbers.length == 0) {
            return Double.NAN;
        }
        final var sum = Arrays.stream(numbers).sum();
        return sum / numbers.length;
    }
    ```
    and `geometricMean` as
    ```java
    public static double geometricMean(double[] numbers) {
        if (numbers.length == 0) {
            return Double.NAN;
        }
        final var product = Arrays.stream(numbers).reduce(1.0, (a, b) -> a * b);
        return Math.pow(product, 1.0 / numbers.length);
    }
    ```
    Make sure to import `java.util.stream.Stream` and `java.util.Arrays`.

src/main/java/com/thealgorithms/maths/Mean.java
punit324: Sure makes sense.

src/main/java/com/thealgorithms/maths/Mean.java
punit324: Will do that.

src/main/java/com/thealgorithms/maths/Mean.java
punit324: Will do that.

punit324:
@vil02 I consider all your suggestion and will incorporate them.
Except one, where you are suggesting to use some iterable for input type other that `double[]`.
As `Mean` can only be computed for numbers, `double` can hold all the numbers including `intergers` and `floats`. 
If you want I can change it to `Array<Double>` class.

siriak:
@vil02 please review again

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,41 @@
...
+ * by: Punit Patel
+ */
+public final class Means {
+
+    private Means() {
+    }
+
+    /**
+     * Arithmetic / Pythagorean Mean = (x₁ + x₂ + ... +  xₙ) / n
+     */
```
vil02, line 17: I would avoid formulas in the doc-strings. What do you think about the following:
```suggestion
    /**
     * @brief computes the [arithmetic mean](https://en.wikipedia.org/wiki/Arithmetic_mean) of the input
     * @param numbers the input numbers
     * @throws IllegalArgumentException empty input
     * @return the arithmetic mean of the input numbers
     */
```

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,41 @@
...
+    public static Double arithmetic(final List<Double> numbers) {
+        if (numbers.isEmpty()) throw new IllegalArgumentException("Emtpy list given for Mean computation.");
```
vil02, line 19:
```suggestion
        if (numbers.isEmpty()) {
            throw new IllegalArgumentException("Emtpy list given for Mean computation.");
        }
```

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,41 @@
...
+
+        return numbers.stream().reduce((x, y) -> x + y).get() / numbers.size();
+    }
+
+    /**
+     * Geometric Mean = (x₁ · x₂ · ... · xₙ) ^ 1/n
+     */
+    public static Double geometric(final List<Double> numbers) {
+        if (numbers.isEmpty()) throw new IllegalArgumentException("Emtpy list given for Mean computation.");
```
vil02, line 28:
```suggestion
        if (numbers.isEmpty()) {
            throw new IllegalArgumentException("Emtpy list given for Mean computation.");
        }
```

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,41 @@
...
+
+        return Math.pow(numbers.stream().reduce((x, y) -> x * y).get(), 1.0 / numbers.size());
+    }
+
+    /**
+     * Harmonic Mean = n / (1/x₁ + 1/x₂ + ... +  1/xₙ)
+     */
+    public static double harmonic(final List<Double> numbers) {
+        if (numbers.isEmpty()) throw new IllegalArgumentException("Emtpy list given for Mean computation.");
```
vil02, line 37:
```suggestion
        if (numbers.isEmpty()) {
            throw new IllegalArgumentException("Emtpy list given for Mean computation.");
        }
```

vil02 (changes requested):
There was nothing bad about `double[] numbers` as input. My concern was that these functions could not be used, when `numbers` are `ArrayList<Double>`, `LinkedList<Double>`, `Vector<Double>`, `TreeSet<Double>` or any other from many more.

For example one could consider rewriting the method to something like:
```java
public static Double geometric(final Iterable<Double> numbers) {
    var geometricConsumer = new Object() {
        int count = 0;
        double product = 1.0;
    };
    numbers.forEach(number -> {
        geometricConsumer.product *= number;
        ++geometricConsumer.count;
        //this could become something like geometricConsumer.update(number);
    });
    
    if (geometricConsumer.count == 0) {
        throw new IllegalArgumentException("Empty iterable given for Mean computation.");
    }

    return Math.pow(geometricConsumer.product, 1.0 / geometricConsumer.count);
    // and the logic above could be expressed as geometricConsumer.getValue();
}
```
Well... this will not work _directly_ with `double[]`.

punit324:
@vil02 
I've incorporated all your suggestions.
Please review again.

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,57 @@
+package com.thealgorithms.maths;
+
+import java.util.stream.StreamSupport;
+import org.apache.commons.collections4.IterableUtils;
```
vil02, line 4: I am not sure what is the policy regarding introducing new _external_ dependencies. @siriak can you comment on that?

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * @brief computes the [Arithmetic Mean](https://en.wikipedia.org/wiki/Arithmetic_mean) of the input
+     * @param numbers the input numbers
+     * @throws IllegalArgumentException empty input
+     * @return the arithmetic mean of the input numbers
+     */
+    public static Double arithmetic(final Iterable<Double> numbers) {
+        if (!numbers.iterator().hasNext()) {
+            throw new IllegalArgumentException("Emtpy list given for Mean computation.");
+        }
```
vil02, line 25: This piece is repeated in all three methods. I would suggest to add some method like
```java
private static void checkIfNotEmpty(final Iterable<Double> numbers) {
    if (!numbers.iterator().hasNext()) {
        throw new IllegalArgumentException("Emtpy list given for Mean computation.");
    }
}
```
and call if in every _mean_ method.

Side question: do you have some idea how to _enforce_ such check in methods calculating different methods added in the future?

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param numbers the input numbers
+     * @throws IllegalArgumentException empty input
+     * @return the harmonic mean of the input numbers
+     */
+    public static double harmonic(final Iterable<Double> numbers) {
+        if (!numbers.iterator().hasNext()) {
+            throw new IllegalArgumentException("Emtpy list given for Mean computation.");
+        }
+
+        return IterableUtils.size(numbers) / StreamSupport.stream(numbers.spliterator(), false).map(x -> 1 / x).reduce((x, y) -> x + y).get();
```
vil02, line 55: I think this can be simplified to
```suggestion
        return IterableUtils.size(numbers) / StreamSupport.stream(numbers.spliterator(), false).reduce(0.0, (x, y) -> x + 1d / y);
```

vil02 (changes requested):
What is the time complexity of `IterableUtils.size(numbers)`? I think the _iterator_ has to be traversed (i.e. in order to compute any mean, we traverse the _iterator_ twice). @siriak can we live with that?

src/main/java/com/thealgorithms/maths/Means.java
punit324: If iterable is a collector then time complexity is O(1). And if its not a collection then worst case O(n).

vil02: Thanks. You are [right](https://commons.apache.org/proper/commons-collections/apidocs/src-html/org/apache/commons/collections4/IterableUtils.html#line.806) regarding the time complexity.

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @brief computes the [Geometric Mean](https://en.wikipedia.org/wiki/Geometric_mean) of the input
+     * @param numbers the input numbers
+     * @throws IllegalArgumentException empty input
+     * @return the geometric mean of the input numbers
+     */
+    public static Double geometric(final Iterable<Double> numbers) {
+        checkIfNotEmpty(numbers);
+        return Math.pow(StreamSupport.stream(numbers.spliterator(), false).reduce((x, y) -> x * y).get(), 1.0 / IterableUtils.size(numbers));
```
vil02, line 35:
```suggestion
        return Math.pow(StreamSupport.stream(numbers.spliterator(), false).reduce((x, y) -> x * y).get(), 1d / IterableUtils.size(numbers));
```

src/main/java/com/thealgorithms/maths/Means.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @brief computes the [Harmonic Mean](https://en.wikipedia.org/wiki/Harmonic_mean) of the input
+     * @param numbers the input numbers
+     * @throws IllegalArgumentException empty input
+     * @return the harmonic mean of the input numbers
+     */
+    public static Double harmonic(final Iterable<Double> numbers) {
+        checkIfNotEmpty(numbers);
+        return IterableUtils.size(numbers) / StreamSupport.stream(numbers.spliterator(), false).reduce(0.0, (x, y) -> x + 1 / y);
```
vil02, line 46:
```suggestion
        return IterableUtils.size(numbers) / StreamSupport.stream(numbers.spliterator(), false).reduce(0d, (x, y) -> x + 1d / y);
```

vil02 (changes requested):
The code and the tests look good for me (lets wait for @siriak to clarify things mentioned above).

The suggestions below are about the _literals_. To make this submission perfect, I would suggest to use _doubles_ (in `harmonic` you have an `int`). You can choose which notation do you prefer (`0.0` vs. `0.` vs. `0d` vs. `0.d`, vs. `0D` ...).

src/main/java/com/thealgorithms/maths/Means.java
punit324: Although I've int in harmonic, result will yeild double value.

src/main/java/com/thealgorithms/maths/Means.java
punit324: Although I've int in harmonic, result will yeild double value.

vil02: Why  not to have the right type from the very beginning?

punit324: To make it consistent we can do that.

vil02 (approved):
Looks good!

@siriak please have a look at the things mentioned above (like [new dependencies](https://github.com/TheAlgorithms/Java/pull/4339#discussion_r1313254771)).

src/main/java/com/thealgorithms/maths/Means.java
siriak: We don't add new dependencies very often, but I think it's a reasonable one to have.

siriak (approved):
Looks good, thanks!

PR #4346, opened 2023-09-01
add Binary Xor Algo

BamaCharanChhandogi:


siriak:
I wouldn't do a xor like that, it's way too inefficient. At the same time, the algorithm doesn't show how to do xor operation because it's too complicated and mostly consists of string manipulations. I don't think there's a way to implement xor so that it's good for learning and not terribly inefficient

BamaCharanChhandogi:
What is Your approach to solving this problem?

siriak:
Not to implement it. Writing Xor algorithm is just writing ^ character in Java. And for an explanation of the operation itself people can read a textbook, we don't heed to have it here.

Issue #4350, opened 2023-09-04
Fet: Add Single Bit Manipulation Operations


BamaCharanChhandogi:
### What would you like to Propose?

- [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [x] This pull request is all my own work -- I have not plagiarized it.
- [x] All filenames are in PascalCase.
- [x] All functions and variable names follow Java naming conventions.
- [x] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.

### Issue details

Single Bit Manipulation Operations:
want to add 3 functionalities **GetBit, SetBit, and ClearBit.**

peeplika:
hey i want to work on this , can it be assigned to me ?

BamaCharanChhandogi:
Hi, @lukasb1b,
All functions should be included in one file named 'Single Bit Manipulation,' like this. That would be better.

vil02:
I would rather suggest to create a class `BitArray` (with length passed in the constructor). The underlying data type should be `int[]` or `char[]` (maybe the `BitArray` should be a _generic type_?) - I am not sure what is better. Some benchmark should be performed.

Such class should have methods like `SetBit`, `ClearBit`, `FlipBit` etc.

Honestly that would be a quite fun task - I can support you here.

The interesting thing: this will not be an _utility class_!

BamaCharanChhandogi:
That's a really great idea.
I will also support. Go on.

PR #4356, opened 2023-09-07
Set bit

lukasb1b:

### DIFF
--- a/src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
+++ b/src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
@@ -1,0 +1,5 @@
+public class GetBit {
+    public static int getBit(int num, int get) {
+        return (num>>get) & 1;
+    }
+}
--- a/src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
+++ b/src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
@@ -5,6 +5,6 @@ package com.thealgorithms.bitmanipulation;
 
 public class GetBit {
     public static int getBit(int num, int get) {
-        return (num>>get) & 1;
+        return (num >> get) & 1;
     }
 }
--- a/src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
+++ b/src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
@@ -1,3 +1,8 @@
+package com.thealgorithms.bitmanipulation;
+/**
+ * Returns the value of bit from num located at get
+ */
+
 public class GetBit {
     public static int getBit(int num, int get) {
         return (num>>get) & 1;
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @siriak @yanglbme @debasishbsws @albina-astr
+* @siriak @yanglbme @debasishbsws @vil02 @BamaCharanChhandogi
--- a/src/main/java/com/thealgorithms/bitmanipulation/ClearBit.java
+++ b/src/main/java/com/thealgorithms/bitmanipulation/ClearBit.java
@@ -1,0 +1,11 @@
+package com.thealgorithms.bitmanipulation;
+/**
+ * Clears the bit located at clear from num
+ */
+
+public class ClearBit {
+    public static int clearBit(int num, int clear) {
+        int mask = ~(1 << clear);
+        return num & mask;
+    }
+}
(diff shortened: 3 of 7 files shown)


lukasb1b:
Issue #4350

lukasb1b:
I want to add to the Issue that SetBit is already implemented and ClearBit is in other pullrequest

siriak:
@vil02 @BamaCharanChhandogi please review

BamaCharanChhandogi:
> I want to add to the Issue that SetBit is already implemented and ClearBit is in other pullrequest

Hi, @lukasb1b,
All functions of single bit manipulation should be included in one file named 'Single Bit Manipulation,'. That would be better.

vil02:
@siriak, @BamaCharanChhandogi, @lukasb1b: please have a look at [my comment in the corresponding issue](https://github.com/TheAlgorithms/Java/issues/4350#issuecomment-1710635231).

src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
```diff
@@ -0,0 +1,10 @@
+package com.thealgorithms.bitmanipulation;
+/**
+ * Returns the value of bit from num located at get
+ */
+
+public class GetBit {
```
vil02, line 6: This is a very good start for the [`BitArray` class](https://github.com/TheAlgorithms/Java/issues/4350#issuecomment-1710635231).

vil02 (changes requested):

src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
siriak: I suggest adding this as-is and consolidating algorithms into a data structure in a following PR. There are already some other functions implemented in this manner and, as you have mentioned, there are open questions about the design of this BitArray.
@lukasb1b will you be able to consolidate these bit functions into a BitArray in the next PR? If no, somebody else could pick up this task.
In the meantime, we can discuss what we want to use as an underlying type.

src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
```diff
@@ -0,0 +1,10 @@
+package com.thealgorithms.bitmanipulation;
+/**
+ * Returns the value of bit from num located at get
+ */
+
+public class GetBit {
+    public static int getBit(int num, int get) {
+        return (num >> get) & 1;
+    }
+}
```
vil02, line 10: In any case such functionality will be needed.

```suggestion
public final class GetBit {
    private GetBit() {
    }
    public static boolean get(final int data, final int pos) {
        if (pos < 0 || pos >= Integer.SIZE) {
            throw new ArrayIndexOutOfBoundsException();
        }
        return ((data >> pos) & 1) > 0;
    }
}
```

I am not sure what is the best way to make it accept any `IntegralType` - I could not find a convenient way how to get the `SIZE`.

Please add tests for the missing _branches_.

vil02 (changes requested):

src/main/java/com/thealgorithms/bitmanipulation/GetBit.java
siriak: Let's stick to int32 then

lukasb1b:
I can add all the SingleBit functions in onte but at first i think we should have a solution for the Problem i personally think a bit array is not as functional but we could do it generic like in the example below. And we add another function for arrays.

public class GetBit<T extends Number>{

    public static <T extends Number> T getBit(T num, int get) throws Exception {
        if(num instanceof Integer){
            return (T)(Integer.valueOf((num.intValue()>>get) & 1));
        }else{
            throw new Exception();
        }
    }

vil02:
> I can add all the SingleBit functions in onte but at first i think we should have a solution for the Problem i personally think a bit array is not as functional but we could do it generic like in the example below. And we add another function for arrays.
> 
> public class GetBit{
> 
> ```
> public static <T extends Number> T getBit(T num, int get) throws Exception {
>     if(num instanceof Integer){
>         return (T)(Integer.valueOf((num.intValue()>>get) & 1));
>     }else{
>         throw new Exception();
>     }
> }
> ```

As [suggested before](https://github.com/TheAlgorithms/Java/pull/4356#discussion_r1319772921), let stick to `int` for now. If you agree with [my suggestion](https://github.com/TheAlgorithms/Java/pull/4356#discussion_r1319405642), please apply it and add missing tests.

I agree that it is a good idea to have one _utilitiy class_ like `BitOperations` with `public static` methods `get` and `set` (this will be needed anyway in the `BitArray`).

lukasb1b:
Hi I closed the request so you can implement it I dont get how you want to do it so I think its quicker if you do it yourself

Issue #4367, opened 2023-09-11
Fet : Enhance knapsack algorithm code


Manan-09:
### What would you like to Propose?

There is scope of enhancement in code of knapsack problem. 
We can reduce space complexity of solution from O(n*w) to O(w)
We can name variables better
We can add more testcases and move to testcase to different test file.

I would like to contribute enhancing this.

### Issue details

Space complexity can be improved , current testcases is not covering all code.

Manan-09:
Raised PR https://github.com/TheAlgorithms/Java/pull/4368

PR #4368, opened 2023-09-11
#4367 Enhance Knapsack problem

Manan-09:

Reduced space complexity of solution from O(n*w) to O(w)
Named variables better
Added testfile

All code is covered in testcases
Screenshot 2023-09-11 at 2 03 35 PM


src/test/java/com/thealgorithms/dynamicprogramming/KnapsackTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Test
+    public void testKnapSackInvalidInput() {
+        int[] weights = {2, 3, 4};
+        int[] values = {3, 4, 5, 6}; // Different length values array.
+        int weightCapacity = 5;
+        // This test should throw an IllegalArgumentException.
+        assertThrows(IllegalArgumentException.class, () -> { Knapsack.knapSack(weightCapacity, weights, values); });
+    }
+}
```
vil02:
```suggestion
}

```

This will solve the `clang-format` failure.

vil02 (changes requested):

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,40 @@
...
- * A DynamicProgramming based solution for 0-1 Knapsack problem
+ * A Dynamic Programming based solution for the 0-1 Knapsack problem.
+ * This class provides a method, `knapSack`, that calculates the maximum value that can be
+ * obtained from a given set of items with weights and values, while not exceeding a
+ * given weight capacity.
+ *
+ * @see <a href="https://en.wikipedia.org/?title=0-1_Knapsack_problem">0-1 Knapsack Problem </a>
  */
-public class Knapsack {
+public final class Knapsack {
```
vil02, line 11:
```suggestion
public final class Knapsack {
    private Knapsack() {
    }
```

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,40 @@
...
+     * Solves the 0-1 Knapsack problem using Dynamic Programming.
+     *
+     * @param weightCapacity The maximum weight capacity of the knapsack.
+     * @param weights        An array of item weights.
+     * @param values         An array of item values.
+     * @return The maximum value that can be obtained without exceeding the weight capacity.
+     * @throws IllegalArgumentException If the input arrays are null or have different lengths.
+     */
+    public static int knapSack(final int weightCapacity, final int[] weights, final int[] values) throws IllegalArgumentException {
+        if (weights == null || values == null || weights.length != values.length) {
```
vil02, line 23: Is it really needed to check if `weights` or `values` are `null`?

src/test/java/com/thealgorithms/dynamicprogramming/KnapsackTest.java
```diff
@@ -0,0 +1,56 @@
+package com.thealgorithms.dynamicprogramming;
+
+import static org.junit.jupiter.api.Assertions.*;
```
vil02, line 3:
```suggestion
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertThrows;
```

vil02 (changes requested):
There are two other solutions of that problem: [DyanamicProgrammingKnapsack](https://github.com/TheAlgorithms/Java/blob/94621fb63eafcf18666dfad176fffc28f82b439a/src/main/java/com/thealgorithms/dynamicprogramming/DyanamicProgrammingKnapsack.java#L5) and [KnapsackMemoization.java](https://github.com/TheAlgorithms/Java/blob/94621fb63eafcf18666dfad176fffc28f82b439a/src/main/java/com/thealgorithms/dynamicprogramming/KnapsackMemoization.java#L12) - what is the relation between them and the one you update?

Besides that I suggest some _cosmetic_ changes.

Manan-09:
Seems like this is  [DyanamicProgrammingKnapsack](https://github.com/TheAlgorithms/Java/blob/94621fb63eafcf18666dfad176fffc28f82b439a/src/main/java/com/thealgorithms/dynamicprogramming/DyanamicProgrammingKnapsack.java#L5) duplicate of what I just refactored, and  [KnapsackMemoization.java](https://github.com/TheAlgorithms/Java/blob/94621fb63eafcf18666dfad176fffc28f82b439a/src/main/java/com/thealgorithms/dynamicprogramming/KnapsackMemoization.java#L12) is recursive solition of this problem.
But my solution has better space complexity with same TC

debasishbsws (dismissed):
Looks good👍

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
vil02: @Manan-09 why did you mark this as resolved?

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
vil02: @Manan-09 why did you mark this as resolved?

vil02:
> Seems like this is [DyanamicProgrammingKnapsack](https://github.com/TheAlgorithms/Java/blob/94621fb63eafcf18666dfad176fffc28f82b439a/src/main/java/com/thealgorithms/dynamicprogramming/DyanamicProgrammingKnapsack.java#L5) duplicate of what I just refactored, and [KnapsackMemoization.java](https://github.com/TheAlgorithms/Java/blob/94621fb63eafcf18666dfad176fffc28f82b439a/src/main/java/com/thealgorithms/dynamicprogramming/KnapsackMemoization.java#L12) is recursive solition of this problem. But my solution has better space complexity with same TC

@siriak how should we proceed?

siriak:
Please remove the duplicate, but leave the version with memoization as it's a different approach

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
siriak: I think it's a reasonable check. There's nothing else to prevent them from being null

siriak (approved):

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
vil02: In this case I would highly recommend adding two more tests where the input arrays are `null` (i.e. one with `weights` being `null` and second with `values` being `null`).

src/test/java/com/thealgorithms/dynamicprogramming/KnapsackTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    @Test
+    public void testKnapSackInvalidInput() {
+        int[] weights = {2, 3, 4};
+        int[] values = {3, 4, 5, 6}; // Different length values array.
+        int weightCapacity = 5;
+        // This test should throw an IllegalArgumentException.
+        assertThrows(IllegalArgumentException.class, () -> { Knapsack.knapSack(weightCapacity, weights, values); });
+    }
```
vil02, line 56:
```suggestion
    @Test
    public void testKnapSackThrowsForInputsOfDifferentLength() {
        int[] weights = {2, 3, 4};
        int[] values = {3, 4, 5, 6}; // Different length values array.
        int weightCapacity = 5;
        assertThrows(IllegalArgumentException.class, () -> { Knapsack.knapSack(weightCapacity, weights, values); });
    }
```

src/test/java/com/thealgorithms/dynamicprogramming/KnapsackTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    public void testKnapSackNullInput() {
+        int[] weights = {2, 3, 4};
+        int[] values = {3, 4, 6};
+        int weightCapacity = 5;
+        // This test should throw an IllegalArgumentException.
+        assertThrows(IllegalArgumentException.class, () -> { Knapsack.knapSack(weightCapacity, null, values); });
+        assertThrows(IllegalArgumentException.class, () -> { Knapsack.knapSack(weightCapacity, weights, null); });
+    }
```
vil02, line 66:
```suggestion
    @Test
    public void testKnapSackThrowsForNullInputs() {
        int[] weights = {2, 3, 4};
        int[] values = {3, 4, 6};
        int weightCapacity = 5;
        assertThrows(IllegalArgumentException.class, () -> { Knapsack.knapSack(weightCapacity, null, values); });
        assertThrows(IllegalArgumentException.class, () -> { Knapsack.knapSack(weightCapacity, weights, null); });
    }
```

vil02 (changes requested):
I would suggest to use _meaningful_ test names instead of comments. Names will appear in the output in case if they fail, Comments will not.

Thanks for adding missing tests!

Besides that: the code looks good.

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,42 @@
...
+    /**
+     * Solves the 0-1 Knapsack problem using Dynamic Programming.
+     *
+     * @param weightCapacity The maximum weight capacity of the knapsack.
+     * @param weights        An array of item weights.
+     * @param values         An array of item values.
+     * @return The maximum value that can be obtained without exceeding the weight capacity.
+     * @throws IllegalArgumentException If the input arrays are null or have different lengths.
+     */
+    public static int knapSack(final int weightCapacity, final int[] weights, final int[] values) throws IllegalArgumentException {
```
vil02, line 24: What if `weightCapacity` would be negative? Since it is checked if `weights` and `values` are not `null`, I think there should also be a check if provided `weightCapacity` is not negative.

vil02 (changes requested):

vil02 (changes requested):
What if one of the entries of `weights` is negative or `0`?

```java
Knapsack.knapSack(1, new int[]{0, 0, 0, 0}, new int[]{1, 1, 1, 1})
```
gives me `16` and I think the correct answer should be `4`.

Manan-09:
> What if one of the entries of `weights` is negative or `0`?
> 
> The following code:
> 
> ```java
> Knapsack.knapSack(1, new int[]{0, 0, 0, 0}, new int[]{1, 1, 1, 1})
> ```
> 
> gives me `16` and I think the correct answer should be `4`.

Knapsack problem have positive weights normally, should we add extra check for non positive weights ?

vil02:
> Knapsack problem have positive weights normally, should we add extra check for non positive weights ?
Yes! Throwing an error is the best way of saying that non-positive weights are forbidden.

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,51 @@
...
+ * @see <a href="https://en.wikipedia.org/?title=0-1_Knapsack_problem">0-1 Knapsack Problem </a>
  */
-public class Knapsack {
+public final class Knapsack {
 
-    private static int knapSack(int W, int[] wt, int[] val, int n) throws IllegalArgumentException {
-        if (wt == null || val == null) {
-            throw new IllegalArgumentException();
+    private Knapsack(){}
+    ;
```
vil02, line 14: This should solve the `clang-tidy` error.

```suggestion
    private Knapsack(){}
```

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,51 @@
...
+            throw new IllegalArgumentException("Weight capacity should not be negative.");
+        }
+        if (weights == null || values == null || weights.length != values.length) {
+            throw new IllegalArgumentException("Input arrays must not be null and must have the same length.");
+        }
+        for (int w : weights) {
+            if (w <= 0) {
+                throw new IllegalArgumentException("Input array should not contain non-positive weight(s).");
+            }
         }
```
vil02, line 36: What do you think about moving it into some method with a signature like 
```java
private static void checkInput(final int weightCapacity, final int[] weights, final int[] values);
```

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,51 @@
...
+        if (weightCapacity < 0) {
+            throw new IllegalArgumentException("Weight capacity should not be negative.");
+        }
+        if (weights == null || values == null || weights.length != values.length) {
+            throw new IllegalArgumentException("Input arrays must not be null and must have the same length.");
+        }
+        for (int w : weights) {
+            if (w <= 0) {
+                throw new IllegalArgumentException("Input array should not contain non-positive weight(s).");
+            }
```
vil02, line 35:
```suggestion
        if (Arrays.stream(weights).anyMatch(w -> w <= 0)) {
            throw new IllegalArgumentException("Input array should not contain non-positive weight(s).");
        }
```

This change requires `import java.util.Arrays;`.

src/test/java/com/thealgorithms/dynamicprogramming/KnapsackTest.java
```diff
@@ -0,0 +1,81 @@
...
+    public void testKnapSackThrowsForNegativeCapacity() {
+        int[] weights = {2, 3, 4, 5};
+        int[] values = {3, 4, 5, 6};
+        int weightCapacity = -5;
+        assertThrows(IllegalArgumentException.class, () -> { Knapsack.knapSack(weightCapacity, weights, values); });
+    }
+
+    @Test
+    public void testKnapSackThrowsForNegativeWeight() {
+        int[] weights = {2, -3, 4};
```
vil02, line 76:
```suggestion
        int[] weights = {2, 0, 4};
```

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,51 @@
...
-        for (i = 0; i <= n; i++) {
-            for (w = 0; w <= W; w++) {
-                if (i == 0 || w == 0) {
-                    rv[i][w] = 0;
-                } else if (wt[i - 1] <= w) {
-                    rv[i][w] = Math.max(val[i - 1] + rv[i - 1][w - wt[i - 1]], rv[i - 1][w]);
-                } else {
-                    rv[i][w] = rv[i - 1][w];
+        // DP table to store the state of the maximum possible return for a given weight capacity.
+        int[] dp = new int[weightCapacity + 1];
```
vil02, line 39: If you insist on having this comment, why not to create a method like `private static int[] initializeDpArray(final int weightCapacity)`;

If you decide to have it, please check here if `weightCapacity >= 0` (not in the method `checkInput` mentioned earlier).

vil02 (changes requested):

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,50 @@
...
+ *
+ * @see <a href="https://en.wikipedia.org/?title=0-1_Knapsack_problem">0-1 Knapsack Problem </a>
  */
-public class Knapsack {
+public final class Knapsack {
 
-    private static int knapSack(int W, int[] wt, int[] val, int n) throws IllegalArgumentException {
-        if (wt == null || val == null) {
-            throw new IllegalArgumentException();
+    private Knapsack(){}
```
vil02, line 13:
```suggestion
    private Knapsack(){
    }
```

vil02 (changes requested):

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,51 @@
...
+ * @see <a href="https://en.wikipedia.org/?title=0-1_Knapsack_problem">0-1 Knapsack Problem </a>
  */
-public class Knapsack {
+public final class Knapsack {
 
-    private static int knapSack(int W, int[] wt, int[] val, int n) throws IllegalArgumentException {
-        if (wt == null || val == null) {
-            throw new IllegalArgumentException();
+    private Knapsack(){
+    }
```
vil02, line 14:
```suggestion
    private Knapsack() {
    }
```

vil02 (changes requested):

BamaCharanChhandogi (dismissed):

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
```diff
@@ -1,38 +1,56 @@
...
-public class Knapsack {
+public final class Knapsack {
+
+    private Knapsack() {
+    }
 
-    private static int knapSack(int W, int[] wt, int[] val, int n) throws IllegalArgumentException {
-        if (wt == null || val == null) {
-            throw new IllegalArgumentException();
+    // Validate input for knapsack problem, otherwise throws exception
```
vil02, line 18:
```suggestion
```
I am not sure if this comment is needed. If you insist on having it, please use _doxygen_-like comment.

src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
Manan-09: Pardon @vil02  , which comment you are talking about?

Manan-09: Sorry @vil02, looks like already replied to your comment but it was in Pending state. Please check my [here](https://github.com/TheAlgorithms/Java/pull/4368#discussion_r1328132121)
Please tell if any changes needed

vil02: > I belive creating other method initializeDpArray will be overkill,

Understood.

> and why can't we check weightCapacity < 0 in our validation method?

If the method `initializeDpArray` would be introduced, I would suggest to make the `weightCapacity < 0` there.

vil02 (approved):

Issue #4382, opened 2023-09-19
[BUG] `MedianOfRunningArray` produces wrong output


vil02:
### Description

The implementation of [`MedianOfRunningArray`](https://github.com/TheAlgorithms/Java/blob/12b6c292438cac17a19154335bb6066676876b20/src/main/java/com/thealgorithms/misc/MedianOfRunningArray.java) has a bug.

### Steps to reproduce

The code:

```java
MedianOfRunningArray p = new MedianOfRunningArray();
p.insert(30);
p.insert(20);
p.insert(10);
System.out.print(p.median());
```

prints `10`.

### Excepted behavior

For such input, the output [should be `20`](https://www.wolframalpha.com/input?i=median+of+%5B30%2C+20%2C+10%5D).

### Additional context

The code:

```java
MedianOfRunningArray p = new MedianOfRunningArray();
p.insert(10);
p.insert(20);
p.insert(30);
System.out.print(p.median());
```

prints `20`.

vil02:
Closed by #4384.

PR #4384, opened 2023-09-20
#4382 Bug Fix

govardhanshah456:
Fixed The Bug for the Test Case mentioned in linked Issue as well as made code more readable


vil02 (changes requested):
@govardhanshah456 thanks for working the issue #4382. I really like that you decided to rename the `p1` and `p2` member variables.

govardhanshah456:
Hey, i have made requested changes, please have a look into it.

src/main/java/com/thealgorithms/misc/MedianOfRunningArray.java
```diff
@@ -8,34 +8,41 @@
...
-            return (p1.peek() + p2.peek()) / 2;
+    public Double median() {
+        if (maxHeap.size() == minHeap.size()) {
+            return (maxHeap.peek() + minHeap.peek()) / 2.0;
         }
-        return p1.size() > p2.size() ? p1.peek() : p2.peek();
+        return maxHeap.size() > minHeap.size() ? maxHeap.peek() * 1.0 : minHeap.peek() * 1.0;
     }
 
     public static void main(String[] args) {
```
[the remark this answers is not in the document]
govardhanshah456: Hey, i have made the required changes.

src/main/java/com/thealgorithms/misc/MedianOfRunningArray.java
```diff
@@ -8,46 +8,42 @@
...
-        /*
-        Testing the median function
-         */
-
-        MedianOfRunningArray p = new MedianOfRunningArray();
-        int[] arr = {10, 7, 4, 9, 2, 3, 11, 17, 14};
-        for (int i = 0; i < 9; i++) {
-            p.insert(arr[i]);
-            System.out.print(p.median() + " ");
+    public Double median() {
```
vil02, line 41: Is there a reason why `Double` (**not** `double`) is used here?

```suggestion
    public double median() {
```

src/main/java/com/thealgorithms/misc/MedianOfRunningArray.java
```diff
@@ -8,46 +8,42 @@
...
+        if (maxHeap.size() == minHeap.size() && maxHeap.size() > 0) {
+            return (maxHeap.peek() + minHeap.peek()) / 2.0;
+        } else if (maxHeap.size() == minHeap.size()) {
+            throw new IllegalArgumentException("Enter at least 1 element, Median of empty list is not defined!");
         }
+        return maxHeap.size() > minHeap.size() ? maxHeap.peek() * 1.0 : minHeap.peek() * 1.0;
```
vil02, line 47: I think one can simplify the logic a bit

```suggestion
        if (maxHeap.isEmpty() && minHeap.isEmpty()) {
            throw new IllegalArgumentException("Enter at least 1 element, Median of empty list is not defined!");
        }

        if (maxHeap.size() == minHeap.size()) {
            return (maxHeap.peek() + minHeap.peek()) / 2.0;
        }
        return maxHeap.size() > minHeap.size() ? maxHeap.peek() * 1.0 : minHeap.peek() * 1.0;
```

src/test/java/com/thealgorithms/misc/MedianOfRunningArrayTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public void testWithDuplicateValues() {
+        MedianOfRunningArray stream = new MedianOfRunningArray();
+        stream.insert(-1);
+        assertEquals(-1, stream.median());
+        stream.insert(-1);
+        assertEquals(-1, stream.median());
+        stream.insert(-1);
+        assertEquals(-1, stream.median());
+    }
+
```
vil02, line 92:
```suggestion

    @Test
    public void testWithDuplicateValuesB() {
        MedianOfRunningArray stream = new MedianOfRunningArray();
        stream.insert(10);
        stream.insert(20);
        stream.insert(10);
        stream.insert(10);
        stream.insert(20);
        stream.insert(0);
        stream.insert(50);
        assertEquals(10, stream.median());
    }

```

src/test/java/com/thealgorithms/misc/MedianOfRunningArrayTest.java
```diff
@@ -0,0 +1,112 @@
...
+        assertEquals(1650000.00, stream.median());
+    }
+
+    @Test
+    public void testWithLargeCountOfValues() {
+        MedianOfRunningArray stream = new MedianOfRunningArray();
+        for (int i = 1; i <= 1000; i++) stream.insert(i);
+        assertEquals(500.5, stream.median());
+    }
+}
```
vil02, line 112:
```suggestion

    @Test
    public void testWithThreeValuesInDescendingOrder() {
        MedianOfRunningArray stream = new MedianOfRunningArray();
        stream.insert(30);
        stream.insert(20);
        stream.insert(10);
        assertEquals(20.0, stream.median());
    }

    @Test
    public void testWithThreeValuesInOrder() {
        MedianOfRunningArray stream = new MedianOfRunningArray();
        stream.insert(10);
        stream.insert(20);
        stream.insert(30);
        assertEquals(20.0, stream.median());
    }

    @Test
    public void testWithThreeValuesNotInOrderA() {
        MedianOfRunningArray stream = new MedianOfRunningArray();
        stream.insert(30);
        stream.insert(10);
        stream.insert(20);
        assertEquals(20.0, stream.median());
    }

    @Test
    public void testWithThreeValuesNotInOrderB() {
        MedianOfRunningArray stream = new MedianOfRunningArray();
        stream.insert(20);
        stream.insert(10);
        stream.insert(30);
        assertEquals(20.0, stream.median());
    }
}
```

vil02 (changes requested):
As a quick bug fix it looks really good. I suggest to add some tests (it is good to have a test case causing the original bug). You can use the `commit suggestion` button.

I know that this is out of scope for this PR, but I think it would be reasonable to make the `MedianOfRunningArray` a _generic type_, so one could compute a median of integers (and the result would be an integer), doubles (and the result would be a double) etc. @govardhanshah456 would you like to work on such an issue?

govardhanshah456:
Hey, as you said its out of scope for this PR, what we can do instead is we can make new enhancement issue of making Median Generic and then work on it, that would appear more appropriate, and yes median would always be double for instance, (1,2) median would be (1+2)/2 which is double although input is int array.

vil02 (approved):
Awesome @govardhanshah456!

Closes #4382.

PR #4388, opened 2023-09-21
negative FindMinTest

lukasb1b:

Changing FindMinTest so it tests for negativ numbers

### DIFF
--- a/src/test/java/com/thealgorithms/maths/FindMinTest.java
+++ b/src/test/java/com/thealgorithms/maths/FindMinTest.java
@@ -1,14 +1,13 @@
 package com.thealgorithms.maths;
 
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+import java.util.stream.Stream;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
-import static org.junit.jupiter.api.Assertions.assertThrows;
-
-
-import java.util.stream.Stream;
 
 public class FindMinTest {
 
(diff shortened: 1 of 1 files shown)


vil02:
@lukasb1b why do you think this test is needed? What _value_ does it add?

lukasb1b:
While it's not mandatory, I believe that a comprehensive test should encompass all available number ranges, including both positive and negative numbers.

vil02:
> While it's not mandatory, I believe that a comprehensive test should encompass all available number ranges, including both positive and negative numbers.

Does `FindMin` distinguish between positive and negative numbers? Is there a place there where negative numbers are traded in a different way than positive?

Why not to add a test, where all of the elements of the input array are divisible by 5?

lukasb1b:
It does not distingiguish but its a different numberrange as you differentiate between negative and "normal" numbers in math. So i thought it would be good to cover all numberranges the algorithem can use

vil02:
> It does not distingiguish but its a different numberrange as you differentiate between negative and "normal" numbers in math. So i thought it would be good to cover all numberranges the algorithem can use

So should one add a test where input array contains only numbers divisible by 71?

Here I really do not see any sense of the new test cases. There must be clear reasons to do that, e.g.:
- new test case increases the _coverage_ (more lines are covered, more branches are covered, a new (significant) _state_ is covered),
- new test case causes a bug.

If you are looking for a task, you could consider making the [`FindMin`](https://github.com/TheAlgorithms/Java/blob/fbe348b105a06258a81538023698117e2bb116cb/src/main/java/com/thealgorithms/maths/FindMin.java) a proper utility class (i.e. remove the `main` method etc.).

If you want to challenge yourself in creating tests (which sometimes might be very challenging) - please let me know. I will find some suitable example and make an issue for you.

siriak (commented):
Overall, it makes sense to have positive, negative numbers, and 0 in your test data even if this doesn't improve branch coverage. There can be other bugs like overflow or unsafe casts that while are not explicitly branches in flow, behave differently for positive and negative numbers.
Could you make it test data that's read by a test? So that we don't have to copy-paste the whole test every time we add a new test case.

lukasb1b:
Do you mean like that?

siriak (commented):
Not like that. I wanted to have a single `assertEquals` per test and array and expected result would be passed as arguments to the test. I can look up an example later, let me know if I need to do that.

lukasb1b:
@siriak like this?

siriak:
I want to make it a parameterized test https://www.baeldung.com/parameterized-tests-junit-5#first-impression

PR #4389, opened 2023-09-21
FindMin more efficient

lukasb1b:

This change schould make FindMin more efficient.


vil02:
I am skeptic about this change. I am not even sure if it really improves the performance. For example: in your code the `i` is a 32-bit integer - is it really the best way to iterate on a 64-bit system?

The generated code is indeed different: [old](https://godbolt.org/z/9ddr9Ea3s) vs. [new](https://godbolt.org/z/hW5Khr4E7).

This file has more important problems.

@siriak what do you think?

siriak:
I'd say that it's more efficient from an algorithmic perspective because it uses 1 iteration less. Not sure if this translates into real-world performance, though. Nevertheless, I like the change because we are concerned with algorithms and this is a good example of how to eliminate a redundant iteration (it's redundant because `array[i] < min` is always `true` on the first iteration)

siriak (approved):

PR #4392, opened 2023-09-21
Make `MedianOfRunningArray` Generic

### LINKED ISSUE #4390: [FEATURE REQUEST] Make `MedianOfRunningArray` generic

I find it a bit odd that the [`MedianOfRunningArray`](https://github.com/TheAlgorithms/Java/blob/fbe348b105a06258a81538023698117e2bb116cb/src/main/java/com/thealgorithms/misc/MedianOfRunningArray.java) only works with integers and returns double as the median. I suggest to make `MedianOfRunningArray` generic accepting any [`Number`](https://docs.oracle.com/javase/8/docs/api/java/lang/Number.html) class.

### Issue details

I suggest to define `MedianOfRunningArray` as

```java
public class MedianOfRunningArray<T extends Number> {
    //...
}
```

and to change the method `insert` to accept objects of `T` and `median` to return `T`.

### Additional Information

Please make sure to include the following test (_rounding down to the integer_):

```Java
var stream = new MedianOfRunningArray<Integer>();
stream.insert(1);
stream.insert(2);
assertEquals(1, stream.median());
```

govardhanshah456:
Closes #4390

### DIFF
--- a/src/main/java/com/thealgorithms/misc/MedianOfRunningArray.java
+++ b/src/main/java/com/thealgorithms/misc/MedianOfRunningArray.java
@@ -7,10 +7,10 @@ import java.util.PriorityQueue;
 /**
  * @author shrutisheoran
  */
-public class MedianOfRunningArray<T extends Number & Comparable<T>> {
+public class MedianOfRunningArray<Value extends Number & Comparable<Value>> {
 
-    private PriorityQueue<T> maxHeap;
-    private PriorityQueue<T> minHeap;
+    private PriorityQueue<Value> maxHeap;
+    private PriorityQueue<Value> minHeap;
 
     // Constructor
     public MedianOfRunningArray() {
@@ -22,8 +22,8 @@ public class MedianOfRunningArray<T extends Number & Comparable<T>> {
       Inserting lower half of array to max Heap
       and upper half to min heap
      */
-    public void insert(T e) {
-        if (!minHeap.isEmpty() && e.compareTo(minHeap.peek()) < 0) {
+    public void insert(Value e) {
+        if (!minHeap.isEmpty() && e.compareValueo(minHeap.peek()) < 0) {
             maxHeap.offer(e);
             if (maxHeap.size() > minHeap.size() + 1) {
                 minHeap.offer(maxHeap.poll());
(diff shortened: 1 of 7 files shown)


govardhanshah456:
@vil02 Hey, i did some changes according to my knowledge of generics. You can check it.

src/main/java/com/thealgorithms/misc/MedianOfRunningArray.java
```diff
@@ -38,14 +38,40 @@ public void insert(Integer e) {
...
+            int sum = ((Short) a) + ((Short) b);
+            return (T) Short.valueOf((short) (sum / 2));
+        } else if (a instanceof Byte) {
+            int sum = ((Byte) a) + ((Byte) b);
+            return (T) Byte.valueOf((byte) (sum / 2));
+        } else {
+            throw new IllegalArgumentException("Unsupported numeric type");
         }
-        return maxHeap.size() > minHeap.size() ? maxHeap.peek() * 1.0 : minHeap.peek() * 1.0;
     }
```
vil02, line 76: This is ugly. I think that it is fair to say that generics in Java are annoying. What do you think about defining the `MedianOfRunningArray` as abstract class (with `computeAverage` as an abstract method) and then have classes like `MedianOfRunningArrayInteger`, `MedianOfRunningArrayDouble` extending it?

In more details (I made some slight changes, please consider them):

```java
public abstract class MedianOfRunningArray<T extends Number & Comparable<T>> {

    private PriorityQueue<T> maxHeap = new PriorityQueue<>(Collections.reverseOrder());
    private PriorityQueue<T> minHeap = new PriorityQueue<>();

    /*
      Inserting lower half of array to max Heap
      and upper half to min heap
     */
    public void insert(final T e) {
        if (!minHeap.isEmpty() && e.compareTo(minHeap.peek()) < 0) {
            maxHeap.offer(e);
            if (maxHeap.size() > minHeap.size() + 1) {
                minHeap.offer(maxHeap.poll());
            }
        } else {
            minHeap.offer(e);
            if (minHeap.size() > maxHeap.size() + 1) {
                maxHeap.offer(minHeap.poll());
            }
        }
    }

    /*
      Returns median at any given point
     */
    public T median() {
        if (maxHeap.isEmpty() && minHeap.isEmpty()) {
            throw new IllegalArgumentException("Enter at least 1 element, Median of empty list is not defined!");
        } else if (maxHeap.size() == minHeap.size()) {
            T maxHeapTop = maxHeap.peek();
            T minHeapTop = minHeap.peek();
            return calculateAverage(maxHeapTop, minHeapTop);
        }
        return maxHeap.size() > minHeap.size() ? maxHeap.peek() : minHeap.peek();
    }

    public abstract T calculateAverage(T a, T b);
}
```

and

```java
final public class MedianOfRunningArrayDouble extends MedianOfRunningArray<Double> {
    @Override
    public Double calculateAverage(final Double a, final Double b) {
        return (a + b) / 2.0d;
    }
}
```

```java
final public class MedianOfRunningArrayInteger extends MedianOfRunningArray<Integer> {
    @Override
    public Integer calculateAverage(final Integer a, final Integer b) {
        return (a + b) / 2;
    }
}
```

Having `calculateAverage` as public will help with testing.

The advantage of this approach is that when someone wants to have a different rounding for integers it is enough to have `MedianOfRunningArrayIntegerCrazyRounding`.

src/test/java/com/thealgorithms/misc/MedianOfRunningArrayTest.java
```diff
@@ -17,71 +17,71 @@ public class MedianOfRunningArrayTest {
```
vil02, line 17: Please clean-up the imports and the comment/documentation. I forgot to write that in the previous PR. Sorry.

vil02 (changes requested):

govardhanshah456:
@vil02 Hey ,as suggested I have made class abstract and did the required changes, have a look.

src/main/java/com/thealgorithms/misc/MedianOfRunningArrayInteger.java
```diff
@@ -0,0 +1,8 @@
+package com.thealgorithms.misc;
+
+final public class MedianOfRunningArrayInteger extends MedianOfRunningArray<Integer> {
```
vil02, line 3:
```suggestion
public final class MedianOfRunningArrayInteger extends MedianOfRunningArray<Integer> {
```

src/main/java/com/thealgorithms/misc/MedianOfRunningArrayLong.java
```diff
@@ -0,0 +1,8 @@
+package com.thealgorithms.misc;
+
+final public class MedianOfRunningArrayLong extends MedianOfRunningArray<Long> {
```
vil02, line 3:
```suggestion
public final class MedianOfRunningArrayLong extends MedianOfRunningArray<Long> {
```

src/test/java/com/thealgorithms/misc/MedianOfRunningArrayTest.java
vil02: `import static org.junit.jupiter.api.Assertions.assertFalse;` and `import org.apache.commons.lang3.tuple.Pair;` are not needed.

The comments
```
/**
 * Test case for Two sum Problem.
 * @author Bama Charan Chhandogi (https://github.com/BamaCharanChhandogi)
 */
```
seems to be wrong.

vil02 (changes requested):
It goes in a very good direction!

govardhanshah456:
@vil02 Hey, modified that comments and removed that impot and committed suggested changes, regarding double  and float i did not know that there exist some way to check precision using assertEqual itself, so was formatting at output side.

vil02 (dismissed):
Looks good. Thanks!

@siriak could you please have a look? There were _lots_ of new files introduced and I would like to double check if the current structure is correct.

src/main/java/com/thealgorithms/misc/MedianOfRunningArrayFloat.java
```diff
@@ -0,0 +1,7 @@
+package com.thealgorithms.misc;
+public final class MedianOfRunningArrayFloat extends MedianOfRunningArray<Float> {
```
vil02, line 2:
```suggestion

public final class MedianOfRunningArrayFloat extends MedianOfRunningArray<Float> {
```

vil02 (approved):

govardhanshah456:
Yeah, thanks.

siriak (approved):
Looks good, thanks!

vil02:
@govardhanshah456 this was a good piece of Java programing. Thanks!

PR #4408, opened 2023-09-26
Create FourSumProblem.java

mouss3abbb:


BamaCharanChhandogi:
Hi, please write only one approach with good time complexity. You don't need to write all approaches. Also, remove the main method and add test cases

mouss3abbb:
Feel free to check the last commits

BamaCharanChhandogi:
@vil02 let's take this 1 PR for four sum problem. 
could you review this one?

src/main/java/com/thealgorithms/misc/FourSumProblem.java
```diff
@@ -0,0 +1,44 @@
...
+                idx.add(i);
+                idx.add(j);
+                map.put(arr.get(i) + arr.get(j), idx);
+            }
+        }
+
+        for (int i = 0; i < arr.size(); i++) {
+            for (int j = i + 1; j < arr.size(); j++) {
+                int t = target - (arr.get(i) + arr.get(j));
+                if (map.containsKey(t) && map.get(t).get(0) != j && map.get(t).get(1) != j && map.get(t).get(0) != i && map.get(t).get(1) != i) {
```
vil02, line 31: This condition... again let's somehow explain what happens here by moving it into a separate method. Current tests do not _cover_ all branches.

src/test/java/com/thealgorithms/misc/FourSumProblemTest.java
```diff
@@ -0,0 +1,59 @@
+package com.thealgorithms.misc;
+
+import static org.junit.jupiter.api.Assertions.*;
```
vil02, line 3:
```suggestion
import static org.junit.jupiter.api.Assertions.assertEquals;
import static org.junit.jupiter.api.Assertions.assertTrue;
```

vil02 (changes requested):
Great that you have added tests! The code needs some additional structure (which will provide explanation). It looks very promising!

mouss3abbb:
@vil02 Thanks for your changes! Hopefully I completed all of them. Let me know if there is any issue

src/main/java/com/thealgorithms/misc/FourSumProblem.java
```diff
@@ -0,0 +1,59 @@
...
+    private static void findAllPairSum(ArrayList<Integer> arr, HashMap<Integer, List<Integer>> map) {
+        for (int i = 0; i < arr.size(); i++) {
+            for (int j = i + 1; j < arr.size(); j++) {
+                List<Integer> idx = new ArrayList<>();
+                idx.add(i);
+                idx.add(j);
+                map.put(arr.get(i) + arr.get(j), idx);
+            }
+        }
+    }
```
vil02, line 58: I do not like the fact that the methods `findAllPairSum` and `checkRemainingSumAvailability` modify their inputs (`map` and `set`). What do you think about the following:
```suggestion
    public static List<List<Integer>> hashing(final ArrayList<Integer> arr, final int target) {
        Set<List<Integer>> set = new HashSet<>();
        // Find the sum of each pair of elements and store it in the map
        var map = findAllPairSum(arr);
        // We iterate over each pair of elements and check if their sum added to any value in the map will equal the target
        for (int i = 0; i < arr.size(); i++) {
            for (int j = i + 1; j < arr.size(); j++) {
                final var subRes = checkRemainingSumAvailability(arr, target, i, j, map);
                if (subRes.size() > 0) {
                    set.add(subRes);
                }
            }
        }
        return new ArrayList<>(set);
    }

    private static List<Integer> checkRemainingSumAvailability(final ArrayList<Integer> arr, final int target, final int i, final int j, final HashMap<Integer, List<Integer>> map) {
        final int remainingSum = target - (arr.get(i) + arr.get(j));
        List<Integer> sum = new ArrayList<>();
        // We need to check is we have taken any of these indices before to avoid taking the same index twice.
        if (isThisIndexTaken(i, j, map, remainingSum)) {
            sum.add(arr.get(i));
            sum.add(arr.get(j));
            sum.add(arr.get(map.get(remainingSum).get(0)));
            sum.add(arr.get(map.get(remainingSum).get(1)));
            Collections.sort(sum);
        }
        return sum;
    }

    private static boolean isThisIndexTaken(final int i, final int j, final HashMap<Integer, List<Integer>> map, final int remainingSum) {
        return map.containsKey(remainingSum) && map.get(remainingSum).get(0) != j && map.get(remainingSum).get(1) != j && map.get(remainingSum).get(0) != i && map.get(remainingSum).get(1) != i;
    }

    private static HashMap<Integer, List<Integer>> findAllPairSum(final ArrayList<Integer> arr) {
        HashMap<Integer, List<Integer>> map = new HashMap<>();
        for (int i = 0; i < arr.size(); i++) {
            for (int j = i + 1; j < arr.size(); j++) {
                List<Integer> idx = new ArrayList<>();
                idx.add(i);
                idx.add(j);
                map.put(arr.get(i) + arr.get(j), idx);
            }
        }
        return map;
    }
```

src/main/java/com/thealgorithms/misc/FourSumProblem.java
```diff
@@ -0,0 +1,59 @@
...
+ * @author mouss3abbb
+ */
+public final class FourSumProblem {
+
+    /*
+    In the hashing solution, we maintain a hashmap of the sum of each pair of elements in the array.
+    We then iterate over each pair of elements in the array and check if the required sum to be equal to target is found in the map previously.
+    The map now acts as a frequency table. However, we also need to check that the indices of all four elements are distinct, that's why we maintain a pair of indices in the map itself and compare it with the pair of elements we are at right now.
+    */
+    public static List<List<Integer>> hashing(final ArrayList<Integer> arr, final int target) {
```
vil02, line 17: `hashing` is not really the best name. Maybe `solve` or `find` would be better?

Also the return type: I think it should return something like a set of (sorted?) tuples (of length 4). It will definitely make testing easier.

src/main/java/com/thealgorithms/misc/FourSumProblem.java
```diff
@@ -0,0 +1,59 @@
+package com.thealgorithms.misc;
+
+import java.util.*;
+
+/**
+ * Solution for The Four Sum Problem.
+ *
+ * @author mouss3abbb
+ */
+public final class FourSumProblem {
```
vil02, line 10: It does not make any sense to have objects of this class (because it has only `static` methods).

```suggestion
public final class FourSumProblem {
    private FourSumProblem() {
    }
```

src/test/java/com/thealgorithms/misc/FourSumProblemTest.java
```diff
@@ -0,0 +1,51 @@
...
+public class FourSumProblemTest {
+
+    @Test
+    void testFourSumExists() {
+        final var values = new ArrayList<>(List.of(-2, 5, 0, 6));
+        Collections.sort(values);
+        final int target = 9;
+        final var result = FourSumProblem.hashing(values, target).get(0);
+        Collections.sort(result);
+        assertEquals(values, result);
```
vil02, line 26: This test could still pass if `FourSumProblem.hashing(...)` would return more than one solution. Please rewrite the tests checking if the actual result equals the expected result (i.e. single `assertEquals`).

Ideally, please rewrite these tests as parameterized tests. Good example to look at: [`FindMaxRecursionTest`](https://github.com/TheAlgorithms/Java/blob/master/src/test/java/com/thealgorithms/maths/FindMaxRecursionTest.java).

This is a complex problem: please add more tests.

vil02 (changes requested):

siddhant2002:
Can I work under this issue?

siriak:
@mouss3abbb are you working on it or @siddhant2002 can pick this up?

mouss3abbb:
> @mouss3abbb are you working on it or @siddhant2002 can pick this up?

I am working on it

PR #4415, opened 2023-09-27
Add SingleBitOperations

lukasb1b:
#4350 as its not done yet


src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperators.java
```diff
@@ -0,0 +1,32 @@
+package com.thealgorithms.bitmanipulation;
+
+/*
+ * Author: lukasb1b (https://github.com/lukasb1b)
+ */
+
+public class SingleBitOperators {
```
vil02, line 7: Let's make this class a _proper utility class_.

```suggestion
public final class SingleBitOperators {
    private SingleBitOperators() {
    }
```

src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperators.java
```diff
@@ -0,0 +1,32 @@
...
+
+/*
+ * Author: lukasb1b (https://github.com/lukasb1b)
+ */
+
+public class SingleBitOperators {
+    /**
+     * Flip the bit at position 'bit' in 'num'
+     */
+    public static int flipBit(int num, int bit) {
```
vil02, line 11:
```suggestion
    public static int flipBit(final int num, final int bit) {

```

src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperators.java
```diff
@@ -0,0 +1,32 @@
...
+        return num ^ (1 << bit);
+    }
+    /**
+     * Set the bit at position 'bit' to 1 in the 'num' variable
+     */
+    public static int setBit(int num, int bit) {
```
vil02, line 17:
```suggestion
    public static int setBit(final int num, final int bit) {

```

src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperators.java
```diff
@@ -0,0 +1,32 @@
...
+        return num | (1 << bit);
+    }
+    /**
+     * Clears the bit located at 'bit' from 'num'
+     */
+    public static int clearBit(int num, int bit) {
```
vil02, line 23:
```suggestion
    public static int clearBit(final int num, final int bit) {

```

src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperators.java
```diff
@@ -0,0 +1,32 @@
...
+        return num & ~(1 << bit);
+    }
+    /**
+     * Get the bit located at 'bit' from 'num'
+     */
+    public static int getBit(int num, int bit) {
```
vil02, line 29:
```suggestion
    public static int getBit(final int num, final int bit) {

```

src/test/java/com/thealgorithms/bitmanipulation/SingleBitOperatorTest.java
```diff
@@ -0,0 +1,32 @@
+package com.thealgorithms.bitmanipulation;
+
+import static org.junit.jupiter.api.Assertions.*;
```
vil02, line 3:
```suggestion
import static org.junit.jupiter.api.Assertions.assertEquals;

```

vil02 (changes requested):
I would suggest to rename this class to `SingleBitOperations`.

src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperator.java
```diff
@@ -0,0 +1,32 @@
+package com.thealgorithms.bitmanipulation;
+
+/*
+ * Author: lukasb1b (https://github.com/lukasb1b)
+ */
+
+public final class SingleBitOperator {
```
vil02, line 7: As suggested [earlier](https://github.com/TheAlgorithms/Java/pull/4415#discussion_r1338339851):
```suggestion
public final class SingleBitOperator {
    private SingleBitOperator() {
    }
```

vil02 (changes requested):
Why do you use the name `SingleBitOperator` and not `SingleBitOperations`?

I think the second name makes more sense in this context:
- this class defines _single bit operations_,
- it is not a _single bit operator_.

Please rebase (cf. #4422).

lukasb1b (commented):
Why does the programm need this?

vil02:
> Why does the programm need this private constructor?

Your class has only `static` methods. Therefore it does not really make sense to have objects of this class. Marking the default constructor as `private` (and not having other ones) makes it impossible to create objects of that class. Without this, one could still write something like `var obj = SingleBitOperator();`, but as said earlier, it does not really make sense.

Also JaCoCo will show now _100%_ coverage for your file.

lukasb1b:
Ok that does make sense

vil02:
@lukasb1b please have a look at [this comment](https://github.com/TheAlgorithms/Java/pull/4415#pullrequestreview-1649293416) about renaming.

lukasb1b:
@vil02 changed the naming

vil02 (approved):

PR #4470, opened 2023-10-01
Update Pangram.java using Java Collection Framework

Ankit-Saha08:
An upgraded approach using Java Collections which is simple to understand and has reduced number of lines of code

### DIFF
--- a/src/main/java/com/thealgorithms/strings/Pangram.java
+++ b/src/main/java/com/thealgorithms/strings/Pangram.java
@@ -25,7 +25,7 @@ public class Pangram {
     public static boolean isPangram(String s) {
         // an approach using Java Collection Framework
         Set<Character> alpha = new HashSet<Character>();
-        for (int i=0; i<s.length(); i++)
+        for (int i = 0; i < s.length(); i++)
             if (s.charAt(i)!=' ') alpha.add(s.charAt(i));
         if (alpha.size() == 26) return true;
         return false;
--- a/src/main/java/com/thealgorithms/strings/Pangram.java
+++ b/src/main/java/com/thealgorithms/strings/Pangram.java
@@ -1,5 +1,6 @@
 package com.thealgorithms.strings;
 
+import java.util.Collections;
 /**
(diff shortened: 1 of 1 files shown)


debasishbsws:
Check the code build is failing. and also have some linting errors

Ankit-Saha08:
Yeah I am trying to figure out the same about the build failing.

src/main/java/com/thealgorithms/strings/Pangram.java
```diff
@@ -22,19 +23,12 @@ public static void main(String[] args) {
...
-                return false;
-            }
-        }
-        return true;
+        // an approach using Java Collection Framework
+        Set<Character> alpha = new HashSet<Character>();
+        for (int i=0; i<s.length(); i++)
+            if (s.charAt(i)!=' ') alpha.add(s.charAt(i));
+        if (alpha.size() == 26) return true;
+        return false;
```
debasishbsws: This approach will not work as 'a' and "A" is different.
for example 'The quick brown fox jumps over the lazy dog' will return false as it will count T and t twice and the size will be greater than 26.
It is better if you change the char to lower case and apply this.

And please leave the previous algorithm there as it is more simple and clear and make another function with this alternate approach.

debasishbsws (changes requested):

src/main/java/com/thealgorithms/strings/Pangram.java
Ankit-Saha08: Well I get that. But certainly if the string is converted to Lower Case and this approach is taken into count, it should be working fine. Isn't it?

Ankit-Saha08: Also I will update this Java Collections approach in another function separately.

Ankit-Saha08: I have committed the fresh changes in another branch and not in this one

debasishbsws:
We do not need thi. already done by other PR

PR #4550, opened 2023-10-02
Fixed: corrected a  wrong code and also added a new code for prevSmallerElement

piyush-gpt:
The code inside the file "NextSmallerElement.java" inside stacks folder is wrong. The code which is currently written in it is for finding the previous smaller element that is smaller element at left instead of smaller element at right which is wrong as we have to find the next smaller element , so I corrected the code to actually find the next smaller element.

Also I added a correct code to find the previous smaller element of the current element in the array that is  smaller element on the left of the current element.


piyush-gpt:
@BamaCharanChhandogi  please check

PR #4571, opened 2023-10-02
Add `MatrixRank`

### LINKED ISSUE #4570: [FEATURE REQUEST] Addition of Rank Of Matrix Algorithm

New Algorithm under
https://github.com/TheAlgorithms/Java/tree/master/src/main/java/com/thealgorithms/maths

File Name Proposed: MatrixRank.java

What?
An implementation of Rank of a Matrix.

### Issue details

More about rank of a matrix: https://en.wikipedia.org/wiki/Rank_(linear_algebra)

Algorithm complexity:
1. Time Complexity: O(M*N*rank)
2. Space Complexity: O(N)

### Additional Information

**Note: This issue is self assigned and PR is raised as a part of hacktoberfest**

anupomkar:

closes #4570


anupomkar:
@yanglbme @debasishbsws @vil02 @BamaCharanChhandogi can you please review this PR. And also please mark this as `hacktoberfeast` and `hackstoberfeast-accepted` while merging

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,62 @@
...
+ * </p>
+ *
+ * <p>
+ * The rank of a matrix can also be defined as the largest order of any non-zero minor in the matrix.
+ * </p>
+ *
+ * <p> Read more about rank of matrix <a href="https://en.wikipedia.org/wiki/Rank_(linear_algebra)">here</a></p>
+ *
+ * @author Anup Omkar
+ */
```
vil02, line 20: Please document your class (and its methods) using _doxygen_ tags.

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,62 @@
+package com.thealgorithms.maths;
+
+/**
+ *
+ *
+ * <h1>Rank of Matrix</h1>
+ *
+ * <p>
+ * The rank of a matrix is the largest number of linearly independent rows/columns of the matrix.
+ * The rank is not only defined for square matrices.
```
vil02, line 10: This sentence is very confusing. And it does not provide much information.

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,62 @@
...
+ * </p>
+ *
+ * <p> Read more about rank of matrix <a href="https://en.wikipedia.org/wiki/Rank_(linear_algebra)">here</a></p>
+ *
+ * @author Anup Omkar
+ */
+public class MatrixRank {
+
+    // Small value to handle precision issues with floating point numbers
+    private final static double EPSILON = 1e-10;
```
vil02, line 24: I would suggest to introduce a method of the type: `bool isZero(final double x)`.

In general I am not sure how to handle this properly. If you will have some suggestions please let me know.

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Returns rank of matrix
```
vil02, line 27:
```suggestion
     * @brief computes the rank of the input matrix
```

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @param matrix matrix whose rank is to be found
```
vil02, line 29:
```suggestion
     * @param matrix the input matrix
```

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,62 @@
...
+     * @return rank of matrix
```
vil02, line 30:
```suggestion
     * @return rank of the input matrix
```

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public static int computeRank(double[][] matrix) {
+        int numRows = matrix.length;
+        int numColumns = matrix[0].length;
```
vil02, line 34: What if someone would call your function with something like `new double[][] {{1, 2}, {10}, {100, 200, 300}}`?

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,62 @@
...
+                }
+                for (int otherRow = 0; otherRow < numRows; ++otherRow) {
+                    if (otherRow != pivotRow && Math.abs(matrix[otherRow][colIndex]) > EPSILON) {
+                        for (int col2 = colIndex + 1; col2 < numColumns; ++col2) {
+                            matrix[otherRow][col2] -= matrix[pivotRow][col2] * matrix[otherRow][colIndex];
+                        }
+                    }
+                }
+            }
+        }
```
vil02, line 59: This code is very complex. Do you see some change to cut it into smaller pieces? The first step it to try to write comments what happens in some of the blocks.

src/test/java/com/thealgorithms/maths/MatrixRankTest.java
```diff
@@ -0,0 +1,36 @@
...
+        assertEquals(2, MatrixRank.computeRank(square3x3Matrix));
+
+        double[][] square4x4Matrix = {{2, 3, 0, 1}, {1, 0, 1, 2}, {-1, 1, 1, -2}, {1, 5, 3, -1}};
+        assertEquals(3, MatrixRank.computeRank(square4x4Matrix));
+
+        double[][] rectangular2X3Matrix = {{1, 2, 3}, {3, 6, 9}};
+        assertEquals(1, MatrixRank.computeRank(rectangular2X3Matrix));
+
+        double[][] rectangular3X4Matrix = {{0.25, 0.5, 0.75, 2}, {1.5, 3, 4.5, 6}, {1, 2, 3, 4}};
+        assertEquals(2, MatrixRank.computeRank(rectangular3X4Matrix));
```
vil02, line 34: Try to express your tests as _parameterized test_, cf. [`FindMaxRecursionTest.java`](https://github.com/TheAlgorithms/Java/blob/master/src/test/java/com/thealgorithms/maths/FindMaxRecursionTest.java).

vil02 (changes requested):

src/main/java/com/thealgorithms/maths/MatrixRank.java
vil02: I think the simplest way would be to add a new argument `double epsilon = 10e-10` to `computeRank` method. So also the method `isZero` should take an additional argument.

anupomkar: Also why do we need this isZero() method, isn't it overkilling for just one check?

src/main/java/com/thealgorithms/maths/MatrixRank.java
anupomkar: changed

src/main/java/com/thealgorithms/maths/MatrixRank.java
vil02: You are right. For time being, just add a `isZero` method with hard coded `epsilon`.

src/test/java/com/thealgorithms/maths/MatrixRankTest.java
anupomkar: Changed

src/main/java/com/thealgorithms/maths/MatrixRank.java
anupomkar: Added condition

src/main/java/com/thealgorithms/maths/MatrixRank.java
anupomkar: Done. Please review the lastet changes.

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,125 @@
...
+     */
+    public static int computeRank(double[][] matrix) {
+        int numRows = matrix.length;
+        int numColumns = matrix[0].length;
+        int rank = 0;
+
+        boolean[] rowMarked = new boolean[numRows];
+        if (isJaggedMatrix(matrix)) {
+            throw new IllegalArgumentException("The input matrix is a jagged matrix");
+        }
```
vil02, line 34: Please `throw` inside some method like `checkInput` calling `isJaggedMatrix`.

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,125 @@
...
+ * 1 2 3
+ * 2 4 6
+ * 3 6 9
+ * Despite having 3 rows and 3 columns, this matrix only has a rank of 1 because all rows (and columns) are multiples of each other.
+ * It's a fundamental concept that gives key insights into the structure of the matrix.
+ * It's important to note that the rank is not only defined for square matrices but for any m x n matrix.
+ *
+ * @author Anup Omkar
+ */
+public class MatrixRank {
```
vil02, line 16:
```suggestion
public final class MatrixRank {
    private MatrixRank() {
    }
```

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private final static double EPSILON = 1e-10;
+
+    /**
+     * @brief Computes the rank of the input matrix
+     *
+     * @param matrix The input matrix
+     * @return The rank of the input matrix
+     */
+    public static int computeRank(double[][] matrix) {
```
vil02, line 26: This method modifies the input `matrix`. It has to be changed.

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,125 @@
...
+ * 3 6 9
+ * Despite having 3 rows and 3 columns, this matrix only has a rank of 1 because all rows (and columns) are multiples of each other.
+ * It's a fundamental concept that gives key insights into the structure of the matrix.
+ * It's important to note that the rank is not only defined for square matrices but for any m x n matrix.
+ *
+ * @author Anup Omkar
+ */
+public class MatrixRank {
+
+    private final static double EPSILON = 1e-10;
```
vil02, line 18: @siriak what is a proper way of handling that? Having an abstract class with abstract method `isZero` is an overkill in my opinion. Sending an additional `epsilon` argument to `computeRank` is cumbersome. Or am I just exaggerating and it should be left as it is?

vil02 (changes requested):

src/main/java/com/thealgorithms/maths/MatrixRank.java
anupomkar: Changed. Passing a copy of a matrix.

src/main/java/com/thealgorithms/maths/MatrixRank.java
siriak: Let's leave it as-is for now. We can move it to a base class if we ever get one for some other reason

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,154 @@
...
+ * 1 2 3
+ * 2 4 6
+ * 3 6 9
+ * Despite having 3 rows and 3 columns, this matrix only has a rank of 1 because all rows (and columns) are multiples of each other.
+ * It's a fundamental concept that gives key insights into the structure of the matrix.
+ * It's important to note that the rank is not only defined for square matrices but for any m x n matrix.
+ *
+ * @author Anup Omkar
+ */
+public final class MatrixRank {
```
vil02, line 16:
```suggestion
public final class MatrixRank {
    private MatrixRank() {
    }
```

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,154 @@
...
+
+    private final static double EPSILON = 1e-10;
```
vil02, line 18:
```suggestion
    private static final double EPSILON = 1e-10;
```

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,154 @@
...
+        int numRows = matrix.length;
+        int numColumns = matrix[0].length;
+        int rank = 0;
+
+        boolean[] rowMarked = new boolean[numRows];
+
+        double[][] matrixCopy = new double[numRows][numColumns];
+        for (int rowIndex = 0; rowIndex < numRows; ++rowIndex) {
+            System.arraycopy(matrix[rowIndex], 0, matrixCopy[rowIndex], 0, numColumns);
+        }
```
vil02, line 38: Please move it into a separate method, like `deepCopy`.

vil02 (changes requested):
Almost there.

src/main/java/com/thealgorithms/maths/MatrixRank.java
anupomkar: Changed it.

vil02: @anupomkar could you please add this test checking that the input is not modified? In general, it is a good practice go _express requirements as tests_.

vil02:
> Also please mark this as hacktoberfeast and hackstoberfeast-accepted

This is really not needed. This repo has the `hacktoberfest` tag, and the [website](https://hacktoberfest.com/) recognizes that.

src/main/java/com/thealgorithms/maths/MatrixRank.java
anupomkar: Is that really needed?
Firstly, all the helper functions like isZero, isJaggred, checkInput and deepCopy should not be part of this algorithm/class. They all should be moved to some common place or helper class where all these helper functionalities must reside and tested. The `MatrixRankTest.java` class should not have responsibility of testing these things. It should only test main business logic / algorithm logic.

src/main/java/com/thealgorithms/maths/MatrixRank.java
```diff
@@ -0,0 +1,164 @@
...
+            if (row == null || row.length == 0) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    /**
+     * @brief Checks if the input matrix is a jagged matrix.
+     * Jaggged matrix is a matrix where the number of columns in each row is not the same.
```
siriak, line 92:
```suggestion
     * Jagged matrix is a matrix where the number of columns in each row is not the same.
```

siriak (approved):

src/main/java/com/thealgorithms/maths/MatrixRank.java
vil02: In my opinion: it is really needed.

Your `MatrixRank.computeRank` seems to be a reasonable _unit_ and it is reasonable to require that it will:
- return correct result,
- throw an error for wrong input,
- not modify the input.

So these aspects should be tested.

Should the mentioned methods be part of this class or different? Depends: if there are other places in this repo that exactly the same functions are needed, then yes - but this if out of scope for this pull-request. I do not see a relation between their location and the tests of your class. Even if the method `isJaggred` would be defined in another class and testes separately, you would still need to check if `MatrixRank.computeRank` throws for an input not being a proper matrix.

anupomkar: Added unit tests. Please check

siriak (approved):
@vil02 please check

vil02 (approved):

PR #4605, opened 2023-10-03
Added NthRoot.java

amul67142:
Problem Statement: we have 2 integers 'n' and 'm'. we have to return the 'nth' root of 'm'.
   If the nth root is not an integer, return -1.

   solution : We will find the nth root of a given integer m using binary search.

### DIFF
--- a/src/main/java/com/thealgorithms/searches/NthRoot.java
+++ b/src/main/java/com/thealgorithms/searches/NthRoot.java
@@ -1,0 +1,44 @@
+package com.thealgorithms.searches;
+
+/* Problem Statement: we have 2 integers 'n' and 'm'. we have to return the 'nth' root of 'm'.
+   If the nth root is not an integer, return -1.
+
+   solution : We will find the nth root of a given integer m using binary search.
+
+  example1: Input: 'n' = 3, 'm' = 27
+
+            Output: 3
+
+            3rd Root of 27 is 3, as (3)^3 equals 27.
+
+  example2: Input: 'n' = 4, 'm' = 69
+
+            Output: -1
+
+            There is no 4th root of 69 present.
+
+  @author - Amul Sharma (https://github.com/amul67142)
+*/
+
+public final class NthRoot {
+    private NthRoot() {
+    }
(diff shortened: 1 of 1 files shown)


src/main/java/com/thealgorithms/searches/NthRoot.java
```diff
@@ -0,0 +1,43 @@
...
+
+            3rd Root of 27 is 3, as (3)^3 equals 27.
+
+  example2: Input: 'n' = 4, 'm' = 69
+
+            Output: -1
+
+            There is no 4th root of 69 present.
+
+  @author - Amul Sharma (https://github.com/amul67142)
```
Manan-09, line 20: @amul67142  We can also do like below, this will give you nth root of number m
`Math.pow(m, 1.0/n)`

In case it is is decimal we can return it as is or we can do ceiling to nearest integer or return -1

I don't think binary search is required here.

Please tell if above approach has some disvantages.

amul67142: If we require only integer roots and performance is a concern for large integers, my code might be more appropriate. If we prioritize simplicity and code clarity, the Math.pow approach can be a good choice.

amul67142: In Math.pow(x, 1.0/n) if x is negative, it will still calculate the root, but the result will be a complex number.

Manan-09: How does your code handle negative number as 'x', from code I can observe it will directly return -1 for all the inputs

amul67142: No i am not saying my code handle negative numbers, i am just talking here about the error handling which is returning -1. Main advantage is the first point i have mentioned.
Also if i change my code for negative integers it just give a same answer close to the ans which math.pow() gives.

amul67142: @debasishbsws

Issue #4617, opened 2023-10-04
[BUG] <Issue in Armstrong Number>


satyabarghav:
### Description

According to the definition: **It is a number that is equal to the sum of its own digits, each raised to the power of the number of digits.** 
But with the given implementation in the repo the digit is not raised to the power of number of Digits but instead raised only to 3.

It can be explained with the help of the Example: 

1634 is an Armstrong Number as => 1^4 + 6^4 + 3^4 + 4^4 = 1634
but according to the implementation we have it will evaluated as 1^3 + 6^3 + 3^3 + 4^3 = 308 != 1634 and will not be an Armstrong Number. The code works fine for the 3 digit numbers but isn't valid for numbers above that threshold

### Generalization
In other words, if you take an n-digit number and raise each of its digits to the nth power, and then sum these values, if the result is equal to the original number, it is considered an Armstrong number.

### Excepted behavior

So we need to change the logic of the code to make the 3 in the Power function variable to the given input.

satyabarghav:
I request the maintainer(s) to label this with "hacktoberfest" label.

VishnuAmit:
Hey @satyabarghav . I would like to work on this issue.  Please assign it to me under Hacktoberfest 23

Tej-as1:
I would like to contribute to this issue. I have a deep knowledge of DSA and CP topics so I think I can contribute to this and more such issues.

Sayandas8466:
I think i will solve this issue

NamanGoyal17:
hy @satyabarghav . can you please label your all issues to hacktoberfest2023 . it will help us to solve your all issues more passionately.

faizan0702:
I have resolved the bug Please review : https://github.com/TheAlgorithms/Java/pull/4625

PR #4625, opened 2023-10-04
#4617 Issue resolved for armstrong Number

faizan0702:
- Resolved bug for armstrong number now it will handle for all numbers.
- Added new test cases for armstrong number.


BamaCharanChhandogi:
thanks, but Its a duplicate PR.

PR #4686, opened 2023-10-06
Herons : Changed the signature of the function

satyabarghav:
- [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [x] This pull request is all my own work -- I have not plagiarized it.
- [x] All filenames are in PascalCase.
- [x] All functions and variable names follow Java naming conventions.
- [x] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.

I have already raised this issue [here](https://github.com/TheAlgorithms/Java/issues/4685#issue-1929808965) 
The required changes were made to make it work with even larger set of data types like double and float and further documentation was also added ro ease the readability and understandability of the code.

PR #4687, opened 2023-10-06
changed the input parameters to double and added triangle validation …

### LINKED ISSUE #4685: [BUG] <Issue in Herons Formula>
### Description
In the implementation of Herons Formula we have the signature of the function as Herons(int a, int b, int c) so when we call the function we are confined to only integer inputs and we will get the error when using sides of double or float data type

### Excepted behavior

So we need to change the signature of the function to even work with data types of even double or float.

I have attached the error we get while calling the function with parameters of double data type.

### Screenshots

image

kushbatla:
Fixes #4685 
-Changes:
1.herons formula parameters changed to double to calculate area
2.Added triangle validation condition 


vil02:
This is a duplicate of #4686.

PR #4701, opened 2023-10-06
 Program Without Using Buildin Power Function

madhavagarwal3012:
- [x ] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [x ] This pull request is all my own work -- I have not plagiarized it.
- [ x] All filenames are in PascalCase.
- [ x] All functions and variable names follow Java naming conventions.
- [ x] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.

### DIFF
--- a/src/main/java/com/thealgorithms/maths/Armstrong.java
+++ b/src/main/java/com/thealgorithms/maths/Armstrong.java
@@ -9,7 +9,7 @@ package com.thealgorithms.maths;
  * 1634 is an Armstrong number because 1^4 + 6^4 + 3^4 + 4^4 = 1634.
  * An Armstrong number is often called a Narcissistic number.
  *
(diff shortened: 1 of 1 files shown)


lukasb1b:
Why would your change it like that your implementation needs more space and time complexity?

madhavagarwal3012:
> Why would your change it like that your implementation needs more space and time complexity?

its just a method assuming that we dont have any build in function so without using build in function how will you work on this problem, it shows this.

lukasb1b (changes requested):
Then just use the original code and t´replace the line sum += Math.pow(digit, power); and dont change the whole file you can repace it with one variable and a loop.

PR #4770, opened 2023-10-11
Bit swap

lukasb1b:
Swaping two bits of a number


src/main/java/com/thealgorithms/bitmanipulation/BitSwap.java
```diff
@@ -0,0 +1,19 @@
+package com.thealgorithms.bitmanipulation;
+
+public final class BitSwap {
+    private BitSwap() {
+    }
+    /*
+     * Swaps the bits at the position swap1 and swap2 from bit
+     */
+    public static int bitSwap(int bit, int swap1, int swap2) {
```
vil02, line 9:
```suggestion
    public static int bitSwap(int data, final int posA, final int posB) {
```

src/main/java/com/thealgorithms/bitmanipulation/BitSwap.java
```diff
@@ -0,0 +1,19 @@
...
+        int val1 = (bit >> swap1) & 1;
+        int val2 = (bit >> swap2) & 1;
```
vil02, line 11: What do you think about applying [`SingleBitOperations.getBit`](https://github.com/TheAlgorithms/Java/blob/1dc64b16859d95446a8bfe1c50692ce57316bc42/src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperations.java#L31) here?

src/test/java/com/thealgorithms/bitmanipulation/BitSwapTest.java
```diff
@@ -0,0 +1,12 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.api.Test;
+public class BitSwapTest {
+    @Test
+    void testHighestSetBit() {
+        assertEquals(3, BitSwap.bitSwap(3, 0, 1));
+        assertEquals(5, BitSwap.bitSwap(6, 0, 1));
+    }
```
vil02, line 11: I would suggest to add a test case with equal positions.

```suggestion
        assertEquals(5, BitSwap.bitSwap(6, 0, 1));
        assertEquals(7, BitSwap.bitSwap(7, 1, 1));
    }
```

vil02 (changes requested):

src/main/java/com/thealgorithms/bitmanipulation/BitSwap.java
lukasb1b: I tought about it too, do you think it would be better to not use functions so it is easier to understand? If you think it is better we can do it.

vil02: In my opinion it is easier to understand if `getBit` is used. Just have a look: `int bitAtPosA = (data >> posA) & 1` vs. `int bitAtPosA = SingleBitOperations.getBit(data, posA)`. The part `(data >> posA) & 1` is a bit _hacky_ and maybe needs a comment like _we extract the bit at position A_. But why not to simply use the method called `getBit`?

Moreover: `getBit` is tested. The line `(data >> posA) & 1` not.

Another advantage is that when you use `getBit` you can eliminate the variables and just have later the line `if (SingleBitOperations.getBit(data, posA) != SingleBitOperations.getBit(data, posB))`.

lukasb1b: I updated it.

src/test/java/com/thealgorithms/bitmanipulation/BitSwapTest.java
```diff
@@ -0,0 +1,13 @@
+package com.thealgorithms.bitmanipulation;
+
+import static org.junit.jupiter.api.Assertions.*;
```
vil02, line 3:
```suggestion
import static org.junit.jupiter.api.Assertions.assertEquals;
```

src/main/java/com/thealgorithms/bitmanipulation/BitSwap.java
```diff
@@ -0,0 +1,17 @@
+package com.thealgorithms.bitmanipulation;
+
+public final class BitSwap {
+    private BitSwap() {
+    }
+    /*
+     * Swaps the bits at the position swap1 and swap2 from bit
```
vil02, line 7:
```suggestion
     * @brief Swaps the bits at the position posA and posB from data
```

src/main/java/com/thealgorithms/bitmanipulation/BitSwap.java
```diff
@@ -0,0 +1,17 @@
...
+     */
+    public static int bitSwap(int data, final int posA, final int posB) {
+
+        if (SingleBitOperations.getBit(data, posA) != SingleBitOperations.getBit(data, posB)) {
+            data = data ^ (1 << posA);
+            data = data ^ (1 << posB);
```
vil02, line 13: What do you think about that?

```suggestion
            data ^= (1 << posA) ^ (1 << posB);
```

vil02 (changes requested):

src/main/java/com/thealgorithms/bitmanipulation/BitSwap.java
```diff
@@ -0,0 +1,16 @@
+package com.thealgorithms.bitmanipulation;
+
+public final class BitSwap {
+    private BitSwap() {
+    }
+    /*
+     * @brief Swaps the bits at the position posA and posB from data
+     */
+    public static int bitSwap(int data, final int posA, final int posB) {
+
```
vil02, line 10:
```suggestion
```

vil02 (approved):

PR #4867, opened 2023-10-21
A algorithm -added algorithm for A* using java 

pawaspy:


debasishbsws (changes requested):
AiThroughJava is not a good package name, use `artificialintelligence` instead.

And Trapezoidal-Simpson is a math problem as far as I know, so try to put it inside the [maths](https://github.com/TheAlgorithms/Java/tree/master/src/main/java/com/thealgorithms/maths) package.

**PUT Junit Test for all of them**

Issue #4896, opened 2023-10-23
[FEATURE REQUEST] <Calculate the square root of x using the bisection method>


yuluo-yx:
### What would you like to Propose?

Calculate the square root of x using the bisection method

### Issue details

Example 1:
- Input: x = 4
- Output: 2

Example 2:
- Input: x = 8
- Output: 2
- Explanation: The arithmetic square root of 8 is 2.82842..., since the return type is an integer, the decimal part will be rounded off.

chauhansumitdev:
Can I get this assigned to me?

PR #4927, opened 2023-10-25
Calculate the square root of x using the bisection method #4896

Vijay-948:


BamaCharanChhandogi:
please don't raise duplicate PR.

PR #4955, opened 2023-11-02
Create jump _greedy-problem.py 

### LINKED ISSUE #4692: [FEATURE REQUEST] <JUMP Game Greedy problem>

You are given an integer array nums. You are initially positioned at the array's first index, and each element in the array represents your maximum jump length at that position.

Return true if you can reach the last index, or false otherwise.

### Issue details

You are given an integer array nums. You are initially positioned at the array's first index, and each element in the array represents your maximum jump length at that position.

Return true if you can reach the last index, or false otherwise.

Input: nums = [2,3,1,1,4]
Output: true
Explanation: Jump 1 step from index 0 to 1, then 3 steps to the last index.

### Additional Information

I want to contribute on this issue for Hacktoberfest

deepak4566:
fixes #4692


vil02:
This is python code.

PR #4956, opened 2023-11-04
changed to randomly assigned value

why20w18:
Integer[] arr = new Integer[9]; // = {4, 23, 6, 78, 1, 54, 231, 9, 12};
 //Random arr create
for(int i = 0 ; i < arr.length ; i++) arr[i] = (int) (Math.random()*231);


vil02:
@why20w18 thanks for your contribution. I see your point with _testing_ the implementation with random values. However, the `SelectionSort` is tested _good enough_ at the moment (please have in mind that its implementation is simple). I would suggest to completely remove the `main` method here. Could you please create another PR removing the `main` method and close this one?

Issue #4506, opened 2023-10-01
[FEATURE REQUEST] <title>Coin Tower Problem Using Dynamic Programming


DineshK3012:
### What would you like to Propose?

I want to add Coin Tower Problem to the Dynamic Programming folder

### Issue details

Coin Tower Problem:

A and B are playing a new game today. They form a tower of N coins and make a move in alternate turns. B plays first. In one step, the player can remove either 1, X, or Y coins from the tower. The person to make the last move wins the game. Can you find out who wins the game?

DineshK3012:
Added coin tower problem code : https://github.com/TheAlgorithms/Java/pull/4547

SarthakChaudhary46:
@vil02 You mentioned having a few ideas for contribution. Please do let me know, brother—I'm all ears. Also, Do you no longer require a solution for this issue, or was my attempt not satisfactory?

PR #5023, opened 2024-01-22
Added Coin Tower Problem Using Dynamic Programming, ISSUE #4506

SarthakChaudhary46:
Added Coin Tower Problem Using Dynamic Programming, ISSUE #4506

### DIFF
--- a/src/main/java/com/thealgorithms/dynamicprogramming/CoinTowerProblem.java
+++ b/src/main/java/com/thealgorithms/dynamicprogramming/CoinTowerProblem.java
@@ -1,30 +1,4 @@
-//Code By @SarthakChaudhary46
-
-import java.util.Scanner;
-import org.junit.Test;
-import static org.junit.Assert.*;
-
 public class CoinTowerProblem {
-
-    // Driver Program
-    public static void main(String[] args) {
-        Scanner scanner = new Scanner(System.in);
-
-        System.out.print("Enter the number of coins in the tower (N): ");
-        int N = scanner.nextInt();
-
-        System.out.print("Enter the number of coins player can remove in one step (X): ");
-        int X = scanner.nextInt();
-
-        System.out.print("Enter the number of coins player can remove in one step (Y): ");
-        int Y = scanner.nextInt();
-
-        char result = coinTowerWinner(N, X, Y);
-        System.out.println("The winner of the Coin Tower Game is Player " + result + ".");
-
-        // Run JUnit tests
-        org.junit.runner.JUnitCore.main("CoinTowerProblemTest");
-    }
 
     /**
      * Determines the winner of the Coin Tower Game.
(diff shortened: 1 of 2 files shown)


src/main/java/com/thealgorithms/dynamicprogramming/CoinTowerProblem.java
```diff
@@ -0,0 +1,62 @@
...
+    // Driver Program
+    public static void main(String[] args) {
+        int N = 5;
+        int X = 3;
+        int Y = 4;
+
+        char result = coinTowerWinner(N, X, Y);
+        System.out.println("The winner of the Coin Tower Game is Player " + result + ".");
+    }
+
```
vil02, line 22:
```suggestion
```

Please add proper tests into [`src/test/java/com/thealgorithms/dynamicprogramming`](https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms/dynamicprogramming).

vil02, line 22:
```suggestion
```

Please add proper tests into [`src/test/java/com/thealgorithms/dynamicprogramming`](https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms/dynamicprogramming).

vil02 (changes requested):

BamaCharanChhandogi:
I think this type of problem falls under algorithmic questions in the context of Data Structures and Algorithms (DSA).

siriak:
It is an algorithmic question, but not a generalized algorithm. It's an application of dynamic programming to a specific problem. As such, we do not add specific implementations because there's an infinite number of them and they are all very similar. (To be completely fair, we add some to demonstrate the approach, but there are already enough examples in `dynamicprogramming` module).

SarthakChaudhary46:
> @BamaCharanChhandogi, @siriak: having in mind this:
> 
> https://github.com/TheAlgorithms/Java/blob/a216cb8a59ad04b3cadcc156e69dea41f3d1b465/CONTRIBUTING.md?plain=1#L3
> 
> 
> do you think that this pull-requests solves a problem, which is _general enough_?

-I can either add different test cases as suggested by @vil02, or I can make it a general solution by taking custom inputs. Which option should I go with? For reference, I used src/main/java/com/thealgorithms/dynamicprogramming/ClimbingStairs.java, which is a general solution to a LeetCode problem, similar to what I did. @vil02

siriak:
Is there a description of this problem somewhere other than on coding practice websites? Maybe Wikipedia or a book? If so, we can accept it as a popular enough. If no, it's just another problem that has slight variations in the problem statement, but doesn't warrant being included as a different algorithm (again, it's not the algorithm itself, but the dynamic programming approach that we are interested in). As you have mentioned, there are already other examples of dynamic programming in the repository that are very similar to this one.

vil02:
@SarthakChaudhary46 I think we do not need another example demonstrating _dynamic programming_. Unless this is some well-know problem. If yes - could you provide some references?

SarthakChaudhary46:
@vil02 @siriak I have generalized the problem; is it sufficiently comprehensive?

vil02:
> @vil02 @siriak I have generalized the problem; is it sufficiently comprehensive?

No.

I guess you made some mistake in your last commit - I do not see any generalization and the tests are gone. Please check if this was some accident.

siriak:
@vil02 but can it even be generalized? I don't think so. The problem itself is very specific, and I don't think we can help with that. There are already enough examples of dynamic programming in the repository, so I suggest just closing this PR.

vil02:
@SarthakChaudhary46 as discussed earlier, this does not fit into this repository.

If you would be looking for an idea for a contribution, please let me know because I have few.

PR #5050, opened 2024-02-19
Added Extreme Case handler & New Test Case in "CalculateMaxOfMin"

SarthakChaudhary46:

### DIFF
--- a/src/main/java/com/thealgorithms/stacks/CalculateMaxOfMin.java
+++ b/src/main/java/com/thealgorithms/stacks/CalculateMaxOfMin.java
@@ -1,6 +1,6 @@
 /**
- * Author : Siddhant Swarup Mallick
- * Github : https://github.com/siddhant2002
+ * Author : Siddhant Swarup Mallick            || Contributor : Sarthak Chaudhary
+ * Github : https://github.com/siddhant2002    || Github :      https://github.com/SarthakChaudhary46
  */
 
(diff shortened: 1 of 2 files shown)


SarthakChaudhary46:
Guys, I've been thinking about the current implementation of the algorithm, and I believe it's a bit wrong and there's a more efficient approach we could take. Let me walk you through it:

- Iterate Through Each Element: We'll start by going through each element in the array.
- Find the Minimum Subarray: For each element, we'll find the minimum value of the subarray starting from that element and ending at each subsequent element. This way, we're essentially exploring all possible subarrays.
- Keep Track of the Maximum Minimum Value: As we traverse the array, we'll keep track of the maximum of these minimum values encountered.
- Return the Maximum Minimum Value Found: Once we've covered the entire array, we'll return the maximum minimum value found.

For example:

Given an array Arr = {10, 20, 30, 50, 10, 70, 30}, the output should be 30. 

If you all agree with this approach, I'd love to work on implementing it! Let me know your thoughts.

vil02:
The current implementation of `CalculateMaxOfMin` is incorrect. Please have a look at the discussion in #5038. As suggested there: this class should be removed and a new implementation should be added. Would you like to do that?

SarthakChaudhary46:
@vil02 Sure, I'll just open a new pr after making the necessary changes.

PR #5096, opened 2024-04-04
Implement Parentheses Generator

TruongNhanNguyen:

### DIFF
--- a/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
+++ b/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
@@ -7,13 +7,7 @@ import java.util.List;
  * This class generates all valid combinations of parentheses for a given number of pairs using backtracking.
  */
 public final class ParenthesesGenerator {
-
-    /**
-     * Private constructor to prevent instantiation of this class.
-     * This class should be used only for its static methods and cannot be instantiated.
-     */
     private ParenthesesGenerator() {
-        // Private constructor to prevent instantiation
     }
 
     /**
--- a/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
+++ b/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
@@ -12,6 +12,7 @@ public class ParenthesesGenerator {
      *
      * @param n The number of pairs of parentheses.
      * @return A list of strings representing valid combinations of parentheses.
+     * @throws IllegalArgumentException if n is less than 0.
      */
     public static List<String> generateParentheses(int n) {
         if (n < 0) {
--- a/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
+++ b/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
@@ -6,7 +6,10 @@ import java.util.List;
 /**
  * This class generates all valid combinations of parentheses for a given number of pairs using backtracking.
  */
-public class ParenthesesGenerator {
+public final class ParenthesesGenerator {
+
+    private ParenthesesGenerator() {
+    }
     /**
      * Generates all valid combinations of parentheses for a given number of pairs.
      *
@@ -14,7 +17,7 @@ public class ParenthesesGenerator {
      * @return A list of strings representing valid combinations of parentheses.
      * @throws IllegalArgumentException if n is less than 0.
      */
-    public static List<String> generateParentheses(int n) {
+    public static List<String> generateParentheses(final int n) {
         if (n < 0) {
             throw new IllegalArgumentException("The number of pairs of parentheses cannot be nagative");
         }
@@ -32,7 +35,7 @@ public class ParenthesesGenerator {
      * @param close   The number of closed parentheses.
      * @param n       The total number of pairs of parentheses.
      */
-    private static void generateParenthesesHelper(List<String> result, String current, int open, int close, int n) {
+    private static void generateParenthesesHelper(List<String> result, final String current, final int open, final int close, final int n) {
         if (current.length() == n * 2) {
             result.add(current);
             return;
--- a/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
+++ b/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
@@ -14,6 +14,9 @@ public class ParenthesesGenerator {
      * @return A list of strings representing valid combinations of parentheses.
      */
     public static List<String> generateParentheses(int n) {
+        if (n < 0) {
+            throw new IllegalArgumentException("The number of pairs of parentheses cannot be nagative");
+        }
         List<String> result = new ArrayList<>();
         generateParenthesesHelper(result, "", 0, 0, n);
         return result;
--- a/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
+++ b/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
@@ -13,7 +13,7 @@ public class ParenthesesGenerator {
      * @param n The number of pairs of parentheses.
      * @return A list of strings representing valid combinations of parentheses.
      */
-    public List<String> generateParentheses(int n) {
+    public static List<String> generateParentheses(int n) {
         List<String> result = new ArrayList<>();
         generateParenthesesHelper(result, "", 0, 0, n);
         return result;
@@ -28,7 +28,7 @@ public class ParenthesesGenerator {
      * @param close   The number of closed parentheses.
      * @param n       The total number of pairs of parentheses.
      */
-    private void generateParenthesesHelper(List<String> result, String current, int open, int close, int n) {
+    private static void generateParenthesesHelper(List<String> result, String current, int open, int close, int n) {
         if (current.length() == n * 2) {
             result.add(current);
             return;
--- a/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
+++ b/src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
@@ -8,8 +8,14 @@ import java.util.List;
  */
 public final class ParenthesesGenerator {
 
+    /**
+     * Private constructor to prevent instantiation of this class.
+     * This class should be used only for its static methods and cannot be instantiated.
+     */
     private ParenthesesGenerator() {
+        // Private constructor to prevent instantiation
     }
+
     /**
      * Generates all valid combinations of parentheses for a given number of pairs.
      *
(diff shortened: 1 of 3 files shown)


src/test/java/com/thealgorithms/backtracking/ParenthesesGeneratorTest.java
```diff
@@ -0,0 +1,44 @@
...
+        assertEquals(List.of("((()))", "(()())", "(())()", "()(())", "()()()"), result);
+    }
+
+    @Test
+    void testGenerateParenthesesWithFourPairs() {
+        ParenthesesGenerator generator = new ParenthesesGenerator();
+        List<String> result = generator.generateParentheses(4);
+        assertEquals(List.of("(((())))", "((()()))", "((())())", "((()))()", "(()(()))", "(()()())", "(()())()", "(())(())", "(())()()", "()((()))", "()(()())", "()(())()", "()()(())", "()()()()"), result);
+    }
+}
```
vil02, line 44:
```suggestion
package com.thealgorithms.backtracking;

import static org.junit.jupiter.api.Assertions.assertEquals;

import java.util.List;
import java.util.stream.Stream;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.params.ParameterizedTest;
import org.junit.jupiter.params.provider.Arguments;
import org.junit.jupiter.params.provider.MethodSource;

public class ParenthesesGeneratorTest {

    @ParameterizedTest
    @MethodSource("tcStream")
    void numberTests(int input, List<String> expected) {
        final var generator = new ParenthesesGenerator();
        assertEquals(expected, generator.generateParentheses(input));
    }

    private static Stream<Arguments> tcStream() {
        return Stream.of(Arguments.of(0, List.of("")), Arguments.of(1, List.of("()")), Arguments.of(2, List.of("(())", "()()")), Arguments.of(3, List.of("((()))", "(()())", "(())()", "()(())", "()()()")),
            Arguments.of(4, List.of("(((())))", "((()()))", "((())())", "((()))()", "(()(()))", "(()()())", "(()())()", "(())(())", "(())()()", "()((()))", "()(()())", "()(())()", "()()(())", "()()()()")));
    }
}
```

(yes, the formatting is super ugly, but I would highly recommend using _parametrized tests_)

src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
```diff
@@ -0,0 +1,43 @@
...
+ * This class generates all valid combinations of parentheses for a given number of pairs using backtracking.
+ */
+public class ParenthesesGenerator {
+    /**
+     * Generates all valid combinations of parentheses for a given number of pairs.
+     *
+     * @param n The number of pairs of parentheses.
+     * @return A list of strings representing valid combinations of parentheses.
+     */
+    public List<String> generateParentheses(int n) {
```
vil02, line 16: What do you think about throwing an error if `n < 0`?

TruongNhanNguyen: Yes, the code should handle exceptions as well. I will add exception handling for the cases where `n < 0`

src/test/java/com/thealgorithms/backtracking/ParenthesesGeneratorTest.java
TruongNhanNguyen: Well, I just follow the previous test cases implementations in the current code bases. I also realize that your suggestion is better. Then I will update the code using parametrized tests.

TruongNhanNguyen:
> Two questions (just questions, I want to know your reasons):
> 
> * why did you use `List<String>` and not another container? (maybe set),
> * what do you think about marking methods of `ParenthesesGenerator` as `static`?
> 
> In general the code looks good.

1. The `List<String>` data structure was chosen for its simplicity and efficiency in this implementation. Using a `List` allows us maintain the order of generated parentheses combinations, which might be desirable for certain use cases. While a `Set<String>` could ensure uniqueness of combinations, it might not preserve the order of generation, which could be important depending on the application requirements.

2. Marking the methods of `ParenthesesGenerator` as static could be appropriate depending on the intended usage and design considerations. If the methods do not rely on instance-specific state or behavior, marking them as static could enhance code readability and usage simplicity. Static methods can be called without needing to instantiate an object of the class, which can be beneficial in scenarios where the class serves primarily as a utility or helper. However, if the methods need access to instance variables or if there is a possibility of different instances requiring different behaviors, then keeping them as instance methods would be more appropriate.

TruongNhanNguyen:
I will change `ParathensesGenerator`'s methods to `static` later. Appreciate your questions.

src/test/java/com/thealgorithms/backtracking/ParenthesesGeneratorTest.java
TruongNhanNguyen: Done.

src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
```diff
@@ -0,0 +1,47 @@
+package com.thealgorithms.backtracking;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class generates all valid combinations of parentheses for a given number of pairs using backtracking.
+ */
+public class ParenthesesGenerator {
```
vil02, line 9: One can make this class a _proper utility_

```suggestion
public final class ParenthesesGenerator {
    private ParenthesesGenerator() {
    }
```

Going back to the `static`/non-static discussion: please have in mind that as soon as you mark something as `static` you lose the _object oriented magic_ (e.g. polymorphism). Even if the object of a given class does not hold any _state_ it  might make sense to still have it.

In case of this class - I would do it in the same way as it is done now.

src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Generates all valid combinations of parentheses for a given number of pairs.
+     *
+     * @param n The number of pairs of parentheses.
+     * @return A list of strings representing valid combinations of parentheses.
+     * @throws IllegalArgumentException if n is less than 0.
+     */
+    public static List<String> generateParentheses(int n) {
```
vil02, line 17:
```suggestion
    public static List<String> generateParentheses(final int n) {
```

src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Helper function for generating all valid combinations of parentheses recursively.
+     *
+     * @param result  The list to store valid combinations.
+     * @param current The current combination being formed.
+     * @param open    The number of open parentheses.
+     * @param close   The number of closed parentheses.
+     * @param n       The total number of pairs of parentheses.
+     */
+    private static void generateParenthesesHelper(List<String> result, String current, int open, int close, int n) {
```
vil02, line 35:
```suggestion
    private static void generateParenthesesHelper(List<String> result, final String current, final int open, final int close, final int n) {
```

src/test/java/com/thealgorithms/backtracking/ParenthesesGeneratorTest.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.List;
+import java.util.stream.Stream;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class ParenthesesGeneratorTest {
+    @ParameterizedTest
+    @MethodSource("tcStream")
+    void numberTests(int input, List<String> expected) {
```
vil02, line 15: Sorry for a bad name suggestion before (copy-paste error). What do you think about:
```suggestion
    void regularInputTests(int input, List<String> expected) {
```

src/test/java/com/thealgorithms/backtracking/ParenthesesGeneratorTest.java
```diff
@@ -0,0 +1,33 @@
...
+        assertEquals(expected, ParenthesesGenerator.generateParentheses(input));
+    }
+
+    @ParameterizedTest
+    @MethodSource("negativeInputStream")
+    void negativeInputTest(int input) {
```
vil02, line 21:
```suggestion
    void throwsForNegativeInputTest(int input) {
```

vil02 (changes requested):
Almost done!

src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
TruongNhanNguyen: In your suggestion, the `ParenthesesGenerator` class has a private constructor, which means it cannot be instantiated from outside the class itself. This typically means that the class is meant to be used for it static method only and not for creating instances.

In this implementation, I am simply using the static method `generateParentheses` provided by the `ParenthesesGenerator` class. This approach is perfectly fine if the class is designed to provide utility methods that don't rely on instance-specific state.

src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
TruongNhanNguyen: Make the method params to be `final` is a good practice, especially for parameters that are not meant to be reassigned within the method. It helps improve code readability and can prevent unintentional modifications to the parameter value.

In the provided `ParenthesesGenerator` class, marking the method parameters as final wouldn't change the behavior of the methods, but it would make the intention clearer to anyone reading the code.

I will apply your suggestion. :clap:

src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
TruongNhanNguyen: Done.

src/test/java/com/thealgorithms/backtracking/ParenthesesGeneratorTest.java
TruongNhanNguyen: Look good.

src/main/java/com/thealgorithms/backtracking/ParenthesesGenerator.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+public final class ParenthesesGenerator {
+
+    /**
+     * Private constructor to prevent instantiation of this class.
+     * This class should be used only for its static methods and cannot be instantiated.
+     */
+    private ParenthesesGenerator() {
+        // Private constructor to prevent instantiation
+    }
```
vil02, line 17:
```suggestion
public final class ParenthesesGenerator {
    private ParenthesesGenerator() {
    }
```

Such comments are just _noise_. `private` default constructor is a quite common construction in Java. Usually one does not write comments explaining basic syntax.

vil02 (approved):

PR #5116, opened 2024-04-23
Adaptive Replacement Cache (ARC)

devxadarsh:


src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
```diff
@@ -0,0 +1,71 @@
...
+ * @param <K> key type
+ * @param <V> value type
+ */
+
+public class ARCCache<K, V> {
+    private final int capacity;
+    private final Map<K, V> cache;
+    private final LinkedHashMap<K, Integer> usageCounts;
+    private final int p;
+    private int totalCount;
```
TruongNhanNguyen, line 26: Please add comments to explain the usage of these fields. Especially, the `p` field is not used at all, can you explain where and why you used this field?

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
```diff
@@ -0,0 +1,71 @@
...
+
```
TruongNhanNguyen, line 27: I would like to suggest adding docstrings for each method like this
```suggestion
/**
 * Retrieves the value associated with the given key from the cache.
 * If the key is present in the cache, its usage count is incremented.
 *
 * @param key the key whose associated value is to be retrieved
 * @return the value associated with the key, or null if the key is not present in the cache
 */
```

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
```diff
@@ -0,0 +1,71 @@
...
+    public ARCCache(int capacity) {
+        this.capacity = capacity;
+        this.cache = new LinkedHashMap<>();
+        this.usageCounts = new LinkedHashMap<>();
+        this.p = capacity / 2;
+        this.totalCount = 0;
+    }
+
+    // Function to get data from cache corresponding to the key passed as parameter
```
TruongNhanNguyen, line 36:
```suggestion
    /**
     * Retrieves the value associated with the given key from the cache.
     * If the key is present in the cache, its usage count is incremented.
     *
     * @param key the key whose associated value is to be retrieved
     * @return the value associated with the key, or null if the key is not present in the cache
     */
```

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
```diff
@@ -0,0 +1,71 @@
...
+    public V get(K key) {
+        if (cache.containsKey(key)) {
+            usageCounts.put(key, usageCounts.getOrDefault(key, 0) + 1);
+            return cache.get(key);
+        }
+        return null;
+    }
+
+    // Function to put the data in the cache corresponding to the key passed as parameter
```
TruongNhanNguyen, line 45:
```suggestion
    /**
     * Adds the specified key-value pair to the cache.
     * If the cache exceeds its capacity after adding the new entry, eviction is performed.
     * Updates the usage count for the added key.
     *
     * @param key   the key with which the specified value is to be associated
     * @param value the value to be associated with the specified key
     */
```

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
```diff
@@ -0,0 +1,71 @@
...
+    public void put(K key, V value) {
+        if (cache.size() >= capacity) {
+            evict();
+        }
+        cache.put(key, value);
+        usageCounts.put(key, 1);
+        totalCount++;
+    }
+
+    // Function to implement the logic of ARCCache
```
TruongNhanNguyen, line 55:
```suggestion
    /**
     * Evicts an item from the cache when it exceeds its capacity.
     * Implements the Adaptive Replacement Cache (ARC) algorithm logic for eviction.
     * Removes the least recently used item based on its usage count.
     */
```

TruongNhanNguyen:
Please ensure that tests are added to handle lines with partial coverage.

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
devxadarsh: p field was for pivot, I updated to that with a full implementation of it. You can check it out and give me suggestion for it. I am new on open source community. I very happy to Engage with the community. Thanks for your great support .

devxadarsh (commented):
Updated comments

TruongNhanNguyen:
Please format your code using `clang-format`.

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
```diff
@@ -0,0 +1,123 @@
...
+    private final int t1Capacity; // Capacity for the t1 cache
+    private final int b1Capacity; // Capacity for the b1 cache
+    private int totalCount;
+/**
+ * Retrieves the value associated with the given key from the cache.
+ * If the key is present in the cache, its usage count is incremented.
+ *
+ * @param key the key whose associated value is to be retrieved
+ * @return the value associated with the key, or null if the key is not present in the cache
+ */
```
vil02, line 34: This seems to document `get`.

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
```diff
@@ -0,0 +1,123 @@
...
+ *     optimizing cache performance in changing workload scenarios.
+ * <a href="https://en.wikipedia.org/wiki/Adaptive_replacement_cache">...</a>
+ * @author Adarsh Pandey (<a href="https://github.com/devxadarsh">...</a>)
+ *
+ * @param <K> key type
+ * @param <V> value type
+ */
+
+public class ARCCache<K, V> {
+    private final int capacity;
```
vil02, line 22: Does it make sense to store `capacity`? What do you think about introducing a method like

```java
private int capacity() {
    return t1Capacity + b1Capacity;
}
```

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
```diff
@@ -0,0 +1,123 @@
...
+    private final int b1Capacity; // Capacity for the b1 cache
+    private int totalCount;
+/**
+ * Retrieves the value associated with the given key from the cache.
+ * If the key is present in the cache, its usage count is incremented.
+ *
+ * @param key the key whose associated value is to be retrieved
+ * @return the value associated with the key, or null if the key is not present in the cache
+ */
+    public ARCCache(int capacity) {
```
vil02, line 35: What if `capacity` is negative?

vil02 (changes requested):

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
devxadarsh: I removed it and updated

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
devxadarsh: It will be a better approach. Updated the code

src/main/java/com/thealgorithms/datastructures/caches/ARCCache.java
devxadarsh: Added the exception handler to for negative capacity.

vil02:
Please add the _missing_ tests and _update_ your branch.

PR #5142, opened 2024-05-05
Implement SplayTree

TruongNhanNguyen:


src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,323 @@
...
+            minValue = root.left.key;
+            root = root.left;
+        }
+        return minValue;
+    }
+
+    /**
+     * Perform an in-order traversal of the SplayTree.
+     *
+     * @return A vector containing the keys in in-order traversal order.
```
vil02, line 252: vector or list?

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,323 @@
...
+        List<Integer> result = new LinkedList<>();
+        inOrderRec(root, result);
+        return result;
+    }
+
+    /**
+     * Recursive function for in-order traversal.
+     *
+     * @param root   The root of the subtree to traverse.
+     * @param result The vector to store the traversal result.
```
vil02, line 264: vector or list?

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,323 @@
...
+            inOrderRec(root.left, result);
+            result.add(root.key);
+            inOrderRec(root.right, result);
+        }
+    }
+
+    /**
+     * Perform a pre-order traversal of the SplayTree.
+     *
+     * @return A vector containing the keys in pre-order traversal order.
```
vil02, line 277: vector or list?

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,323 @@
...
+        List<Integer> result = new LinkedList<>();
+        preOrderRec(root, result);
+        return result;
+    }
+
+    /**
+     * Recursive function for pre-order traversal.
+     *
+     * @param root   The root of the subtree to traverse.
+     * @param result The vector to store the traversal result.
```
vil02, line 289: vector or list?

vil02 (changes requested):
What do you think about introducing a method: `List<Integer> traverse(TraverseOrder traverseOrder)`, where `TraverseOrder` is an _enum_? This will allow to make the other methods private.

Does your implementation work for duplicated keys?

Please add the _missing_ tests and please update your branch (there were some changes regarding the static analysis).

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: I am confused, I'll update the docstrings

TruongNhanNguyen:
> Does your implementation work for duplicated keys?

_The current implementation of the Splay Tree does not handle duplicate keys explicitly._ Here are a few possible approaches:
1. __Allow Duplicates__: We can modify the implementation to allow duplicate keys. When inserting a node with a duplicate key, we can insert it as a new node in the tree or update an existing node with the same key.
2. __Reject Duplicates__: Alternatively, we can reject duplicate keys altogether. In this case, when inserting a node with a key already in the tree, we can either ignore the insertion or throw an exception to indicate that duplicates are not allowed.
3. __Especially handle Duplicates__: We can define a specific behavior for handling duplicate keys, such as keeping track of the number of occurrences of each key or maintaining a list of values associated with each key.

I think the simplest and most used approach is that when inserting a node with a duplicated key, we should update the existing node with the same key.

What do you think about these approaches, we will choose the one to implement.

vil02:
I think we should not allow duplicate keys and explicitly throw an exception if one is trying to include already existing key. I looked at the existing implementations of BST in this repository and it is a total mess - each implementation reacts differently.

TruongNhanNguyen:
I have updated the test cases, but there are a few partially covered lines.

TruongNhanNguyen:
> I think we should not allow duplicate keys and explicitly throw an exception if one is trying to include an already existing key. I looked at the existing implementations of BST in this repository and it is a total mess - each implementation reacts differently.

I have refactored the implementation to disallow inserting duplicated keys and to throw an error if this occurs. I also handle deletion on an empty tree and simplify the deletion process.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,296 @@
...
+    /**
+     * Deletes a key from the SplayTree.
+     *
+     * @param key The key to delete.
+     * @throws IllegalArgumentException If the tree is empty.
+     */
+    public void delete(int key) {
+        if (root == null) {
+            throw new IllegalArgumentException("Cannot delete from an empty tree");
+        }
```
vil02, line 207: Please add an `isEmpty` method. Otherwise using this class might be very annoying.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,296 @@
...
+        case IN_ORDER:
+            inOrderRec(root, result);
+            break;
+        case PRE_ORDER:
+            preOrderRec(root, result);
+            break;
+        case POST_ORDER:
+            postOrderRec(root, result);
+            break;
+        }
```
vil02, line 246: Please add `default` (cf. #5179).

vil02 (changes requested):

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: I think we can define the `isEmpty` method and use it to check the entire tree's `root` and continue using `root == null` checks within recursive calls because `root` is a local variable that represents the current node being processed in the recursion.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: Thanks

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,308 @@
...
+ *
+ * <p>
+ * Splay trees have applications in various areas, including caching, network routing,
+ * and dynamic optimality analysis.
+ * </p>
+ */
+public class SplayTree {
+
+    private static class Node {
+        int key;
```
vil02, line 31:
```suggestion
        final int key;
```

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,308 @@
...
+                // Recursive call to splay on grandchild
+                root.left.left = splay(root.left.left, key);
+                // Perform zig operation on parent
+                root = rotateRight(root);
+            } // Zig-Zag case
+            else if (root.left.key < key) {
+                // Recursive call to splay on grandchild
+                root.left.right = splay(root.left.right, key);
+                // Perform zag operation on parent
+                if (root.left.right != null) root.left = rotateLeft(root.left);
```
vil02, line 141: This condition is never `false`. Please add a tests covering such situation.

Furthermore:
```suggestion
                if (root.left.right != null) {
                    root.left = rotateLeft(root.left);
                }
```

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,308 @@
...
+                if (root.right.left != null) root.right = rotateRight(root.right);
+            } // Zag-Zig case
+            else if (root.right.key < key) {
+                // Recursive call to splay on grandchild
+                root.right.right = splay(root.right.right, key);
+                // Perform zag operation on parent
+                root = rotateLeft(root);
+            }
+
+            return (root.right == null) ? root : rotateLeft(root);
```
vil02, line 161: Same here.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,308 @@
...
+
+    /**
+     * Search for a key in the SplayTree.
+     *
+     * @param key The key to search for.
+     * @return True if the key is found, otherwise false.
+     */
+    public boolean search(int key) {
+        root = splay(root, key);
+        return root != null && root.key == key;
```
vil02, line 205: Same here. I think that testing `search` on an empty tree will do the thing.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,308 @@
...
+            postOrderRec(root.left, result);
+            postOrderRec(root.right, result);
+            result.add(root.key);
+        }
+    }
+
+    /**
+     * Enum to specify the order of traversal.
+     */
+    public enum TraverseOrder { IN_ORDER, PRE_ORDER, POST_ORDER, INVALID }
```
vil02, line 307:
```suggestion
    public enum TraverseOrder { IN_ORDER, PRE_ORDER, POST_ORDER }
```

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,308 @@
...
+        if (root != null) {
+            postOrderRec(root.left, result);
+            postOrderRec(root.right, result);
+            result.add(root.key);
+        }
+    }
+
+    /**
+     * Enum to specify the order of traversal.
+     */
```
vil02, line 306: There is really no need to document everything. I think in this case reading the code is equally easy as reading the documentation.

```suggestion
```

src/test/java/com/thealgorithms/datastructures/trees/SplayTreeTest.java
```diff
@@ -0,0 +1,125 @@
...
+        SplayTree tree = createComplexTree();
+        assertThrows(IllegalArgumentException.class, () -> tree.insert(value));
+    }
+
+    @Test
+    public void testInvalidTraversalOrderExceptionMessage() {
+        SplayTree tree = createComplexTree();
+        IllegalArgumentException exception = assertThrows(IllegalArgumentException.class, () -> tree.traverse(SplayTree.TraverseOrder.INVALID));
+        assertEquals("Invalid traversal order: INVALID", exception.getMessage());
+    }
```
vil02, line 79: One can skip it. I know that that we will loose the `100%` coverage, but it does not matter in this case.

```suggestion
```

One could consider to skip the `switch` statement and use a _polymorphism_. For example, have an interface `TreeTraversal` with 3 implementations `PreOrder`, `InOrder` and `PostOrder`. The `SplayTree` could use an instance of one of them to perform the desired traversal.

vil02 (changes requested):

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: I always like my code to be understandable. Thanks for your suggestions.

src/test/java/com/thealgorithms/datastructures/trees/SplayTreeTest.java
TruongNhanNguyen: Well, I will try the idea.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: I have added a test for the `search` method on an empty tree.

src/test/java/com/thealgorithms/datastructures/trees/SplayTreeTest.java
TruongNhanNguyen: I have created the `TreeTraversal` interface and 3 implementations of this interface: `PreOrderTraversal`, `InOrderTraversal`, and `PostOrderTraversal`. You can see them in the next commits.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: I removed the `enum TraverseOrder` and implemented 3 classes of the `TreeTraversalOrder` interface instead.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: I have tried to make the tree to be enough complex so that it can cover this case.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: I added the test for searching on an empty tree, I hope it will work and cover this line.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,304 @@
...
+        return new InOrderTraversal();
+    }
+
+    public static TreeTraversal getPreOrderTraversal() {
+        return new PreOrderTraversal();
+    }
+
+    public static TreeTraversal getPostOrderTraversal() {
+        return new PostOrderTraversal();
+    }
```
vil02, line 303:
```suggestion
    public static final TreeTraversal PRE_ORDER = new PreOrderTraversal();
    public static final TreeTraversal IN_ORDER = new InOrderTraversal();
    public static final TreeTraversal POST_ORDER = new PostOrderTraversal();
```

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,304 @@
...
+                root.left.left = splay(root.left.left, key);
+                // Perform zig operation on parent
+                root = rotateRight(root);
+            } // Zig-Zag case
+            else if (root.left.key < key) {
+                root.left.right = splay(root.left.right, key);
+                // Perform zag operation on parent
+                if (root.left.right != null) {
+                    root.left = rotateLeft(root.left);
+                }
```
vil02, line 139: There is no test with `root.left.right == null`.

vil02 (changes requested):

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: Thanks.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: I will test the delete and search methods on more items. I am doing my best to ensure that the tests cover this line.

TruongNhanNguyen: I'm unsure how to get this line covered by the current tests. I've tried searching and deleting multiple keys to trigger the event played by this line, but it hasn't been successful. Is it necessary for this implementation to have a 100% coverage ratio? If so, can you propose a solution to make the tests cover this line? I would appreciate it if you give suggestions for tests that make the implementation perfect.

TruongNhanNguyen:
Can you have a look at the PR @vil02.

src/test/java/com/thealgorithms/datastructures/trees/SplayTreeTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+public class SplayTreeTest {
+
+    @ParameterizedTest
+    @MethodSource("traversalStrategies")
+    public void testTraversal(SplayTree.TreeTraversal traversal, List<Integer> expected) {
+        SplayTree tree = createComplexTree();
+        List<Integer> result = tree.traverse(traversal);
+        assertEquals(expected, result);
+    }
```
vil02, line 22: In order to have the _100%_ coverage, I think that this this test should be also parametrized over the `tree`. Could you please rewrite this test?

src/test/java/com/thealgorithms/datastructures/trees/SplayTreeTest.java
```diff
@@ -0,0 +1,110 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class SplayTreeTest {
+
+    @ParameterizedTest
+    @MethodSource("traversalStrategies")
+    public void testTraversal(SplayTree.TreeTraversal traversal, List<Integer> expected) {
+        SplayTree tree = createComplexTree();
+        List<Integer> result = tree.traverse(traversal);
+        assertEquals(expected, result);
```
vil02, line 21:
```suggestion
        assertEquals(expected, tree.traverse(traversal));
```

vil02 (changes requested):

src/test/java/com/thealgorithms/datastructures/trees/SplayTreeTest.java
TruongNhanNguyen: Sure, I will do it

TruongNhanNguyen: If necessary, I'll open another pull request to achieve 100% code coverage. I don't have much time to do it in this PR. Thanks for your suggestion @vil02. Can we merge this PR into the repo when the coverage is not 100%?

vil02: > Can we merge this PR into the repo when the coverage is not 100%?

No.

TruongNhanNguyen:
I have added a separate test for the uncovered line, I hope it covers that line.

siriak (commented):
@alxkm @vil02 please take a look

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,306 @@
...
+        Node(int key) {
+            this.key = key;
+            left = null;
+            right = null;
+        }
+    }
+
+    private Node root;
+
+    public SplayTree() {
```
alxkm, line 44: Default constructor is generated by compiler by default. And root is by default null. Maybe remove it, or it is here for other reasons (code clarity, plugin checks)?

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,306 @@
...
+        return root;
+    }
+
+    /**
+     * Search for a key in the SplayTree.
+     *
+     * @param key The key to search for.
+     * @return True if the key is found, otherwise false.
+     */
+    public boolean search(int key) {
```
alxkm, line 207: What do you think about methods reordering? Because usually standard ordering is assumed such order:

1. Class (static) variables: First the public class variables, then the protected, and then the private.

2. Instance variables: First public, then protected, and then private.

3. Constructors

4. Methods: should be grouped by functionality rather than by scope or accessibility. For example, a private class method can be in between two public instance methods. The goal is to make reading and understanding the code easier.

https://www.oracle.com/java/technologies/javase/codeconventions-fileorganization.html

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,306 @@
...
+        public void traverse(Node root, List<Integer> result) {
+            if (root != null) {
+                traverse(root.left, result);
+                traverse(root.right, result);
+                result.add(root.key);
+            }
+        }
+    }
+
+    public static final TreeTraversal PRE_ORDER = new PreOrderTraversal();
```
alxkm, line 303: According to standard Java code convention, it will be better to have them at beginning of the class.

alxkm (changes requested):

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * Deletes a key from the SplayTree.
+     *
+     * @param key The key to delete.
+     * @throws IllegalArgumentException If the tree is empty.
+     */
+    public void delete(int key) {
+        if (isEmpty()) {
+            throw new IllegalArgumentException("Cannot delete from an empty tree");
```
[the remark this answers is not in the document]
TruongNhanNguyen: Thanks for bringing up this excellent point.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: Just remove it.

src/main/java/com/thealgorithms/datastructures/trees/SplayTree.java
TruongNhanNguyen: These best practices should be followed.

TruongNhanNguyen:
Can you have a look, @alxkm?

alxkm:
@TruongNhanNguyen, everything looks good. However, the build is currently failing. You need to add a serialVersionUID field to the exception classes, similar to this:

`private static final long serialVersionUID = 1L;`

Additionally, you haven't marked the key field as final in a few places. While it's optional, declaring it as final would indicate that the parameter is immutable and establish a constraint that can help prevent future modifications.

TruongNhanNguyen:
Let's have a look, @alxkm. I have updated the code based on the proposals and removed the redundant tests that are not necessary anymore. Note that, the current tests do not fully cover the code, there is still 1 line partially covered.

siriak (approved):

Issue #5119, opened 2024-04-25
[OTHER] Enable more checks in checkstyle


vil02:
### What would you like to share?

#5110 introduced checkstyle. The [config file](https://github.com/TheAlgorithms/Java/blob/master/checkstyle.xml) has some of the rules _commented out_. It might a good idea to introduce some of them and discuss why the other ones are not needed in this repository.

## List of checks which should be introduced

Few of the _obvious_ checks, which should be introduced:
- [x] ConstantName (#5139 by @marysiuniq)
- [x] LocalFinalVariableName (#5172 by @CodaBlurd)
- [x] LocalVariableName (#5191 by @cpu-pixel)
- [x] MemberName (#5193 by @cpu-pixel)
- [x] MethodName (#5182 by @CodaBlurd)
- [x] ParameterName (#5196 by @CodaBlurd)
- [x] StaticVariableName (#5173 by @vaibhav9t1)
- [x] TypeName (#5214 by @StarDxxx)
- [x] AvoidStarImport (#5141)
- [x] ModifierOrder (#5132 by @marysiuniq)
- [x] RedundantModifier (#5140)
- [x] EmptyStatement (#5120)
- [x] MultipleVariableDeclarations (#5175 by @vaibhav9t1)
- [x] HideUtilityClassConstructor (#5147)
- [x] ArrayTypeStyle (#5145)
- [x] InvalidJavadocPosition (#5237 by @samuelfac)
- [x] OperatorWrap (#5212 by @StarDxxx)
- [x] ParenPad (#5226 by @samuelfac)
- [x] EqualsHashCode (#5157 by @CodaBlurd)
- [x] InnerAssignment (#5162 by @CodaBlurd)
- [x] FinalClass (#5154  by @CodaBlurd)
- [x] MissingSwitchDefault (#5179 by @BamaCharanChhandogi)
- [x] NeedBraces (#5227 by @samuelfac)
- [x] AvoidNestedBlocks (#5228 by @samuelfac)
- [x] WhitespaceAround (#5241)
- [x] TodoComment (#5785 by @nishchalgv1)

(@siriak, @BamaCharanChhandogi feel free to edit the list above)

## How to contribute

1. select one _check_ form the list above such that no one else is working on (and is not done yet),
2. write a comment "Hey, I am working on `XYZ`", where `XYZ` is your selected check,
3. in your fork, please create a branch like enable_XYZ,
[...]

PR #5157, opened 2024-05-12
style: enable EqualsHashCode in checkstyle

CodaBlurd:
Related to #5119.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -161,7 +161,7 @@
     <!-- Checks for common coding problems               -->
     <!-- See https://checkstyle.org/checks/coding/index.html -->
     <module name="EmptyStatement"/>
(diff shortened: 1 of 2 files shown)


vil02 (approved):

vil02:
Thanks!

I have two remarks:
- if you want to reference an issue or another PR (from the same repository) do it by writing `#` followed by its number,
- it seems to me that the method `equals` could be simplified to the form:
```java
        @Override
        public boolean equals(Object obj) {
            if (obj instanceof Point other) {
                return Arrays.equals(other.coordinates, this.coordinates);
            }
            return false;
        }
```
Would you like to open another PR doing that?

CodaBlurd:
Yes I will open another PR with the recommended changes. Thank you!

PR #5236, opened 2024-06-18
Implement Smooth Sort

JAPNITSINGH:


JAPNITSINGH:
Additionaly done the following to fix pervious issues:

- [x] I have run `mvn pmd:check` to ensure there are no pmd errors
- [x] I have run `mvn checkstyle:checkstyle` to ensure correct style is followed

src/main/java/com/thealgorithms/sorts/SmoothSort.java
vil02, line 1: Is it possible to derive your class from [`SortAlgorithm`](https://github.com/TheAlgorithms/Java/blob/39e065437c4c70d9d99c0de3343a980f423d5bc5/src/main/java/com/thealgorithms/sorts/SortAlgorithm.java)?

src/test/java/com/thealgorithms/sorts/SmoothSortTest.java
vil02, line 1: If the answer to the above questions is _yes_, then if should be enough to derive the `SmoothSortTest` from [`SortingAlgorithmTest`](https://github.com/TheAlgorithms/Java/blob/39e065437c4c70d9d99c0de3343a980f423d5bc5/src/test/java/com/thealgorithms/sorts/SortingAlgorithmTest.java).

vil02 (changes requested):

src/test/java/com/thealgorithms/sorts/SmoothSortTest.java
JAPNITSINGH: I believe it is possible, I just have to use generic `T[]` in place of `Integer[]` modify the code a little bit and use `.compareTo()` and it should be good. I shall work on it and update this PR. I'll comment here once changes are ready.

JAPNITSINGH: I have updated the code, haven't pushed it yet. The smooth sort algorithm I implemented seems to be failing for most arrays of size greater than 700. I guess the problem is with how I am calculating the right child of a node. I am having a look into it. I shall provide an update here once I am done with the fix

JAPNITSINGH: I have updated to code,  derived the class from SortAlgorithm, removed unnecessary return statements , renamed variables to make the code more understandable.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+ * Wikipedia: https://en.wikipedia.org/wiki/Smoothsort
+ */
+public final class SmoothSort implements SortAlgorithm {
+
+    public SmoothSort() {
+    }
+
+    private static Integer[] getLeonardoNumbers() {
+        Integer[] leonardoNumbers = {1, 1, 3, 5, 9, 15, 25, 41, 67, 109, 177, 287, 465, 753, 1219, 1973, 3193, 5167, 8361, 13529, 21891, 35421, 57313, 92735, 150049, 242785, 392835, 635621, 1028457, 1664079, 2692537, 4356617, 7049155, 1405773, 18454929, 29860703, 48315633, 78176337, 126491971,
+            204668309, 331160281, 535828591};
```
vil02, line 17: Why not to use [`LeonardoNumber`](https://github.com/TheAlgorithms/Java/blob/a710fe11c47bdfb80342f851c189d167035b7b1c/src/main/java/com/thealgorithms/maths/LeonardoNumber.java) here?

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+            int j = i;
+            while (array[prevRootNodeIndex].compareTo(array[currentRootNodeIndex]) > 0) {
+                int currentLeonardoLevel = currentLeonardoTreeLevels[j];
+                if (currentLeonardoLevel > 1) {
+                    // compare child and swap
+
+                    int indexOfRightChild = rootNodeIndices.get(j) - 1; // right child is of level n-2
+                    int indexOfLeftChild = rootNodeIndices.get(j) - 1 - getLeonardoNumbers()[currentLeonardoLevel - 2];
+                    if (array[prevRootNodeIndex].compareTo(array[indexOfRightChild]) > 0 && array[prevRootNodeIndex].compareTo(array[indexOfLeftChild]) > 0) {
+                        swap(array, prevRootNodeIndex, currentRootNodeIndex);
```
vil02, line 158:
```suggestion
                        SortUtils.swap(array, prevRootNodeIndex, currentRootNodeIndex);
```

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+                        rootNodeIndexForHeapify = prevRootNodeIndex;
+                        leonardoTreeLevelforHeapify = currentLeonardoTreeLevels[j - 1];
+                    } else {
+                        maxHeapifyLeonardoTree(currentRootNodeIndex, currentLeonardoLevel, array);
+                    }
+                } else {
+                    // swap
+                    swap(array, prevRootNodeIndex, currentRootNodeIndex);
```
vil02, line 166:
```suggestion
                    SortUtils.swap(array, prevRootNodeIndex, currentRootNodeIndex);
```

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+        }
+
+        maxHeapifyLeonardoTree(rootNodeIndexForHeapify, leonardoTreeLevelforHeapify, array); // for the last tree if needed
+    }
+
+    private static <T> void swap(T[] array, int idx, int idy) {
+        T swap = array[idx];
+        array[idx] = array[idy];
+        array[idy] = swap;
+    }
```
vil02, line 184:
```suggestion
```

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+
+        if (array[rightChildIndex].compareTo(array[leftChildIndex]) >= 0) {
+            childIndexForSwap = rightChildIndex;
+        } else {
+            childIndexForSwap = leftChildIndex;
+        }
+
+        if (array[childIndexForSwap].compareTo(array[currentRootNodeIndex]) > 0) {
+            // swap(And keep on swapping I guess, I did not implement that which might be causing issue?)
+            swap(array, currentRootNodeIndex, childIndexForSwap);
```
vil02, line 212:
```suggestion
            SortUtils.swap(array, currentRootNodeIndex, childIndexForSwap);
```

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+        // maxHeapifyLeonardoTree(leftChildIndex, currentLeonardoLevel - 1, array);
+
+        if (array[rightChildIndex].compareTo(array[leftChildIndex]) >= 0) {
+            childIndexForSwap = rightChildIndex;
+        } else {
+            childIndexForSwap = leftChildIndex;
+        }
+
+        if (array[childIndexForSwap].compareTo(array[currentRootNodeIndex]) > 0) {
+            // swap(And keep on swapping I guess, I did not implement that which might be causing issue?)
```
vil02, line 211: ?

src/main/java/com/thealgorithms/sorts/SmoothSort.java
vil02, line 166:
```suggestion
                    SortUtils.swap(array, prevRootNodeIndex, currentRootNodeIndex);
```
No need for the comment. The name of the method `swap` is pretty clear.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+                        maxHeapifyLeonardoTree(currentRootNodeIndex, currentLeonardoLevel, array);
+                    }
+                } else {
+                    // swap
+                    swap(array, prevRootNodeIndex, currentRootNodeIndex);
+                    rootNodeIndexForHeapify = prevRootNodeIndex;
+                    leonardoTreeLevelforHeapify = currentLeonardoTreeLevels[j - 1];
+                }
+                j = j - 1;
+                if (j == i - 1) {
```
vil02, line 171: `j == i - 1` is never `false` (among the existing test cases).

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+            return; // Trees with one node are in already max-heapified.
+        }
+
+        int currentRootNodeIndex = rootNodeIndex;
+        int rightChildIndex = rootNodeIndex - 1;
+        int leftChildIndex = rootNodeIndex - getLeonardoNumbers()[currentLeonardoLevel - 2] - 1;
+        int childIndexForSwap = -1;
+
+        // maxHeapifyLeonardoTree(rightChildIndex, currentLeonardoLevel - 2, array);
+        // maxHeapifyLeonardoTree(leftChildIndex, currentLeonardoLevel - 1, array);
```
vil02, line 202: ?

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+            if (consecutiveTreeIndices[0] != -1) {
+                // if 0th or 1st index is -1 that implies there are no concequtive trees
+                leonardoLevelTracker = leonardoLevelTracker & ~(1 << consecutiveTreeIndices[0]);
+                leonardoLevelTracker = leonardoLevelTracker & ~(1 << consecutiveTreeIndices[1]);
+                leonardoLevelTracker = leonardoLevelTracker | (1 << consecutiveTreeIndices[1] + 1);
+            } else if ((leonardoLevelTracker & 2) == 0) {
+                leonardoLevelTracker = leonardoLevelTracker | (1 << 1);
+            } else {
+                leonardoLevelTracker = leonardoLevelTracker | (1 << 0);
+            }
```
vil02, line 44: Is it possible to apply the [`SingleBitOperations`](https://github.com/TheAlgorithms/Java/blob/a710fe11c47bdfb80342f851c189d167035b7b1c/src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperations.java) (or something from [`bitmanipulation`](https://github.com/TheAlgorithms/Java/tree/master/src/main/java/com/thealgorithms/c))?

vil02 (changes requested):
At the moment the code is way too complex. I would suggest to:
- implement `LeonardoHeap` (or use one of the existing implementations in [`heaps`](https://github.com/TheAlgorithms/Java/tree/master/src/main/java/com/thealgorithms/datastructures/heaps)),
- use it in the implementation of `SmoothSort`.

JAPNITSINGH:
> At the moment the code is way too complex. I would suggest to:
> 
> * implement `LeonardoHeap` (or use one of the existing implementations in [`heaps`](https://github.com/TheAlgorithms/Java/tree/master/src/main/java/com/thealgorithms/datastructures/heaps)),
> * use it in the implementation of `SmoothSort`.

I did try to use existing implementaions of Heaps, however Leonardo Heap is a bit different from the others since this heap contains multiple loenardo trees for very specific shapes and the root node being the right most elemnent in the list. 

I can crate a new class `LeonardoHeap` in [heaps](https://github.com/TheAlgorithms/Java/tree/master/src/main/java/com/thealgorithms/datastructures/heaps), however I believe I might not endup using the interface [Heap.java](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/heaps/Heap.java) in this implementation, would that be alright?

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,227 @@
...
+            int prevRootNodeIndex = rootNodeIndices.get(i - 1);
+            int j = i;
+            while (array[prevRootNodeIndex].compareTo(array[currentRootNodeIndex]) > 0) {
+                int currentLeonardoLevel = currentLeonardoTreeLevels[j];
+                if (currentLeonardoLevel > 1) {
+                    // compare child and swap
+
+                    int indexOfRightChild = rootNodeIndices.get(j) - 1; // right child is of level n-2
+                    int indexOfLeftChild = rootNodeIndices.get(j) - 1 - getLeonardoNumbers()[currentLeonardoLevel - 2];
+                    if (array[prevRootNodeIndex].compareTo(array[indexOfRightChild]) > 0 && array[prevRootNodeIndex].compareTo(array[indexOfLeftChild]) > 0) {
```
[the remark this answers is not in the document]
JAPNITSINGH: Sure, I shall add a specific test case in `SmoothSortTest.java`.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
JAPNITSINGH: The [LeonardoNumber](https://github.com/TheAlgorithms/Java/blob/a710fe11c47bdfb80342f851c189d167035b7b1c/src/main/java/com/thealgorithms/maths/LeonardoNumber.java) class is recursively calculating the Leonardo Numbers. I did not want to calculate(or re-calculate) the numbers everytime. I shall modify my code and use [LeonardoNumber](https://github.com/TheAlgorithms/Java/blob/a710fe11c47bdfb80342f851c189d167035b7b1c/src/main/java/com/thealgorithms/maths/LeonardoNumber.java) to store the results in an array before I start creating the heap and use the results stored in that array when needed.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
JAPNITSINGH: This comment was added during debugging, I shall remove this in the next commit.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
JAPNITSINGH: I do have a test case (not currently committed) with around 900 entries where this evalueates to false during creation of the heap. I can try to reduce the entires in such a way that the condition still evaluates to true, I will for sure end up having an array of size greater than 30 ( as i will need 3 trees of non consecutive levels .... L1 L3 and L5 at least ). 

Would it be fine if I added a test case in `SmoothSortTest.java` with around 30 elements in the test? If yes, I shall work on it at the very end after resolving the other conversations.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
JAPNITSINGH: This too was added for debugging, I shall remove this in my next commit. please ignore this.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
JAPNITSINGH: I did not come accross this class earlier. 
Yes, this class can definately be used. I shall modify this section of the code as well.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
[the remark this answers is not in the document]
JAPNITSINGH: I see why this is happening and I shall try my best to explain in this comment. 

Leonardo Heap has the following three properties that define it (Going through the array from left to right):
1. The trees are in decreasing order of levels
2. Root nodes of the trees are in increasing order
3. Each tree is in a 'max-heap' state. 

The Level-1 leonardo Tree(L1) and Level 0 leonardo Tree (L0) are the 'base case' trees as they have exactly one node. 

At any given valid leonardo heap, if the heap contains both L1 and L0 tree then the element in L0 tree will always be greater than element at L1 tree, because of the second property mentioned above ( L0 >= L1 )

When L2 tree is being constructed, the level0 tree becomes the right child level1 tree becomes the left child. And hence it is not possible to have a condition such that level0 < prevRootNode < level1 as it would be a voilation of the above. 

[This](https://en.wikipedia.org/wiki/Smoothsort#Sifting_down) section of the wiki page and [this ](https://www.keithschwarz.com/smoothsort/) article (under the image`An erroneous swap` ) mention that root needs to be bigger than the new element and the roots of its child nodes for a swap to happen. However [this ](https://www.enterag.ch/hartwig/order/smoothsort.pdf) article(in page 6) mentions that the new elment should exceede its largest son. 

I believe the statement in wiki is more genreic statement than a technical one and this condition can be optimized to just having `array[prevRootNodeIndex].compareTo(array[indexOfLeftChild]) > 0`. I shall however look around for more articles that suggest otherwise before I make changes to this.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
JAPNITSINGH: @vil02 I would however think recrsivly calculating these numbers ( before or during the execution ) might not be a great idea and would just increase the execution time as it will run the same recursion stacks multiple times. If you still suggest using LeonardoNumber.java I shall go ahead and make changes as mentioned above.

src/main/java/com/thealgorithms/sorts/SmoothSort.java
JAPNITSINGH: Update: Yes, this condition is never false, I shall take a look and update the condition. I belive what i intented to do was run `maxHeapifyLeonardoTree` if you reached the right most tree and further swaps are still present. I am checking this as well.

JAPNITSINGH:
Adding this comment to provide updates and ensure this PR is not marked as stale: 

I have refactored the code, not yet pushed as I want to do some more changes and try to use the interface [Heap.java](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/heaps/Heap.java)
I shall comment here once the changes are ready to be reviewed.

JAPNITSINGH:
Hi @vil02 , 

I have made some modifications to the code, can you please have a look and provide comments on it. 

1. I have created a new class LeonardoHeap where I have moved all the heap's functionality. 
2. SmoothSort.java is now just putting elements into the heap and extracting max elements from the heap and putting it in the array. 
3. I had tried to use the interface [Heap.java](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/heaps/Heap.java) however it did not quite work out properly with the `generic T` so I ended up not using it. 
4. The condition `array[prevRootNodeIndex].compareTo(array[indexOfRightChild]) > 0` evaluating to `true` and `array[prevRootNodeIndex].compareTo(array[indexOfLeftChild]) > 0` to `false` now happens (I was greedly max-heapfying earlier which might have been the problem but it seems to be fixed now)
5. I have updated the condition on when "maxHeapifyLeonardoTree" is supposed to be called (earlier it was being called everytime)
6. I am using [LeonardoNumber](https://github.com/TheAlgorithms/Java/blob/a710fe11c47bdfb80342f851c189d167035b7b1c/src/main/java/com/thealgorithms/maths/LeonardoNumber.java) in my code, however I strongly feel it is better to use dynamic programing instead of recursively calculating the solutions everytime.
7. I have used [SingleBitOperations](https://github.com/TheAlgorithms/Java/blob/a710fe11c47bdfb80342f851c189d167035b7b1c/src/main/java/com/thealgorithms/bitmanipulation/SingleBitOperations.java) for updating leonardoLevelTracker

src/main/java/com/thealgorithms/sorts/SmoothSort.java
vil02: @JAPNITSINGH we can define this array as `static`. For time being, I do not have a clear preference. The option for the future, is definitely to provide a way to compute Leonardo numbers more efficiently.

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
vil02, line 1: I need to take a closer look on this one. In the meantime: could you please create `LeonardoHeapTest.java`?

vil02 (changes requested):

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
JAPNITSINGH: Yes sure, I am a bit occupied for the next couple of days, and will start on creating a new test file after that.

JAPNITSINGH: I have added the LeonardoHeapTest.java file, there weren't many HeapTest files that I could refer to, so I have added the test cases of my own, please let me know if something in those tests needs to be improved or changed.

JAPNITSINGH:
Hi @vil02 , are any further changes required in the test cases? Or in the code perhaps?

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,31 @@
...
+        LeonardoHeap<T> leonardoHeap = new LeonardoHeap<T>();
+
+        for (int i = 0; i < array.length; i++) {
+            leonardoHeap.addElement(array[i]);
+        }
+
+        for (int i = 0; i < array.length; i++) {
+            final T maxElement = leonardoHeap.removeElement();
+            array[array.length - i - 1] = maxElement;
+        }
```
vil02, line 23:
```suggestion
        for (final var element : array) {
            leonardoHeap.addElement(element);
        }

        for (int i = 0; i < array.length; i++) {
            array[array.length - i - 1] = leonardoHeap.removeElement();
        }
```

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,225 @@
...
+import java.util.List;
+
+/**
+ * Wikipedia: https://en.wikipedia.org/wiki/Smoothsort
+ */
+public class LeonardoHeap<T extends Comparable<T>> {
+
+    private int leonardoLevelTracker;
+    private int leonardoHeapSize;
+    private final List<T> leonardoHeap;
```
vil02, line 15:
```suggestion
    private int leonardoLevelTracker = 0;
    private int leonardoHeapSize = 0;
    private final List<T> leonardoHeap = new ArrayList<T>();
```

Furthermore:
- is that true that `leonardoHeapSize == leonardoHeap.size()`? If yes, I would remove `leonardoHeapSize`,
- please consider renaming these members - the `leonardo` prefix is redundant.

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,225 @@
...
+
+    public LeonardoHeap() {
+        this.leonardoHeap = new ArrayList<T>();
+        this.leonardoLevelTracker = 0;
+        this.leonardoHeapSize = 0;
+    }
+
+    public int getHeapsize() {
```
vil02, line 23:
```suggestion
    public int getHeapSize() {
```

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
vil02, line 1: Some of the pieces of this class are not covered by unit tests. Please run the shell command like
```shell
mvn clean && mvn test -Dtest=LeonardoHeapTest
```
and examine the generated report `target/site/jacoco/index.html`.

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,225 @@
...
+                    answer.set(1, i);
+                }
+                prevOneIndex = i;
+            } else {
+                prevOneIndex = -1;
+            }
+            num >>>= 1;
+        }
+        return answer;
+    }
```
vil02, line 192: This should be move to a separate class.

vil02 (changes requested):

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
JAPNITSINGH: I can create a new class for this, probably with the name of `LeonardoTreeHelper.java` and add it's respective test file. 

The other thing that can be done is that we can move this function to one of the bit manipulation class? Since all this function is doing is returning first pair of indixes of consecutive 1 bits in an integer. 

Which among these seems the right option @vil02?

JAPNITSINGH:
Pushed some changes, I am yet to push some more. I will request review once done.

JAPNITSINGH:
Hi @vil02 I have made the changes listed below and needed a few inputs:

Changes:
1. Used for-each loop in `SmoothSort.java`
2. Removed `leonardoHeapSize` 
3. Removed redundant `leonardo`  prefix in `LeonardoHeap.java`
4. Moved the initialization outside the constructor in `LeonardoHeap.java`
5. Removed `getHeapsize` method. 
6. Modified `LeonardoHeapTest.java` to provide 100% coverage on running `mvn clean && mvn test -Dtest=LeonardoHeapTest`

I wanted your inputs on this part: I have added a `LeonardoHeapHelper.java` class where I have moved `findConsecutiveLeonardoTreeIndices` and `findAllTreeIndices` to the Helper class. I wanted to know if this class( and it's function names) should be renamed and moved to the bitwise operations package, or do you wish to place them else where. 

For context :
`findConsecutiveLeonardoTreeIndices`  takes in a number as an input and returns the indeices first occourence of consecutive 1 bits in it's binary form. 
`findAllTreeIndices` takes in a number as an input and returns all bits that are 1 in it's binary form (in decending order)

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,173 @@
...
+        // BASE CASE
+        if (currentLevel == 0 || currentLevel == 1) {
+            return; // Trees with one node are in already max-heapified.
+        }
+
+        int currentRootNodeIndex = rootNodeIndex;
+        int rightChildIndex = rootNodeIndex - 1;
+        int leftChildIndex = rootNodeIndex - LeonardoNumber.leonardoNumber(currentLevel - 2) - 1;
+        int childIndexForSwap = -1;
+
```
vil02, line 56:
```suggestion
        final int rightChildIndex = rootNodeIndex - 1;
        final int leftChildIndex = rootNodeIndex - LeonardoNumber.leonardoNumber(currentLevel - 2) - 1;
        final int childIndexForSwap = (heap.get(rightChildIndex).compareTo(heap.get(leftChildIndex)) >= 0) ? rightChildIndex : leftChildIndex;

        final int currentRootNodeIndex = rootNodeIndex;
```

Yes, the formatting in this repo is ugly.

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,173 @@
...
+            swap(currentRootNodeIndex, childIndexForSwap);
+            if (childIndexForSwap == rightChildIndex) {
+                maxHeapifyTree(rightChildIndex, currentLevel - 2);
+            } else { // swap happened with the left child
+                maxHeapifyTree(leftChildIndex, currentLevel - 1);
+            }
+        }
+    }
+
+    private void shiftRootAndRestoreHeap() {
```
vil02, line 73: Do you see some change to _chop_ this method into few smaller ones?

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,173 @@
...
+        // iterate over the currentTreeLevels and get roots
+
+        for (int i = 0; i < currentTreeLevels.length; i++) {
+            rootNodeIndices.add(previousTreeSizeCumulative + LeonardoNumber.leonardoNumber(currentTreeLevels[i]) - 1);
+            previousTreeSizeCumulative = previousTreeSizeCumulative + LeonardoNumber.leonardoNumber(currentTreeLevels[i]);
+        }
+
+        int rootNodeIndexForHeapify = rootNodeIndices.getLast();
+        int treeLevelforHeapify = currentTreeLevels[currentTreeLevels.length - 1];
+        boolean swaped = false;
```
vil02, line 93:
```suggestion
        boolean swapped = false;
```
You can consider moving this declaration _into_ the loop (smaller scope is always better). This should work, because this variable is always `false` at the end of of each loop.

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        return position;
+    }
+
+    private void swap(int i, int j) {
+        T temp = heap.get(i);
+        heap.set(i, heap.get(j));
+        heap.set(j, temp);
+    }
```
vil02, line 158:
```suggestion
    private void swap(final int i, final int j) {
        Collections.swap(heap, i, j);
    }
```

make sure to
```java
import java.util.Collections;
```

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,173 @@
...
+
+    public void addElement(T element) {
+        increaseLevelTracker();
+        heap.add(element);
+        shiftRootAndRestoreHeap();
+    }
+
+    public T removeElement() {
+        decreaseLevelTracker();
+        T element = heap.removeLast();
```
vil02, line 168:
```suggestion
        final T element = heap.removeLast();
```

vil02 (changes requested):

src/main/java/com/thealgorithms/sorts/SmoothSort.java
```diff
@@ -0,0 +1,30 @@
+package com.thealgorithms.sorts;
+
+import com.thealgorithms.datastructures.heaps.LeonardoHeap;
+
+/**
+ * Wikipedia: https://en.wikipedia.org/wiki/Smoothsort
+ */
+public final class SmoothSort implements SortAlgorithm {
+
+    public SmoothSort() {
```
alxkm, line 10: You don't need to write an empty contstructor; the Java compiler will automatically insert one for you (only if you haven't defined any other constructors that take some args).

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeapHelper.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+// TODO: CHeck file name. Check if the file needs to be here or in bitwose operations.
+public final class LeonardoHeapHelper {
+
+    private LeonardoHeapHelper() {
+    }
+
+    public static ArrayList<Integer> findConsecutiveTreeIndices(int num) {
```
alxkm, line 12: Better use List<Integer> here as It's best to return the most generic type that's appropriate for your interface.

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeapHelper.java
```diff
@@ -0,0 +1,45 @@
...
+        int prevOneIndex = -1;
+        int currentLevel = 0;
+
+        ArrayList<Integer> answer = new ArrayList<Integer>();
```
alxkm, line 16: Same here, better to:

`List<Integer> answer = new ArrayList<>();`

src/test/java/com/thealgorithms/datastructures/heaps/LeonardoHeapHelperTest.java
```diff
@@ -0,0 +1,74 @@
...
+        ArrayList<Integer> arrayList = LeonardoHeapHelper.findConsecutiveTreeIndices(0);
+        ArrayList<Integer> expectedList = new ArrayList<>(Arrays.asList(-1, -1));
+
+        assertEquals(expectedList, arrayList);
+    }
+
+    @Test
+    public void testConsecutiveTreeIndicesForEleven() {
+        ArrayList<Integer> arrayList = LeonardoHeapHelper.findConsecutiveTreeIndices(11);
+        ArrayList<Integer> expectedList = new ArrayList<>(Arrays.asList(0, 1));
```
alxkm, line 24: Looks like use `List.of()` is more appropriate here:

`List<Integer> expectedList = new ArrayList<>(List.of(0, 1));`

src/test/java/com/thealgorithms/datastructures/heaps/LeonardoHeapTest.java
```diff
@@ -0,0 +1,100 @@
...
+        heap.addElement(5);
+
+        assertEquals(20, heap.removeElement());
+        assertEquals(10, heap.removeElement());
+        assertEquals(5, heap.removeElement());
+    }
+
+    @Test
+    public void testAddElementStrings() {
+        LeonardoHeap<String> heap = new LeonardoHeap<String>();
```
alxkm, line 33: Why to do not write just:

`LeonardoHeap<String> heap = new LeonardoHeap<>();`

As `String `already used in type declaration, other part will be infered.

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
JAPNITSINGH: Might be possible, I'll take a look into using some form of recursion instead of the while loop I am using. Recursion might give a bit more readability.

JAPNITSINGH:
Pushed some changes removed the Helper class and added `final` access modifier to variable that weren't subject to change, I am yet to push some more changes (that is the changes to `shiftRootAndRestoreHeap` ). I will request review once done.

JAPNITSINGH:
Hey @vil02 , I have made some changes:

1. removed the Helper class 
2. added final access modifier to variable that weren't subject to change
3. Added a few more functions to split `shiftRootAndRestoreHeap` into smaller chuncks make the code more readable

I tried to get rid of the `while` loop and use recursion but I ran into problem where I wanted the function to return two things `rootNodeIndexForHeapify `and `treeLevelforHeapify`, and I coulnd't achieve that in a clean way. So the loop exists in the code for now and I have optimized the way I used `swapped` (now renamed to `swapRequired`)

JAPNITSINGH:
Hi @vil02, would you kindly have a look and let me know if there are more changes to be made?

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
```diff
@@ -0,0 +1,203 @@
...
+                boolean swapRequired = compareChildren(currentRootNodeIndex, prevRootNodeIndex, currentLevel);
+                if (swapRequired) {
+                    // compare child and swap
+                    swap(prevRootNodeIndex, currentRootNodeIndex);
+                    rootNodeIndexForHeapify = prevRootNodeIndex;
+                    treeLevelforHeapify = currentTreeLevels[j - 1];
+                } else {
+                    maxHeapifyTree(currentRootNodeIndex, currentLevel);
+                }
+                j = j - 1;
```
vil02, line 96:
```suggestion
                --j;
```

src/main/java/com/thealgorithms/datastructures/heaps/LeonardoHeap.java
vil02: I would suggest to _refactor_ this method into a shorter one.

src/test/java/com/thealgorithms/datastructures/heaps/LeonardoHeapTest.java
```diff
@@ -0,0 +1,121 @@
...
+        assertEquals(33, heap.removeElement());
+        assertEquals(33, heap.removeElement());
+        assertEquals(29, heap.removeElement());
+        assertEquals(28, heap.removeElement());
+        assertEquals(15, heap.removeElement());
+        assertEquals(13, heap.removeElement());
+        assertEquals(12, heap.removeElement());
+        assertEquals(5, heap.removeElement());
+        assertEquals(2, heap.removeElement());
+    }
```
vil02, line 120: I added some more `assertEquals` here. A test, which does not _assert_ can lead to confusion. Please have a look if it makes sense to _parametrize_ these tests.

The `LeonardoHeap` misses a method `isEmpty()`. There is simply no way to know if user cans still `removeElement`.

vil02 (changes requested):

src/test/java/com/thealgorithms/datastructures/heaps/LeonardoHeapTest.java
JAPNITSINGH: @vil02 , I had a public heap size function earlier which was being used to check if the heap is empty (or heap size is 0). It was removed in one of the previous conversations. I can explicitly add an isEmpty() function instead of the heap size one.

vil02: @JAPNITSINGH what can I say: it is my bad. Sorry for that. But honestly: I think having a `isEmpty()` method makes sense.

JAPNITSINGH: Alright, I'll make the changes in a couple of days and update here.

JAPNITSINGH:
Hi @vil02 , I need a liitle more time to complete the tasks, I am a bit occupird with my college. I shall update here as soon as I get time to cmplete this.

PR #5244, opened 2024-06-20
Improving BitonicSort

alxkm:
The previous sorting algorithm only worked with arrays whose sizes were powers of two. This issue has been fixed.

Additionally, the code has been refactored to follow a common approach, and tests for the sort function have been added.

### DIFF
--- a/src/main/java/com/thealgorithms/sorts/BitonicSort.java
+++ b/src/main/java/com/thealgorithms/sorts/BitonicSort.java
@@ -1,6 +1,7 @@
 package com.thealgorithms.sorts;
 
 import java.util.Arrays;
+import java.util.function.BiPredicate;
 
 /**
  * BitonicSort class implements the SortAlgorithm interface using the bitonic sort technique.
@@ -15,38 +16,38 @@ public class BitonicSort implements SortAlgorithm {
      * Sorts the given array using the Bitonic Sort algorithm.
      *
      * @param <T> the type of elements in the array, which must implement the Comparable interface
-     * @param unsorted the array to be sorted
+     * @param array the array to be sorted
      * @return the sorted array
      */
     @Override
-    public <T extends Comparable<T>> T[] sort(T[] unsorted) {
-        if (unsorted == null || unsorted.length == 0) {
-            return unsorted;
+    public <T extends Comparable<T>> T[] sort(T[] array) {
+        if (array.length == 0) {
+            return array;
         }
 
-        final int paddedSize = nextPowerOfTwo(unsorted.length);
-        T[] paddedArray = Arrays.copyOf(unsorted, paddedSize);
+        final int paddedSize = nextPowerOfTwo(array.length);
+        T[] paddedArray = Arrays.copyOf(array, paddedSize);
 
         // Fill the padded part with a maximum value
-        final T maxValue = max(unsorted);
-        Arrays.fill(paddedArray, unsorted.length, paddedSize, maxValue);
+        final T maxValue = max(array);
+        Arrays.fill(paddedArray, array.length, paddedSize, maxValue);
 
         bitonicSort(paddedArray, 0, paddedSize, Direction.ASCENDING);
-        return Arrays.copyOf(paddedArray, unsorted.length);
+        return Arrays.copyOf(paddedArray, array.length);
     }
 
-    private <T extends Comparable<T>> void bitonicSort(T[] arr, int low, int cnt, Direction dir) {
+    private <T extends Comparable<T>> void bitonicSort(final T[] array, final int low, final int cnt, final Direction dir) {
         if (cnt > 1) {
             final int k = cnt / 2;
 
             // Sort first half in ascending order
-            bitonicSort(arr, low, k, Direction.ASCENDING);
+            bitonicSort(array, low, k, Direction.ASCENDING);
 
             // Sort second half in descending order
-            bitonicSort(arr, low + k, cnt - k, Direction.DESCENDING);
+            bitonicSort(array, low + k, cnt - k, Direction.DESCENDING);
 
             // Merge the whole sequence in ascending order
-            bitonicMerge(arr, low, cnt, dir);
+            bitonicMerge(array, low, cnt, dir);
         }
     }
 
@@ -57,15 +58,15 @@ public class BitonicSort implements SortAlgorithm {
      * @param arr the array containing the bitonic sequence to be merged
      * @param low the starting index of the sequence to be merged
      * @param cnt the number of elements in the sequence to be merged
-     * @param dir the direction of sorting: true for ascending, false for descending
+     * @param dir the direction of sorting
      */
-    private <T extends Comparable<T>> void bitonicMerge(T[] arr, int low, int cnt, Direction dir) {
+    private <T extends Comparable<T>> void bitonicMerge(final T[] arr, final int low, final int cnt, final Direction dir) {
         if (cnt > 1) {
             final int k = cnt / 2;
 
+            final BiPredicate<T, T> areSorted = (dir == Direction.ASCENDING) ? (a, b) -> a.compareTo(b) < 0 : (a, b) -> a.compareTo(b) > 0;
             for (int i = low; i < low + k; i++) {
-                boolean condition = (dir == Direction.ASCENDING) ? arr[i].compareTo(arr[i + k]) > 0 : arr[i].compareTo(arr[i + k]) < 0;
-                if (condition) {
+                if (!areSorted.test(arr[i], arr[i + k])) {
                     SortUtils.swap(arr, i, i + k);
                 }
             }
@@ -105,13 +106,7 @@ public class BitonicSort implements SortAlgorithm {
      * @return the maximum element in the array
      * @throws IllegalArgumentException if the array is null or empty
      */
-    private static <T extends Comparable<T>> T max(T[] array) {
-        T max = array[0];
-        for (T element : array) {
-            if (SortUtils.greater(element, max)) {
-                max = element;
-            }
-        }
-        return max;
+    private static <T extends Comparable<T>> T max(final T[] array) {
+        return Arrays.stream(array).max(Comparable::compareTo).orElseThrow();
     }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/com/thealgorithms/sorts/BitonicSort.java
```diff
@@ -1,79 +1,109 @@
...
+        int paddedSize = nextPowerOfTwo(n);
+        T[] paddedArray = Arrays.copyOf(arr, paddedSize);
+
+        // Fill the padded part with a maximum value
+        T maxValue = findMax(arr);
+        Arrays.fill(paddedArray, n, paddedSize, maxValue);
+
+        bitonicSort(paddedArray, 0, paddedSize, true);
+
+        return Arrays.copyOf(paddedArray, n);
```
vil02, line 33:
```suggestion
        final int paddedSize = nextPowerOfTwo(arr.length);
        T[] paddedArray = Arrays.copyOf(arr, paddedSize);

        // Fill the padded part with a maximum value
        final T maxValue = findMax(arr);
        Arrays.fill(paddedArray, arr.length, paddedSize, maxValue);

        bitonicSort(paddedArray, 0, paddedSize, true);

        return Arrays.copyOf(paddedArray, arr.length);
```

src/main/java/com/thealgorithms/sorts/BitonicSort.java
```diff
@@ -1,79 +1,109 @@
...
 
-    /* It recursively sorts a bitonic sequence in ascending
-  order, if dir = 1, and in descending order otherwise
-  (means dir=0). The sequence to be sorted starts at
-  index position low, the parameter cnt is the number
-  of elements to be sorted.*/
-    void bitonicMerge(int[] a, int low, int cnt, int dir) {
+    private <T extends Comparable<T>> void bitonicSort(T[] arr, int low, int cnt, boolean dir) {
         if (cnt > 1) {
             int k = cnt / 2;
```
vil02, line 38:
```suggestion
            final int k = cnt / 2;
```

vil02 (changes requested):
I have few more remarks, which I will share tomorrow.

alxkm:
There are issues with the build, not related to any recent code changes. The problem seems to be with the Codecov tool.

src/main/java/com/thealgorithms/sorts/BitonicSort.java
```diff
@@ -1,79 +1,94 @@
...
+        return Arrays.copyOf(paddedArray, arr.length);
     }
 
-    /* It recursively sorts a bitonic sequence in ascending
-  order, if dir = 1, and in descending order otherwise
-  (means dir=0). The sequence to be sorted starts at
-  index position low, the parameter cnt is the number
-  of elements to be sorted.*/
-    void bitonicMerge(int[] a, int low, int cnt, int dir) {
+    private <T extends Comparable<T>> void bitonicSort(T[] arr, int low, int cnt, boolean dir) {
```
vil02, line 34: What do you think about introducing:
```java
private enum Direction {
    DESCENDING,
    ASCENDING,
}
```

src/main/java/com/thealgorithms/sorts/BitonicSort.java
```diff
@@ -1,79 +1,94 @@
...
+    /**
+     * Sorts the given array using the Bitonic Sort algorithm.
+     *
+     * @param <T> the type of elements in the array, which must implement the Comparable interface
+     * @param arr the array to be sorted
+     * @return the sorted array
+     */
+    @Override
+    public <T extends Comparable<T>> T[] sort(T[] arr) {
+        if (arr == null || arr.length == 0) {
```
vil02, line 19: As far as I see the sorts derived from `SortAlgorithm` assume that the input is not `null`. I think it is better to be consistent here.

src/test/java/com/thealgorithms/sorts/BitonicSortTest.java
```diff
@@ -0,0 +1,10 @@
+package com.thealgorithms.sorts;
+
+public class BitonicSortTest extends SortingAlgorithmTest {
+    private final BitonicSort bitonicSort = new BitonicSort();
+
+    @Override
+    SortAlgorithm getSortAlgorithm() {
+        return bitonicSort;
```
vil02, line 8: Why not simply? It should be rather cheap to create an instance of `BitonicSort`.
```suggestion
    @Override
    SortAlgorithm getSortAlgorithm() {
        return new BitonicSort();
```

src/main/java/com/thealgorithms/sorts/SortUtils.java
```diff
@@ -116,4 +116,26 @@ public static <T extends Comparable<T>> boolean isSorted(List<T> list) {
...
+
+    /**
+     * Finds the maximum element in the given array.
+     *
+     * @param <T> the type of elements in the array, which must implement the Comparable interface
+     * @param array the array to be searched
+     * @return the maximum element in the array
+     * @throws IllegalArgumentException if the array is null or empty
+     */
+    public static <T extends Comparable<T>> T max(T[] array) {
```
vil02, line 128:
```suggestion
    private static <T extends Comparable<T>> T max(T[] array) {
```

src/main/java/com/thealgorithms/sorts/SortUtils.java
```diff
@@ -116,4 +116,26 @@ public static <T extends Comparable<T>> boolean isSorted(List<T> list) {
...
+        if (array == null || array.length == 0) {
+            throw new IllegalArgumentException("Array must not be null or empty");
+        }
```
vil02, line 131: When this method is marked as `private` I am not really sure if this is needed. I do not have a strong opinion.

vil02 (changes requested):

vil02:
> There are issues with the build, not related to any recent code changes. The problem seems to be with the Codecov tool.

Codecov behaves strange for few days.

src/main/java/com/thealgorithms/sorts/BitonicSort.java
alxkm: Agree. Code using enums is easier to read.

src/main/java/com/thealgorithms/sorts/BitonicSort.java
alxkm: Actually, there are two test cases that will not be covered if we remove this:

shouldAcceptWhenEmptyArrayIsPassed
shouldAcceptWhenEmptyListIsPassed

These are part of the SortingAlgorithmTest.

vil02 (changes requested):
The code looks good. Here are some small remarks to make it _perfect_.

src/main/java/com/thealgorithms/sorts/BitonicSort.java
```diff
@@ -1,79 +1,117 @@
...
+     * @param cnt the number of elements in the sequence to be merged
+     * @param dir the direction of sorting: true for ascending, false for descending
+     */
+    private <T extends Comparable<T>> void bitonicMerge(T[] arr, int low, int cnt, Direction dir) {
+        if (cnt > 1) {
+            final int k = cnt / 2;
+
+            for (int i = low; i < low + k; i++) {
+                boolean condition = (dir == Direction.ASCENDING) ? arr[i].compareTo(arr[i + k]) > 0 : arr[i].compareTo(arr[i + k]) < 0;
+                if (condition) {
```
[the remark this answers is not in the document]
alxkm: Possibly. I've made the change. It might be more complicated, but it definitely looks clearer.

vil02:
> The same for codecov. "Failed to properly create report: The process '/home/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1"

It seems to be relates to codecov/codecov-action#1487.

src/main/java/com/thealgorithms/sorts/BitonicSort.java
```diff
@@ -1,79 +1,112 @@
...
+        bitonicSort(paddedArray, 0, paddedSize, Direction.ASCENDING);
+        return Arrays.copyOf(paddedArray, array.length);
     }
 
-    /* This funcion first produces a bitonic sequence by
-  recursively sorting its two halves in opposite sorting
-  orders, and then calls bitonicMerge to make them in
-  the same order */
-    void bitonicSort(int[] a, int low, int cnt, int dir) {
+    private <T extends Comparable<T>> void bitonicSort(final T[] array, final int low, final int cnt, final Direction dir) {
```
vil02, line 39:
```suggestion
    private <T extends Comparable<T>> void bitonicSort(final T[] array, final int low, final int count, final Direction direction) {
```
Regarding `final T[] array` - I do not have a strong opinion here. Formally it is correct.

src/main/java/com/thealgorithms/sorts/BitonicSort.java
```diff
@@ -1,79 +1,112 @@
...
+    /**
+     * Merges the bitonic sequence in the specified direction.
+     *
+     * @param <T> the type of elements in the array, which must be Comparable
+     * @param arr the array containing the bitonic sequence to be merged
+     * @param low the starting index of the sequence to be merged
+     * @param cnt the number of elements in the sequence to be merged
+     * @param dir the direction of sorting
+     */
+    private <T extends Comparable<T>> void bitonicMerge(final T[] arr, final int low, final int cnt, final Direction dir) {
```
vil02, line 63:
```suggestion
    private <T extends Comparable<T>> void bitonicMerge(final T[] array, final int low, final int count, final Direction direction) {
```

vil02 (changes requested):

src/main/java/com/thealgorithms/sorts/BitonicSort.java
alxkm: I wasn't sure either, which is why I removed them. They are likely redundant since this is a very internal method for sorting. We typically make something final to prevent modification, but in this case, as an internal method, it shouldn't need to be modified.

vil02 (approved):

vil02:
I cannot merge it now, because the _upload to codecov_ fails. I will rerun the build in a while, so maybe it will pass.

PR #5248, opened 2024-06-23
fix: avoid infinite loop in `SwapSort`

### LINKED ISSUE #5249: [BUG] Infinite Loop in Swap Sort Implementation
### Description

Description
The current implementation of the Swap Sort algorithm can enter an infinite loop, particularly when dealing with arrays containing duplicate values. This issue was observed in the test case shouldAcceptWhenArrayWithDuplicateValueIsPassed.

Problem
The index variable in the sorting logic is not being updated correctly. As a result, the index may never reach len - 1, causing the algorithm to loop indefinitely.

Solution
Fix the logic to ensure the index variable is updated appropriately.
Add the missing test case to validate the sorting of arrays with duplicate values.
Steps to Reproduce

### Steps to reproduce

Run the test case shouldAcceptWhenArrayWithDuplicateValueIsPassed.
Observe that the algorithm enters an infinite loop.
Expected Behavior

### Excepted behavior

The algorithm should sort the array correctly and terminate without entering an infinite loop.

### Screenshots

_No response_

### Additional context

_No response_

alxkm:
There was an issue where this sorting implementation could enter an infinite loop, such as in the test shouldAcceptWhenArrayWithDuplicateValueIsPassed. 

The index variable was not being updated correctly.
The index may never reach len - 1 due to the flawed logic, leading to a potential infinite loop. 

This fix addresses the issue and adds the missing test for this sorting algorithm.

Fixes #5249.

### DIFF
--- a/src/main/java/com/thealgorithms/sorts/SwapSort.java
+++ b/src/main/java/com/thealgorithms/sorts/SwapSort.java
@@ -9,25 +9,32 @@ package com.thealgorithms.sorts;
  */
 public class SwapSort implements SortAlgorithm {
 
-    /**
-     * Sorts the input array using the swap sort algorithm.
-     *
-     * @param array the array to be sorted
-     * @param <T>   the type of elements in the array, which must be Comparable
-     * @return the sorted array
-     */
     @Override
     public <T extends Comparable<T>> T[] sort(T[] array) {
         int len = array.length;
+        int index = 0;
 
-        for (int i = 0; i < len - 1; i++) {
-            for (int j = i + 1; j < len; j++) {
-                if (SortUtils.less(array[j], array[i])) {
-                    SortUtils.swap(array, i, j);
-                }
+        while (index < len - 1) {
+            int amountSmallerElements = this.getSmallerElementCount(array, index);
+
+            if (amountSmallerElements > 0) {
+                SortUtils.swap(array, index, index + amountSmallerElements);
+            } else {
+                index++;
             }
         }
 
         return array;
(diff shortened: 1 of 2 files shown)


src/main/java/com/thealgorithms/sorts/SwapSort.java
vil02, line 1: I think this has become some bubble sort. In order to fix this issue, I would suggest something like this:
```java
public class SwapSort implements SortAlgorithm {

    @Override
    public <T extends Comparable<T>> T[] sort(T[] array) {
        int len = array.length;
        int index = 0;

        while (index < len - 1) {
            int amountSmallerElements = this.getSmallerElementCount(array, index);

            if (amountSmallerElements > 0) {
                SortUtils.swap(array, index, index + amountSmallerElements);
            } else {
                index++;
            }
        }

        return array;
    }

    private <T extends Comparable<T>> int getSmallerElementCount(T[] array, int index) {
        int counter = 0;
        for (int i = index + 1; i < array.length; i++) {
            if (SortUtils.less(array[i], array[index])) {
                counter++;
            }
        }

        return counter;
    }
}
```

vil02 (changes requested):

src/main/java/com/thealgorithms/sorts/SwapSort.java
alxkm: You are right. Thanks. Missed that such variant looks like bubble sort.

vil02 (approved):

PR #5275, opened 2024-07-03
refactor: cleanup `CountingSort`

alxkm:
Cleanup CountingSort
Adding test for it

Extracted method CountingSort using stream into separate class also added test for it


src/main/java/com/thealgorithms/sorts/CountingSort.java
```diff
@@ -1,93 +1,49 @@
...
-        List<T> sortedArray = new ArrayList<>(list.size());
+        // TreeMap to maintain order of elements naturally.
+        Map<T, Integer> frequencyMap = new TreeMap<>();
 
-        // Counting the frequency of @param array elements
-        list.forEach(v -> frequency.put(v, frequency.getOrDefault(v, 0) + 1));
+        // Count the frequency of each element in the list.
+        for (T element : list) {
+            frequencyMap.put(element, frequencyMap.getOrDefault(element, 0) + 1);
+        }
```
vil02, line 37: This part should be moved to a method like `computeHistogram` or `computeFequencyMap` (_method name is a best comment, because nobody reads comments, and some people read method names_).

src/main/java/com/thealgorithms/sorts/CountingSort.java
```diff
@@ -1,93 +1,49 @@
...
-        // Filling the sortedArray
-        for (Map.Entry<T, Integer> element : frequency.entrySet()) {
-            for (int j = 0; j < element.getValue(); j++) {
-                sortedArray.add(element.getKey());
+        // Create the sorted list based on the frequency map.
+        List<T> sortedList = new ArrayList<>(list.size());
+        for (Map.Entry<T, Integer> entry : frequencyMap.entrySet()) {
+            for (int i = 0; i < entry.getValue(); i++) {
+                sortedList.add(entry.getKey());
             }
```
vil02, line 44: Similarly here: maybe `extractSorted(frequencyMap)`.

vil02 (changes requested):

src/main/java/com/thealgorithms/sorts/CountingSort.java
alxkm: Ok, thanks. Thought that refactoring should be very light to keep refactoring near to original code. But for better understanding, move code to method is a better approach.

src/main/java/com/thealgorithms/sorts/CountingSort.java
alxkm: Yes, thought that maybe in can be named as extractSortedArrayFromFrequencyMap, but map is already as parameter

src/main/java/com/thealgorithms/sorts/CountingSort.java
```diff
@@ -1,93 +1,53 @@
...
-        System.out.println("Before Sorting:");
-        print(unsortedStrings);
-
-        // Output => a a b c c d e f g
-        System.out.println("After Sorting:");
-        print(countingSort.sort(unsortedStrings));
-
-        System.out.println("After Sorting By Streams:");
-        print(streamSort(unsortedStrings));
+    private static <T extends Comparable<T>> Map<T, Integer> computeFequencyMap(List<T> list) {
```
vil02, line 45:
```suggestion
    private static <T extends Comparable<T>> Map<T, Integer> computeFequencyMap(final List<T> list) {
```

src/main/java/com/thealgorithms/sorts/CountingSort.java
```diff
@@ -1,93 +1,53 @@
...
+        // TreeMap to maintain order of elements naturally.
```
vil02, line 46:
```suggestion
```

src/main/java/com/thealgorithms/sorts/CountingSort.java
```diff
@@ -1,93 +1,53 @@
...
+        Map<T, Integer> frequencyMap = new TreeMap<>();
+        for (T element : list) {
```
vil02, line 48:
```suggestion
        for (final T element : list) {
```

src/main/java/com/thealgorithms/sorts/CountingSort.java
```diff
@@ -1,93 +1,53 @@
...
+        return extractSortedArray(frequencyMap);
+    }
 
-        // Filling the sortedArray
-        for (Map.Entry<T, Integer> element : frequency.entrySet()) {
-            for (int j = 0; j < element.getValue(); j++) {
-                sortedArray.add(element.getKey());
+    private static <T extends Comparable<T>> List<T> extractSortedArray(Map<T, Integer> frequencyMap) {
+        List<T> sortedList = new ArrayList<>();
+        for (Map.Entry<T, Integer> entry : frequencyMap.entrySet()) {
```
vil02, line 37:
```suggestion
        for (final Map.Entry<T, Integer> entry : frequencyMap.entrySet()) {
```

src/main/java/com/thealgorithms/sorts/CountingSort.java
```diff
@@ -1,93 +1,53 @@
...
     @Override
     public <T extends Comparable<T>> List<T> sort(List<T> list) {
-        Map<T, Integer> frequency = new TreeMap<>();
-        // The final output array
-        List<T> sortedArray = new ArrayList<>(list.size());
-
-        // Counting the frequency of @param array elements
-        list.forEach(v -> frequency.put(v, frequency.getOrDefault(v, 0) + 1));
+        Map<T, Integer> frequencyMap = computeFequencyMap(list);
+        return extractSortedArray(frequencyMap);
```
vil02, line 32: In order to stop all of the discussions about the type, name and if it should be `final`, I would simply
```suggestion
        return extractSortedArray(computeFequencyMap(list));
```

src/main/java/com/thealgorithms/sorts/CountingSort.java
```diff
@@ -1,93 +1,53 @@
...
-
-        // Output => a a b c c d e f g
-        System.out.println("After Sorting:");
-        print(countingSort.sort(unsortedStrings));
-
-        System.out.println("After Sorting By Streams:");
-        print(streamSort(unsortedStrings));
+    private static <T extends Comparable<T>> Map<T, Integer> computeFequencyMap(List<T> list) {
+        // TreeMap to maintain order of elements naturally.
+        Map<T, Integer> frequencyMap = new TreeMap<>();
```
vil02, line 47: I think that in the context of _cout sort_ it seams more natural to have a hash map (amortised $O(n)$ vs. $O(n \log n)$). 

```suggestion
        Map<T, Integer> frequencyMap = new HashMap<>();
```

vil02 (changes requested):

src/main/java/com/thealgorithms/sorts/CountingSort.java
alxkm: This TreeMapprovide us with ordering of keys, formation of a sorted array. And also helps to avoid standard code for countSort:

```
    public static int[] sort(int[] array) {
        if (array.length == 0) {
            return array;
        }

        int max = array[0];
        int min = array[0];

        for (int value : array) {
            if (value > max) {
                max = value;
            }
            if (value < min) {
                min = value;
            }
        }

        int range = max - min + 1;
        int[] count = new int[range];

        for (int value : array) {
            count[value - min]++;
        }

        int[] output = new int[array.length];
        int index = 0;

        for (int i = 0; i < range; i++) {
            while (count[i] > 0) {
                output[index++] = i + min;
                count[i]--;
            }
        }

        return output;
    }
```

So if it is removed, we will have to write another sorting implementation.

vil02: This class does not implement [counting sort](https://en.wikipedia.org/wiki/Counting_sort).

alxkm: Yes, but this class already was before my changes, in sorting algorithms, and I just wanted to make some refactoring, to make it more corresponding according to other sorting algorithms, to general approach. If we will remove TreeMap this current approach will break. Possible options can be rename this class to CountingSortUsingTreeMap and create separate traditional implementation of it.

What you think, we need todo?

vil02: Your update is a absolutely a good thing to do.

The implementation from the very beginning was not a _counting sort_.

One thing to do is to simply... delete this class. The reasons for that is simply, that in certain situations, the implementation can lead to strange bugs. Imagine that you work with a type `IntWithId`, which has two fields: `value` and `id`. `value` is used to compare the objects of that class. Consider the input like `[(4, 1), (3, 2), (3, 3), (3, 4)]`. The output will be like [(3, 2), (3, 2), (3, 2), (4, 1)].

Second argument: imagine someone preparing for an exam about sorting algorithms. And then, on the exam, there is a question _What is counting sort?_

alxkm: Educational exam case fully describe complexity of naming situation.

I don't know resolve my last changes the issue. But I have added canonical counting sort, which is pretty standard for any educational purposes. 

And renamed previous sorting with index Pseudo to highlight that it uses counting sort idea but it doesn't match with counting sort.

alxkm:
Hello @vil02 . Could you please advise on how to handle the Pseudo Counting Sort classes? To preserve the integrity and original purpose of this repository, it might be best to remove them since they do not actually implement Counting Sort. However, they have been part of the repository for some time, and I'm unsure of the best course of action.

vil02:
@alxkm: I would suggest to remove them. I would guess, the original implementation was merged by accident. Introducing some `QuasiPseudoAlmostCountingSort` does not make much sense: it does not have any special properties.

Therefore I suggest to:
- remove these classes (with this PR, or close this one and open a new one; from documentation point of view the second option is better),
- implement a _proper_ `CountingSort` (i.e. only nonnegative integers as an input) in a new PR.

alxkm:
> @alxkm: I would suggest to remove them. I would guess, the original implementation was merged by accident. Introducing some `QuasiPseudoAlmostCountingSort` does not make much sense: it does not have any special properties.
> 
> Therefore I suggest to:
> 
> * remove these classes (with this PR, or close this one and open a new one; from documentation point of view the second option is better),
> * implement a _proper_ `CountingSort` (i.e. only nonnegative integers as an input) in a new PR.
> 
> In any case: thanks to you we have spotted a quite serious bug!

Ok, thank you. Got it. This is really proper way to resolve this issue

alxkm:
This implementation  of CountingSort is wrong. Because it works only using TreeMap, which causes wrong asymptotic for CountingSort. 

This issue will be resolved in such way:

- Removing existed Counting sort in separate PR 
- Add correct implementation for Counting Sort (this is already exist in this pull request)

All these discussed in this PR

PR #5280, opened 2024-07-09
refactor: cleanup `RadixSort`

alxkm:
Cleanup RadixSort

Adding test

Update DIRECTORY.md


src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -2,61 +2,58 @@
...
+     * @return the sorted array
+     */
+    public static int[] sort(int[] array) {
+        if (array.length == 0) {
+            return array;
         }
-        return mx;
+
+        final int max = Arrays.stream(array).max().getAsInt();
+        for (int exp = 1; max / exp > 0; exp *= 10) {
```
vil02, line 25: Please consider using [`NumberOfDigits`](https://github.com/TheAlgorithms/Java/blob/57878cac555eebf33a86b3d808e7b91f26174458/src/main/java/com/thealgorithms/maths/NumberOfDigits.java) to determine the number of iterations in the loop.

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -2,61 +2,58 @@
...
+     */
+    private static void countingSortByDigit(int[] array, int exp) {
+        int[] output = new int[array.length];
         int[] count = new int[10];
-        Arrays.fill(count, 0);
 
-        for (i = 0; i < n; i++) {
-            count[(arr[i] / exp) % 10]++;
+        for (int i = 0; i < array.length; i++) {
+            count[(array[i] / exp) % 10]++;
```
vil02, line 43: What do you think about introducing a method like `private static int getDigit(int number, int position)`?

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -2,61 +2,58 @@
...
-    private static void countSort(int[] arr, int n, int exp) {
-        int[] output = new int[n];
-        int i;
+    /**
+     * A utility method to perform counting sort of array[] according to the digit represented by exp.
+     *
+     * @param array the array to be sorted
+     * @param exp   the exponent representing the current digit position
+     */
+    private static void countingSortByDigit(int[] array, int exp) {
```
vil02, line 38: I think think that method could be split into smaller pieces like in #5287.

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -2,61 +2,58 @@
...
+        int[] output = new int[array.length];
         int[] count = new int[10];
```
vil02, line 40: What do you think about introducing
```java
private static final int BASE = 10;
```

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -2,61 +2,58 @@
...
-            if (arr[i] > mx) {
-                mx = arr[i];
-            }
+    /**
+     * Sorts an array of integers using the radix sort algorithm.
+     *
+     * @param array the array to be sorted
+     * @return the sorted array
+     */
+    public static int[] sort(int[] array) {
```
vil02, line 19: The current implementation crashes for negative inputs.

src/test/java/com/thealgorithms/sorts/RadixSortTest.java
```diff
@@ -0,0 +1,23 @@
...
+    @MethodSource("provideTestCases")
+    public void test(int[] inputArray, int[] expectedArray) {
+        assertArrayEquals(RadixSort.sort(inputArray), expectedArray);
+    }
+
+    private static Stream<Arguments> provideTestCases() {
+        return Stream.of(Arguments.of(new int[] {170, 45, 75, 90, 802, 24, 2, 66}, new int[] {2, 24, 45, 66, 75, 90, 170, 802}), Arguments.of(new int[] {3, 3, 3, 3}, new int[] {3, 3, 3, 3}), Arguments.of(new int[] {9, 4, 6, 8, 14, 3}, new int[] {3, 4, 6, 8, 9, 14}),
+            Arguments.of(new int[] {10, 90, 49, 2, 1, 5, 23}, new int[] {1, 2, 5, 10, 23, 49, 90}), Arguments.of(new int[] {1, 3, 4, 2, 7, 8}, new int[] {1, 2, 3, 4, 7, 8}), Arguments.of(new int[] {}, new int[] {}), Arguments.of(new int[] {1}, new int[] {1}),
+            Arguments.of(new int[] {1, 2, 3, 4, 5, 6, 7, 8, 9}, new int[] {1, 2, 3, 4, 5, 6, 7, 8, 9}), Arguments.of(new int[] {9, 8, 7, 6, 5, 4, 3, 2, 1}, new int[] {1, 2, 3, 4, 5, 6, 7, 8, 9}),
+            Arguments.of(new int[] {1000000000, 999999999, 888888888, 777777777}, new int[] {777777777, 888888888, 999999999, 1000000000}));
```
vil02, line 21:
```suggestion
            Arguments.of(new int[] {1000000000, 999999999, 888888888, 777777777}, new int[] {777777777, 888888888, 999999999, 1000000000}), Arguments.of(new int[] {123, 9, 54321, 123456789, 0}, new int[] {0, 9, 123, 54321, 123456789}));
```
I would suggest a test case with the inputs having different magnitude.

vil02 (changes requested):

src/test/java/com/thealgorithms/sorts/RadixSortTest.java
alxkm: Thank you. For this type of comparison based on "counting and comparing numbers", this is a very representative test.

src/main/java/com/thealgorithms/sorts/RadixSort.java
alxkm: Yes. But I will add this. But implementation will be more complicated

src/main/java/com/thealgorithms/sorts/RadixSort.java
alxkm: This change will definitely improve the readability of the code.

src/main/java/com/thealgorithms/sorts/RadixSort.java
alxkm: Thanks, I don't know about these methods. It's kind of redundant. Because code will look like this:

```
for (int i = 0, exp = 1; i < NumberOfDigits.numberOfDigits(max); i++, exp *= BASE) {
    countingSortByDigit(array, exp);
}
```

But it will improve the readability of the code and its understanding.

src/main/java/com/thealgorithms/sorts/RadixSort.java
alxkm: Thanks, need to add. This is a complex expression.

src/main/java/com/thealgorithms/sorts/RadixSort.java
alxkm: When I wrote the initial comment, I thought it would be challenging to add support for negative numbers. However, the implementation for handling negative numbers has turned out to be quite straightforward. This is because the same radix sort logic used for positive numbers can be applied to negative numbers as well.

Thanks to your feedback on the pull request, the code is now highly modular, and each method is clearly defined and distinguishable.

Here is the part that handles negative numbers:

```
int[] negatives = Arrays.stream(array).filter(x -> x < 0).map(x -> -x).toArray();
if (negatives.length > 0) {
    radixSort(negatives);
    reverseAndNegate(negatives);
}

```
And here are the two methods involved:
```
private static void reverseAndNegate(int[] array) {
    for (int i = 0; i < array.length / 2; i++) {
        int temp = array[i];
        array[i] = -array[array.length - 1 - i];
        array[array.length - 1 - i] = -temp;
    }
    if (array.length % 2 != 0) {
        array[array.length / 2] = -array[array.length / 2];
    }
}

private static void mergeNegativeAndPositiveArrays(int[] negatives, int[] sortedArray, int[] positives) {
    System.arraycopy(negatives, 0, sortedArray, 0, negatives.length);
    System.arraycopy(positives, 0, sortedArray, negatives.length, positives.length);
}
```
These methods are well-separated and clear, ensuring the implementation is easy to understand.

However, if you feel it would be better to remove them, that's also acceptable.

vil02: Basically this approach is based on the fact that all of the negative numbers are smaller than non-negative.

I would keep it simple and implement it only for non-negative numbers.

alxkm: @vil02 reverted

alxkm:
Hello @vil02, all issues for this PR already fixed (also including negative numbers logic remove), could you please review this PR. Thanks.

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -1,62 +1,97 @@
...
+        copyOutput(array, output);
     }
 
-    private static void radixsort(int[] arr, int n) {
-        int m = getMax(arr, n);
+    private static void countDigits(int[] array, int exp, int[] count) {
+        for (int i = 0; i < array.length; i++) {
+            count[getDigit(array[i], exp)]++;
+        }
+    }
```
vil02, line 74:
```suggestion
    private static void countDigits(int[] array, int exp) {
        int[] count = new int[BASE];
        for (int i = 0; i < array.length; i++) {
            count[getDigit(array[i], exp)]++;
        }
        return count;
    }
```

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -1,62 +1,97 @@
...
 
-    static void print(int[] arr, int n) {
-        for (int i = 0; i < n; i++) {
-            System.out.print(arr[i] + " ");
+    private static void buildOutput(int[] array, int exp, int[] output, int[] count) {
+        for (int i = array.length - 1; i >= 0; i--) {
+            int digit = getDigit(array[i], exp);
+            output[count[digit] - 1] = array[i];
+            count[digit]--;
         }
```
vil02, line 91:
```suggestion
    private static void buildOutput(int[] array, int exp, int[] count) {
        int[] output = new int[array.length];
        for (int i = array.length - 1; i >= 0; i--) {
            int digit = getDigit(array[i], exp);
            output[count[digit] - 1] = array[i];
            count[digit]--;
        }
        return output;
```

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -1,62 +1,97 @@
...
     }
 
-    private static int getMax(int[] arr, int n) {
-        int mx = arr[0];
-        for (int i = 1; i < n; i++) {
-            if (arr[i] > mx) {
-                mx = arr[i];
-            }
+    /**
+     * Sorts an array of positive integers using the radix sort algorithm.
```
vil02, line 17:
```suggestion
     * Sorts an array of nonnegative integers using the radix sort algorithm.
```

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -1,62 +1,97 @@
...
-        }
+    /**
+     * A utility method to perform counting sort of array[] according to the digit represented by exp.
+     *
+     * @param array the array to be sorted
+     * @param exp   the exponent representing the current digit position
+     */
+    private static void countingSortByDigit(int[] array, int exp) {
+        int[] output = new int[array.length];
+        int[] count = new int[BASE];
```
vil02, line 62:
```suggestion
```

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -1,62 +1,97 @@
...
 
-        System.arraycopy(output, 0, arr, 0, n);
+        countDigits(array, exp, count);
```
vil02, line 64:
```suggestion
        int[] count = countDigits(array, exp);
```

src/main/java/com/thealgorithms/sorts/RadixSort.java
```diff
@@ -1,62 +1,97 @@
...
+        accumulateCounts(count);
+        buildOutput(array, exp, output, count);
```
vil02, line 66:
```suggestion
        int[] output = buildOutput(array, exp, count);
```

vil02 (changes requested):

src/main/java/com/thealgorithms/sorts/RadixSort.java
alxkm: Thanks, this is a more accurate description.

PR #5285, opened 2024-07-12
sleep sort

aishuarya:


src/main/java/com/thealgorithms/sorts/SleepSort.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+
+final class SleepSort {
+
+    private SleepSort() {
+    }
+
+    public static void sleepSort(ArrayList<Integer> array) {
```
vil02, line 25: What do you think about changing the signature to
```suggestion
    public static ArrayList<Integer> sleepSort(ArrayList<Integer> array) {
```
or
```suggestion
    public static int[] sleepSort(int[] array) {
```
This would allow to update this method in such a way, that it will not print anything to the console (and simply testing).

src/main/java/com/thealgorithms/sorts/SleepSort.java
```diff
@@ -0,0 +1,59 @@
...
+        ArrayList<Thread> threads = new ArrayList<>();
+
+        for (int numbers : array) {
+            Thread thread = new Thread(() -> {
+                try {
+                    Thread.sleep(numbers); // Sleep for 'num' milliseconds
```
vil02, line 31: Do you think it is a good idea to throw `IllegalArgumentException`, when any of the entries is negative?

src/main/java/com/thealgorithms/sorts/SleepSort.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public static void main(String[] args) {
+        // Example usage
+        ArrayList<Integer> numbers = new ArrayList<>();
+        Collections.addAll(numbers, 15, 23, 8, 41, 30);
+
+        System.out.println("Sorting numbers using Sleep Sort:");
+        sleepSort(numbers);
+    }
```
vil02, line 58:
```suggestion
```

vil02 (changes requested):

aishuarya:
@vil02 Addressed you comments ! Thanks .

src/main/java/com/thealgorithms/sorts/SleepSort.java
```diff
@@ -0,0 +1,59 @@
...
+
+final class SleepSort {
+
+    private SleepSort() {
+    }
+
+    public static void sleepSort(ArrayList<Integer> array) {
+        ArrayList<Thread> threads = new ArrayList<>();
+
+        for (int numbers : array) {
```
alxkm: If `numbers` represent only one number, better to rename it to `number`

src/main/java/com/thealgorithms/sorts/SleepSort.java
```diff
@@ -0,0 +1,68 @@
...
+    public static List<Integer> sleepSort(List<Integer> array) {
+
+        // List to collect sorted elements
+        List<Integer> sortedList = Collections.synchronizedList(new ArrayList<>());
+        List<Thread> threads = new ArrayList<>();
+
+        for (int number : array) {
+            if (number < 0) {
+                throw new IllegalArgumentException("All numbers must be non-negative. Found: " + number);
+            }
```
vil02, line 39: This should happen, before starting any of the new threads.

src/main/java/com/thealgorithms/sorts/SleepSort.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param array the list of integers to sort
+     * @return the sorted list of integers
+     */
+    public static List<Integer> sleepSort(List<Integer> array) {
+
+        // List to collect sorted elements
+        List<Integer> sortedList = Collections.synchronizedList(new ArrayList<>());
+        List<Thread> threads = new ArrayList<>();
+
+        for (int number : array) {
```
vil02, line 36:
```suggestion
        for (final int number : array) {
```

vil02 (changes requested):
These are just some _obvious_ remarks. I will provide more feedback, when I will have a bit more time.

aishuarya:
@vil02 Addressed you comments  .

PR #5288, opened 2024-07-13
feat: add `PatienceSort`

alxkm:
Implementation for PatienceSort

Adding standard test for it

Update DIRECTORY.md


alxkm:
This PR provide implementation for this [issue](https://github.com/TheAlgorithms/Java/issues/5291)

src/main/java/com/thealgorithms/sorts/PatienceSort.java
```diff
@@ -0,0 +1,114 @@
...
+     * @return a list of piles
+     */
+    private static <T extends Comparable<T>> List<List<T>> formPiles(final T[] array) {
+        final List<List<T>> piles = new ArrayList<>();
+        final List<T> lastElements = new ArrayList<>();
+
+        for (T x : array) {
+            int pos = Collections.binarySearch(lastElements, x);
+            if (pos < 0) {
+                pos = ~pos; // bitwise complement, which is equivalent to (-pos - 1)
```
vil02, line 50: Why not just simply:
```suggestion
                pos = -pos - 1;
```
`pos = ~pos` _assumes_ how the integers are represented etc. With `pos = -pos - 1` I immediately see that if `pos < 0`, then `pos` becomes non-negative.

src/main/java/com/thealgorithms/sorts/PatienceSort.java
```diff
@@ -0,0 +1,114 @@
...
+     * prioritized.
+     *
+     * @param piles the list of piles to be merged
+     * @param <T> the type of elements in the piles, must be comparable
+     * @return a priority queue containing the top element of each pile
+     */
+    private static <T extends Comparable<T>> PriorityQueue<PileNode<T>> mergePiles(final List<List<T>> piles) {
+        PriorityQueue<PileNode<T>> pq = new PriorityQueue<>();
+        for (List<T> pile : piles) {
+            if (!pile.isEmpty()) {
```
vil02, line 78: Can it actually happen that the `pile` is empty here?

vil02 (changes requested):

src/main/java/com/thealgorithms/sorts/PatienceSort.java
alxkm: You are right. Fixed. But this was added because, this is more compact. But for educational point of view this provides more complexity.

src/main/java/com/thealgorithms/sorts/PatienceSort.java
alxkm: This is was like additional checking. But tried to find such case, but didn't find. So removed.

vil02 (approved):

PR #5310, opened 2024-08-03
Update GrahamScan.java

bayramtturgutt:
improved the Javadoc comments, clarified some methods in the Point class, and corrected some text.


src/main/java/com/thealgorithms/geometry/GrahamScan.java
```diff
@@ -20,17 +19,12 @@ public class GrahamScan {
...
-         */
+        // Pre-process points: sort by y-coordinate, then by polar order with respect to the first point
         Arrays.sort(points);
         Arrays.sort(points, 1, points.length, points[0].polarOrder());
         hull.push(points[0]);
 
-        // find index of first point not equal to a[0] (indexPoint1) and the first point that's not
-        // collinear with either (indexPoint2).
+        // Find the first point not equal to points[0] and the first point not collinear with the previous points
         int indexPoint1;
```
alxkm, line 28: This name does not provide any information about this variable. 

Maybe this one firstNonEqualIndex will give more understanding and clearity. 

And firstNonCollinearIndex instead of indexPoint2.

src/main/java/com/thealgorithms/geometry/GrahamScan.java
```diff
@@ -61,14 +55,14 @@ public GrahamScan(Point[] points) {
     }
 
     /**
-     * @return A stack of points representing the convex hull.
+     * @return An iterable collection of points representing the convex hull.
      */
     public Iterable<Point> hull() {
-        Stack<Point> s = new Stack<>();
+        Stack<Point> result = new Stack<>();
```
alxkm, line 61: Is it important use here Stack? If we replace it via LinkedList for example, than we will have only:
`return new LinkedList<>(hull);`

src/main/java/com/thealgorithms/geometry/GrahamScan.java
```diff
@@ -20,17 +19,12 @@ public class GrahamScan {
     private final Stack<Point> hull = new Stack<>();
 
     public GrahamScan(Point[] points) {
```
alxkm, line 21: It seems that this implementation does not validate the input data, that the minimum number of points must be 3 to construct a convex hull. And also if all the points are the same, then it also would not be able to process such input. 

Checks can be look something like this:

```
if (points == null || points.length < 3) {
    throw new IllegalArgumentException("At least 3 points are required to compute the convex hull");
}

// Check if all points are the same
boolean allSame = true;
for (int i = 1; i < points.length; i++) {
    if (!points[i].equals(points[0])) {
        allSame = false;
        break;
    }
}

if (allSame) {
    hull.push(points[0]);
    return;
}
```

It is also good to have appropriate tests for such cases.

alxkm (changes requested):

src/main/java/com/thealgorithms/geometry/GrahamScan.java
bayramtturgutt: added 👍

src/main/java/com/thealgorithms/geometry/GrahamScan.java
bayramtturgutt: rollbacked 👍

src/main/java/com/thealgorithms/geometry/GrahamScan.java
bayramtturgutt: added 👍

src/main/java/com/thealgorithms/geometry/GrahamScan.java
vil02: @alxkm what is wrong in computing the convex hull of two points, one point or even zero points? In such case (if the points are in the _general position_) the output should be the same as the input.

If some of the input points are _the same_, they are simply _collinear_, and the algorithms should handle that.

alxkm: @vil02, I thought that the convex hull (CH) of two points is problematic because a CH requires at least three non-collinear points to form a valid polygon, and CH is defined as the smallest convex polygon that can enclose a set of points. With only two points, you can't form a polygon, you can only form a line segment.

But there are two points of view. From mathematical perspective, and a standard expectations from CH is. 

From mathematical perspective all these cases (0, 1, 2) are trivial, and return just themselves. But I left comment from result perspective that we expect that algorithm just forming a polygon.

Thanks for clarification. I understand that we support mathematical point of view here (from logical perspective it is only right).

alxkm: @bayramtturgutt sorry for misunderstanding, but this suggestion for checking that points are the same, and input expectations for 3 points are wrong.  

Could you please revert these changes.

bayramtturgutt: No worries, I'll revert those changes :)

alxkm (approved):

src/main/java/com/thealgorithms/geometry/GrahamScan.java
```diff
@@ -61,7 +55,7 @@ public GrahamScan(Point[] points) {
     }
 
     /**
-     * @return A stack of points representing the convex hull.
+     * @return An iterable collection of points representing the convex hull.
      */
     public Iterable<Point> hull() {
         Stack<Point> s = new Stack<>();
```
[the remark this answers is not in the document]
bayramtturgutt: I'll update as you advised 👍

src/main/java/com/thealgorithms/geometry/GrahamScan.java
```diff
@@ -20,17 +19,12 @@ public class GrahamScan {
...
-         */
+        // Pre-process points: sort by y-coordinate, then by polar order with respect to the first point
         Arrays.sort(points);
         Arrays.sort(points, 1, points.length, points[0].polarOrder());
         hull.push(points[0]);
 
-        // find index of first point not equal to a[0] (indexPoint1) and the first point that's not
-        // collinear with either (indexPoint2).
+        // Find the first point not equal to points[0]/firstNonEqualIndex and the first point not collinear firstNonCollinearIndex with the previous points
         int indexPoint1;
```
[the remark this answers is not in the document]
bayramtturgutt: you are right, it's my bad, I'll fix it, thx :)

alxkm (approved):
Thank you for your contribution.

PR #5314, opened 2024-08-10
refactor: cleanup `BucketSort` 

alxkm:
### Refactoring of BucketSort

Several improvements have been made:

1. **Comment Refinement**: Redundant comments were removed, and relevant descriptions were moved directly into method-level documentation.

2. **Main Method Removal**: The `main` method was removed as it was redundant. Testing can be effectively handled through the provided unit tests, or by writing a simple test case if needed.

3. **Additional Unit Tests**: New unit tests were added.

4. **Stream API Usage**: During refactoring Java Streams were introduced to simplify and modernize the code.


src/test/java/com/thealgorithms/sorts/BucketSortTest.java
```diff
@@ -7,42 +7,82 @@
 public class BucketSortTest {
```
siriak, line 7: Why does it not use the common test base class for sorts?

alxkm: Because in that case we would have to remove all this code and provide a new one, since the current solution only supports int values.

Such test class:

```
public class BucketSortTest extends SortingAlgorithmTest {
    @Override
    SortAlgorithm getSortAlgorithm() {
        return new BucketSort();
    }
}
```
Require from us that our BucketSort implement this method `public <T extends Comparable<T>> T[] sort(T[] array);`

I will try to fix this. Common approach is much better and provides better testing.

alxkm: In fact, everything turned out to be simpler.

siriak (approved):
Looks good, thanks!

PR #5048, opened 2024-02-14
Temperature conversion added to conversions - issue#5042

TharunTech:
- [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [x] This pull request is all my own work -- I have not plagiarized it.
- [x] All filenames are in PascalCase.
- [x] All functions and variable names follow Java naming conventions.
- [x] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.
- [x] All new code is formatted with `clang-format -i --style=file path/to/your/file.java`

PR #5315, opened 2024-08-10
feat: add temperature unit conversions

vil02:
This PR adds a _framework_ to work with unit conversions. I did not want to make the things too complicated so I only handle the case, when relation between units is _linear_.

Please review this very carefully, because probably many people would like to add more types of unit conversions in the future. Please pay special attention to:
- `UnitConversionsTest` - I did not want to make it too complicated. It is not perfect, because adding a new _quantity_ (like mass, length, time etc.) requiters adding 3 methods. But I think that it is still reasonable,
- `UnitConversions` - my idea is to have `static` members like `TEMPERATURE`, `LENGTH`, `MASS`, `TIME` etc. there. Feels a bit weird to me, but nothing else comes to my mind,
- `UnitsConverter.computeAllConversions` - probably it not the most efficient way, but there are not that many units in general. I guess it can be optimized in the _future_ without changing the API,
- is it fine to represent units as `String`s?

Please let's do not rush here.

This PR is partially inspired by #5048.


src/main/java/com/thealgorithms/conversions/AffineConverter.java
```diff
@@ -0,0 +1,23 @@
+package com.thealgorithms.conversions;
```
alxkm, line 1: Maybe it would be more clear and structured to have its own package for this (eg units) inside the conversions? And also number systems (eg numsystems)? Because these classes pretty special "class" of convertors in many other converters. And in the future, if someone wants to add something, it will be easier to find.

src/main/java/com/thealgorithms/conversions/UnitConversions.java
```diff
@@ -0,0 +1,14 @@
...
+import static java.util.Map.entry;
+
+import java.util.Map;
+import org.apache.commons.lang3.tuple.Pair;
+
+public final class UnitConversions {
+    private UnitConversions() {
+    }
+
+    public static final UnitsConverter TEMPERATURE = new UnitsConverter(Map.ofEntries(entry(Pair.of("Kelvin", "Celsius"), new AffineConverter(1.0, -273.15)), entry(Pair.of("Celsius", "Fahrenheit"), new AffineConverter(9.0 / 5.0, 32.0)),
```
alxkm, line 12: As you mentioned "is it fine to represent units as Strings?". 
What do you think about introduce enum here:

`public enum TemperatureUnit {
    KELVIN,
    CELSIUS,
    FAHRENHEIT,
    REAUMUR,
    DELISLE,
    RANKINE
}
` 
If you want to save correctness for naming have enums with specialized language description for each.
Something like this:

```
public enum TemperatureUnit {
    KELVIN("Kelvin");
    private final String unit;
    
    TemperatureUnit(String unit) {
        this.unit = unit;
    }
    
    public String getUnit() {
        return unit;
    }
}
```

Enums would provide type safety, enhance readability, and make it easier to manage different unit systems. And the listings should not overly complicated this framework.

src/main/java/com/thealgorithms/conversions/AffineConverter.java
```diff
@@ -0,0 +1,23 @@
...
+
+public final class AffineConverter {
+    final double slope;
```
alxkm, line 4: What do you think about making them `private`?

Or these values ​​should be used in the future for some cases, such as `package private`?

src/main/java/com/thealgorithms/conversions/UnitsConverter.java
```diff
@@ -0,0 +1,88 @@
...
+        conversions = computeAllConversions(basicConversions);
+        units = extractUnits(conversions);
+    }
+
+    public double convert(final String inputUnit, final String outputUnit, final double value) {
+        if (inputUnit.equals(outputUnit)) {
+            throw new IllegalArgumentException("inputUnit must be different from outputUnit.");
+        }
+        final var conversionKey = Pair.of(inputUnit, outputUnit);
+        return conversions.get(conversionKey).convert(value);
```
alxkm, line 82: Could it be that there is no converter here and we need to check for null here and throw an `IllegalArgumentException`?

src/main/java/com/thealgorithms/conversions/UnitsConverter.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+
+public final class UnitsConverter {
+    private final Map<Pair<String, String>, AffineConverter> conversions;
+    private final Set<String> units;
+
+    private static void putIfNeeded(Map<Pair<String, String>, AffineConverter> conversions, final String inputUnit, final String outputUnit, final AffineConverter converter) {
+        if (!inputUnit.equals(outputUnit)) {
+            final var key = Pair.of(inputUnit, outputUnit);
+            if (!conversions.containsKey(key)) {
```
alxkm, line 16: It looks like this can be simplified by:
`conversions.putIfAbsent(key, converter);`

src/main/java/com/thealgorithms/conversions/UnitsConverter.java
```diff
@@ -0,0 +1,88 @@
+package com.thealgorithms.conversions;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import org.apache.commons.lang3.tuple.Pair;
+
+public final class UnitsConverter {
+    private final Map<Pair<String, String>, AffineConverter> conversions;
```
alxkm, line 10: For me it is better to use `Pair<K, V>`, it is short and clear.
But if you want to achieve simplicity, what do you think to simplify Pair usage and introduce such class, for easier reading:

```
public final class ConversionKey {
    private final String inputUnit;
    private final String outputUnit;

    public ConversionKey(String inputUnit, String outputUnit) {
        this.inputUnit = inputUnit;
        this.outputUnit = outputUnit;
    }

    public String getInputUnit() {
        return inputUnit;
    }

    public String getOutputUnit() {
        return outputUnit;
    }

    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;
        ConversionKey that = (ConversionKey) o;
        return Objects.equals(inputUnit, that.inputUnit) &&
               Objects.equals(outputUnit, that.outputUnit);
    }

    @Override
    public int hashCode() {
        return Objects.hash(inputUnit, outputUnit);
    }

    @Override
    public String toString() {
        return "ConversionKey{" +
                "inputUnit='" + inputUnit + '\'' +
                ", outputUnit='" + outputUnit + '\'' +
                '}';
    }
}
```

alxkm (changes requested):

src/main/java/com/thealgorithms/conversions/AffineConverter.java
vil02: Ups. You are right.

src/main/java/com/thealgorithms/conversions/UnitConversions.java
vil02: This will be cumbersome: think about other units like meters, inches, etc.

src/main/java/com/thealgorithms/conversions/UnitsConverter.java
vil02: All of the classes like `AffineConverter`, `UnitsConverter` are not really meant to be used by anyone.

src/main/java/com/thealgorithms/conversions/UnitsConverter.java
vil02: Not really: all of the converters are created _by us_ in `UnitConversions`.

src/main/java/com/thealgorithms/conversions/AffineConverter.java
siriak: @vil02 what do you think?

BamaCharanChhandogi (approved):
LGTM, thanks.

alxkm (approved):

PR #5337, opened 2024-08-16
refactor: `DecimalToHexadecimal`

alxkm:
Rename the previous file.
Improving the code to make it more readable.
Adding additional tests to this algorithm.


src/test/java/com/thealgorithms/conversions/DecimalToHexadecimalTest.java
```diff
@@ -0,0 +1,14 @@
+package com.thealgorithms.conversions;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class DecimalToHexadecimalTest {
+    @ParameterizedTest
+    @CsvSource({"0, 00000000", "1, 00000001", "10, 0000000a", "15, 0000000f", "16, 00000010", "255, 000000ff", "190, 000000be", "1800, 00000708"})
```
siriak, line 10: Why does it return leading zeroes?

alxkm: Such approach was before. The leading zeros in the output occur because the method adds the hexadecimal result string to always represent an 8-character string, corresponding to the full 32-bit integer size in hexadecimal form. This happens because the loop runs for `SIZE_OF_INT_IN_HALF_BYTES `iterations (8), even when the actual value needs fewer digits. 

So if this refactoring is important we can try to, remove leading zeros after. 
Or modify code to skip leading zeros (but initial code will look overcomplicated, because it is not related to converting).

siriak: I think the simplest way would be to trim leading 0s just before returning the result because I don't see a way to short-circuit the computation

alxkm: Ok, thanks fixed using such approach.

siriak (approved):

PR #5370, opened 2024-08-23
Traveling Salesman Problem Added

DBasu2610:
This is the Traveling Salesman Problem which gives the shortest possible route among a set of cities where every city is visited atleast once and the starting city is returned to

It is done by using the Held-Karp algo, with memorization

This approach is much more efficient than the brute-force approach for small to moderately sized instances of TSP.


alxkm:
Hi @DBasu2610 , build job does not pass. It has some checkstyle issues. Please fix them and we can start the review

PR #5383, opened 2024-08-24
refactor: `RootPrecision`

alxkm:
Class cleanup.
Readability improvements.
Added tests.


src/main/java/com/thealgorithms/others/RootPrecision.java
```diff
@@ -1,38 +1,21 @@
...
-        rv = (int) root;
-        return rv / precision;
+    /**
+     * Calculates the square root of a number with the specified precision.
+     *
+     * @param number    The number to calculate the square root of.
+     * @param precision The number of decimal places of precision.
+     * @return The square root of the number with the specified precision.
+     */
+    public static double calculateSquareRoot(int number, int precision) {
```
siriak, line 14: Is it even worth leaving this algorithm? It doesn't do much by itself as it relies on Math.sqrt. I'd expect it to get a square root without Math.sqrt (or Math.pow).

alxkm: Yes, from my perspective it is very strange 'algorithm'. And during refactoring, surprised that this is algorithm. Let's remove it. Because, as you said this is only wrapper under `Math.sqrt`

siriak (approved):

PR #5439, opened 2024-08-31
refactor: change List<Integer> to LinkedList<Integer> and remove 'n < 0'

xizhengl:
- [ ] change List<Integer> to LinkedList<Integer> and remove 'n < 0'

### DIFF
--- a/src/main/java/com/thealgorithms/backtracking/ArrayCombination.java
+++ b/src/main/java/com/thealgorithms/backtracking/ArrayCombination.java
@@ -1,6 +1,7 @@
 package com.thealgorithms.backtracking;
 
 import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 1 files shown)


src/main/java/com/thealgorithms/backtracking/ArrayCombination.java
```diff
@@ -18,16 +19,16 @@ private ArrayCombination() {
      * @return A list of all combinations of length k.
      */
     public static List<List<Integer>> combination(int n, int k) {
-        if (n < 0 || k < 0 || k > n) {
+        if (k < 0 || k > n) {
```
alxkm, line 22: The n < 0 check is important because it ensures that the input values are valid within the context of combinations, prevents undefined behavior.

src/main/java/com/thealgorithms/backtracking/ArrayCombination.java
```diff
@@ -18,16 +19,16 @@ private ArrayCombination() {
...
         }
 
         List<List<Integer>> combinations = new ArrayList<>();
-        combine(combinations, new ArrayList<>(), 0, n, k);
+        combine(combinations, new LinkedList<>(), 0, n, k);
         return combinations;
     }
 
-    private static void combine(List<List<Integer>> combinations, List<Integer> current, int start, int n, int k) {
+    private static void combine(List<List<Integer>> combinations, LinkedList<Integer> current, int start, int n, int k) {
```
alxkm, line 31: Using LinkedList instead of List in method parameters is considered bad practice as it provides tight coupling, and issues with extensibility.

xizhengl: List doesn't seem to have a removeLast method

alxkm: It has. In code below, you can see that it already used. And code compile and pass tests.

src/main/java/com/thealgorithms/backtracking/ArrayCombination.java
alxkm: It looks like if you remove this check everything will work correctly. And you're right, we can remove it. But this check provides readable constraints and with this check the code becomes more readable.

PR #5532, opened 2024-10-03
Add function documentation and parameterized tests to Sudoku.java

Hardvan:

### DIFF
--- a/src/main/java/com/thealgorithms/others/Sudoku.java
+++ b/src/main/java/com/thealgorithms/others/Sudoku.java
@@ -1,33 +1,44 @@
 package com.thealgorithms.others;
 
+/**
+ * A class that provides methods to solve a 9x9 Sudoku puzzle using a backtracking approach.
+ * The Sudoku board is represented as a 2D array, and the methods are designed to 
+ * check for safe placements of numbers, solve the puzzle recursively, and print the board.
+ */
 final class Sudoku {
+
     private Sudoku() {
     }
 
+    /**
+     * Checks if placing a number in a specific position on the Sudoku board is safe.
+     * The number is considered safe if it does not violate any of the Sudoku rules:
+     * - It should not be present in the same row.
+     * - It should not be present in the same column.
+     * - It should not be present in the corresponding 3x3 subgrid.
+     *
+     * @param board The current state of the Sudoku board.
+     * @param row   The row index where the number is to be placed.
+     * @param col   The column index where the number is to be placed.
+     * @param num   The number to be placed on the board.
+     * @return True if the placement is safe, false otherwise.
+     */
     public static boolean isSafe(int[][] board, int row, int col, int num) {
-        // Row has the unique (row-clash)
+        // Check the row for duplicates
         for (int d = 0; d < board.length; d++) {
-            // Check if the number we are trying to
-            // place is already present in
-            // that row, return false;
             if (board[row][d] == num) {
                 return false;
             }
         }
 
-        // Column has the unique numbers (column-clash)
+        // Check the column for duplicates
         for (int r = 0; r < board.length; r++) {
-            // Check if the number
-            // we are trying to
-            // place is already present in
-            // that column, return false;
             if (board[r][col] == num) {
                 return false;
             }
         }
 
-        // Corresponding square has
-        // unique number (box-clash)
+        // Check the corresponding 3x3 subgrid for duplicates
         int sqrt = (int) Math.sqrt(board.length);
         int boxRowStart = row - row % sqrt;
         int boxColStart = col - col % sqrt;
@@ -40,22 +51,31 @@ final class Sudoku {
             }
         }
 
-        // if there is no clash, it's safe
         return true;
     }
 
+    /**
+     * Solves the Sudoku puzzle using backtracking.
+     * The algorithm finds an empty cell and tries placing numbers from 1 to 9.
+     * If a number placement is valid (checked via `isSafe`), the number is placed and the function
+     * recursively attempts to solve the rest of the puzzle.
+     * If no solution is possible, the number is removed (backtracked), and the process is repeated.
+     *
+     * @param board The current state of the Sudoku board.
+     * @param n     The size of the Sudoku board (typically 9 for a standard puzzle).
+     * @return True if the Sudoku puzzle is solvable, false otherwise.
+     */
     public static boolean solveSudoku(int[][] board, int n) {
         int row = -1;
         int col = -1;
         boolean isEmpty = true;
+
+        // Find the next empty cell (marked by 0)
         for (int i = 0; i < n; i++) {
             for (int j = 0; j < n; j++) {
                 if (board[i][j] == 0) {
                     row = i;
                     col = j;
-
-                    // We still have some remaining
-                    // missing values in Sudoku
                     isEmpty = false;
                     break;
                 }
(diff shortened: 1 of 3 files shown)


src/main/java/com/thealgorithms/others/Sudoku.java
```diff
@@ -1,33 +1,44 @@
 package com.thealgorithms.others;
 
+/**
+ * A class that provides methods to solve a 9x9 Sudoku puzzle using a backtracking approach.
```
alxkm, line 4: Does this code only work with 9x9 sizes?

src/main/java/com/thealgorithms/others/Sudoku.java
```diff
@@ -1,33 +1,44 @@
...
+
     private Sudoku() {
     }
 
+    /**
+     * Checks if placing a number in a specific position on the Sudoku board is safe.
+     * The number is considered safe if it does not violate any of the Sudoku rules:
+     * - It should not be present in the same row.
+     * - It should not be present in the same column.
+     * - It should not be present in the corresponding 3x3 subgrid.
```
alxkm, line 18: Is this method only applicable for 3x3 subgrid, or need to change this?

src/main/java/com/thealgorithms/others/Sudoku.java
```diff
@@ -40,22 +51,31 @@ public static boolean isSafe(int[][] board, int row, int col, int num) {
             }
         }
 
-        // if there is no clash, it's safe
         return true;
     }
 
+    /**
+     * Solves the Sudoku puzzle using backtracking.
+     * The algorithm finds an empty cell and tries placing numbers from 1 to 9.
```
alxkm, line 59: Maybe we should also point out here that 1...9 is standard, but it could be for 1...n too

alxkm (changes requested):

src/main/java/com/thealgorithms/others/Sudoku.java
```diff
@@ -86,22 +105,36 @@ public static boolean solveSudoku(int[][] board, int n) {
...
+    /**
+     * Prints the current state of the Sudoku board in a readable format.
+     * Each row is printed on a new line, with numbers separated by spaces.
+     *
+     * @param board The current state of the Sudoku board.
+     * @param n     The size of the Sudoku board (typically 9 for a standard puzzle).
+     */
     public static void print(int[][] board, int n) {
-        // We got the answer, just print it
+        // Print the board in a 9x9 grid format
```
alxkm, line 116: Same question here also.

alxkm (changes requested):

src/main/java/com/thealgorithms/others/Sudoku.java
```diff
@@ -70,12 +90,11 @@ public static boolean solveSudoku(int[][] board, int n) {
             return true;
         }
 
-        // Else for each-row backtrack
+        // Try placing numbers 1 to 9 in the empty cell
```
alxkm, line 93: And here also

alxkm (changes requested):

src/main/java/com/thealgorithms/others/Sudoku.java
Hardvan: Yes, it works only with 9x9 grids, as that is the standard variation of the Sudoku puzzle.

src/main/java/com/thealgorithms/others/Sudoku.java
Hardvan: The size of the subgrid is based on the length of the board (`sqrt = (int) Math.sqrt(board.length)`). For a standard 9x9 Sudoku, the subgrid size will be 3 (since `Math.sqrt(9) == 3`), meaning the method will check within a 3x3 subgrid. However, if the board size is different, for example, a 16x16 Sudoku, the subgrid size would be 4 (`Math.sqrt(16) == 4`), and it would check within a 4x4 subgrid.

src/main/java/com/thealgorithms/others/Sudoku.java
Hardvan: Okay, I added the following comment to this function documentation:
`The standard version of Sudoku uses numbers from 1 to 9, so the algorithm can be easily modified for other variations of the game.`

src/main/java/com/thealgorithms/others/Sudoku.java
Hardvan: Changed this comment to `Print the board in a nxn grid format`

src/main/java/com/thealgorithms/others/Sudoku.java
Hardvan: Changed this comment to: `// Try placing numbers 1 to n in the empty cell (typically n=9)`

src/main/java/com/thealgorithms/others/Sudoku.java
alxkm: So the changed comment should be look like this:

`It should not be present in the corresponding subgrid, which is sqrt(n) x sqrt(n) in size (e.g., for a 9x9 grid, the subgrid will be 3x3).`

src/main/java/com/thealgorithms/others/Sudoku.java
alxkm: While this code is commonly used for solving the standard 9x9 Sudoku grid, it is not limited to that size. The algorithm is recursive and works for any n x n grid, as long as n is a perfect square. This is because the algorithm relies on dividing the grid into square subgrids, which are sqrt(n) x sqrt(n) in size (e.g., 3x3 subgrids for a 9x9 puzzle, or 4x4 subgrids for a 16x16 puzzle).

And in this documentation, it’s essential to specify the grid size. For example, for a 9x9 grid, the subgrids will be 3x3. For a 16x16 grid, the subgrids will be 4x4. This ensures clarity about how the algorithm handles different board dimensions and prevents confusion regarding compatibility with other grid sizes.

And will be good to highlight like this: 

`A class that provides methods to solve Sudoku puzzles of any n x n size using a backtracking approach, where n must be a perfect square. The algorithm checks for safe number placements in rows, columns, and subgrids (which are sqrt(n) x sqrt(n) in size) and recursively solves the puzzle. Though commonly used for 9x9 grids, it is adaptable to other valid Sudoku dimensions.`

src/main/java/com/thealgorithms/others/Sudoku.java
alxkm: And we need to have this here:

`Solves the Sudoku puzzle using backtracking. The algorithm finds an empty cell and tries placing numbers from 1 to n, where n is the size of the board (for example, from 1 to 9 in a standard 9x9 Sudoku).`

alxkm (changes requested):

src/main/java/com/thealgorithms/others/Sudoku.java
Hardvan: Changed the class documentation to reflect the above insight.

src/main/java/com/thealgorithms/others/Sudoku.java
Hardvan: Added the above line as a bullet point in the function documentation.

src/main/java/com/thealgorithms/others/Sudoku.java
Hardvan: Changed the function docstring to include the above statement.

Hardvan:
I have changed the class and function documentations to reflect the aforementioned suggested changes.

alxkm (approved):

Issue #5529, opened 2024-10-02
Binary Search Algorithm in Java


Sameeratweb:
### What would you like to Propose?

An Java Program performing Binary Search Operation

### Issue details

Adding An program that performs Binary search using Java

purnimasharma06:
Hi, I'd like to contribute by adding a Java program that performs a binary search operation. Could you please assign this task to me?

Thank you!

AnushkJain2201:
Hey!! I will contribute by adding this algorithm. Please assign me this issue.

Thank You!!

YoosufAathil:
Hi !!!, I would like to work on this, could you please assign me on this.

Thank You in Advance

tejaswi0910:
Hey @Sameeratweb 
I think it is already implemented [here](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/searches/BinarySearch.java)

shubh-dev1:
hey, is the repo still open for the hacktoberfest contribution

PR #5542, opened 2024-10-03
Added Solution for Binary Search in Java (issue #5529)

Sidhesh-Goud:
### Enhancements:
_Overflow Prevention:_ The mid-point calculation is optimized to avoid overflow when dealing with large arrays. Instead of using (left + right) / 2, we use left + (right - left) / 2.

_Edge Case Handling:_ The algorithm checks if the input array is null or empty and handles it gracefully by returning -1 early.

_Debugging Statements:_ Added debug statements to show the mid index and mid element during the execution, useful for understanding how the algorithm works step-by-step.

_Readability:_ The code is organized with clear comments explaining each step, which is helpful when understanding or modifying the algorithm.

_Time Complexity:_
Best case: O(1) (if the target is at the middle)
Average and worst case: O(logn), where n is the number of elements in the array.

### DIFF
--- a/src/test/java/com/thealgorithms/searches/BinarySearch.java
+++ b/src/test/java/com/thealgorithms/searches/BinarySearch.java
@@ -1,7 +1,10 @@
 package com.thealgorithms.searches;
 
 class BinarySearch {
-
+    // Private constructor to prevent instantiation
+    private BinarySearch() {
+        throw new UnsupportedOperationException("Utility class");
+    }
     // Enhanced binary search method
     public static int binarySearch(int[] arr, int target) {
         // Edge case: If the array is empty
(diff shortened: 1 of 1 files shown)


alxkm:
There are already a bunch of different binary searches in the repository, what is the difference between these implementations?

https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/searches/BinarySearch.java
https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/searches/IterativeBinarySearch.java
https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/searches/PerfectBinarySearch.java

PR #5552, opened 2024-10-03
Add `PrefixToInfix.java` new algorithm

Hardvan:


src/main/java/com/thealgorithms/stacks/PrefixToInfix.java
```diff
@@ -0,0 +1,68 @@
...
+
+                // Push the resulting infix expression back onto the stack
+                stack.push(infix);
+            } else {
+                // Push operand onto stack
+                stack.push(Character.toString(token));
+            }
+        }
+
+        return stack.pop(); // final element on the stack is the full infix expression
```
alxkm, line 66: As far as I understand, there should only be one element in the stack. Can you imagine a situation where the final stack will not have any elements, which would mean that the expression is not well formed?

Do you think it is worth adding such a constraint before return:

```
if (stack.size() != 1) {
    throw new ArithmeticException("Malformed prefix expression");
}
```

alxkm (changes requested):

src/main/java/com/thealgorithms/stacks/PrefixToInfix.java
```diff
@@ -0,0 +1,68 @@
+package com.thealgorithms.stacks;
+
+import java.util.Stack;
+
+/**
+ * Prefix to Infix implementation via Stack
```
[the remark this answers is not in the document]
Hardvan: This documentation is much clearer & concise! I have implemented this change now.

src/main/java/com/thealgorithms/stacks/PrefixToInfix.java
Hardvan: Yes, such an edge case is possible.
Example: For the input `+ A B C`, the final stack contents will be `["C", "(A + B)"]`, indicating `stack.size() is > 1` and an erroneous prefix expression.

I have added the suggested stack size check in the code, as well as an additional test for this:

```java
@Test
void testMalformedPrefixExpression() {
    assertThrows(ArithmeticException.class, () -> PrefixToInfix.getPrefixToInfix("+ABC"));
}
```

Hardvan:
I have implemented the suggested changes, and added the test for `stack size != 1` before the return statement.

alxkm (approved):

alxkm (approved):

PR #5639, opened 2024-10-08
Add boundary traversal of binary tree

xuyang471:
This PR implements the boundary traversal algorithm for a binary tree as requested in the related issue. The boundary traversal is performed in three parts:

Left Boundary: Traverses the left boundary of the tree, excluding leaf nodes.
Leaf Nodes: Collects all leaf nodes from left to right.
Right Boundary: Traverses the right boundary of the tree, excluding leaf nodes, and adds the nodes in reverse order to ensure correct boundary order.
Both time complexity and space complexity of this implementation are O(n), where n is the number of nodes in the binary tree.

The boundary traversal is performed both iteratively and recursively, and edge cases such as single-node trees and trees without left or right subtrees are handled appropriately.
All methods in the test class passed the test.


siriak (approved):
Looks good, thanks!

PR #5651, opened 2024-10-08
Optimised Space Complexity To O(sum)

### LINKED ISSUE #5600: [FEATURE REQUEST] Optimized approach to Subset Sum problem

I would like to propose an space optimized solution to the Subset Sum problem.

### Issue details

The existing subset sum problem solution's space complexity is `O(n*sum)`. 
By optimizing the algorithm/ approach further, we can achieve the space complexity of `O(sum)` which is more efficient than the existing one.

Algorithm link: 
https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/dynamicprogramming/SubsetSum.java

### Additional Information

_No response_

sankethl27:
fixes  #5600

### Problem:
This PR implements a solution to the "Subset Sum Equals to K" problem using dynamic programming while optimizing space complexity from O(n * sum) to O(sum). The problem is to determine whether there is a subset of a given array whose sum is exactly equal to a target integer `K`.

### Approach:
The standard DP approach for the subset sum problem involves creating a 2D table where each entry dp[i][j] represents whether a subset of the first `i` elements of the array can sum to `j`. This has a space complexity of O(n * sum), where `n` is the number of elements in the array and `sum` is the target sum.

To optimize space complexity, we can use a 1D array of size `sum + 1` instead of the 2D table. The key idea is that the current state of the DP table only depends on the previous row (i.e., whether the sum was possible with or without the current element). Thus, we can update the DP array in reverse order, ensuring that we do not overwrite values that are still needed.

### Space Optimization:
The space-optimized approach reduces the space complexity to O(sum) by reusing a 1D DP array. At each iteration, we update the DP array from right to left to ensure we only use the previous values from the last iteration.

### DIFF
--- a/src/main/java/com/thealgorithms/dynamicprogramming/SubsetSum.java
+++ b/src/main/java/com/thealgorithms/dynamicprogramming/SubsetSum.java
@@ -13,24 +13,27 @@ public final class SubsetSum {
      */
     public static boolean subsetSum(int[] arr, int sum) {
         int n = arr.length;
-        boolean[][] isSum = new boolean[n + 1][sum + 1];
 
-        // Initialize the first column to true since a sum of 0 can always be achieved with an empty subset.
-        for (int i = 0; i <= n; i++) {
-            isSum[i][0] = true;
-        }
+        //Intialize Two Arrays to store current and prev states
+        boolean[] isSumCurr = new boolean[sum + 1];
+        boolean[] isSumPrev = new boolean[sum+1];
+
+        // Mark prev[0] = true as it is true to make sum = 0
+        // using 0 elements
+        isSumPrev[0] = true;
 
         // Fill the subset sum matrix
         for (int i = 1; i <= n; i++) {
-            for (int j = 1; j <= sum; j++) {
+            for (int j = 0; j <= sum; j++) {
                 if (arr[i - 1] <= j) {
-                    isSum[i][j] = isSum[i - 1][j] || isSum[i - 1][j - arr[i - 1]];
+                    isSumCurr[j] = isSumPrev[j] || isSumPrev[j - arr[i - 1]];
                 } else {
-                    isSum[i][j] = isSum[i - 1][j];
+                    isSumCurr[j] = isSumPrev[j];
                 }
             }
+            isSumPrev = isSumCurr.clone();
         }
 
-        return isSum[n][sum];
+        return isSumPrev[sum];
     }
 }


src/main/java/com/thealgorithms/dynamicprogramming/SubsetSum.java
```diff
@@ -9,28 +9,31 @@ private SubsetSum() {
...
+            for (int j = 0; j <= sum; j++) {
                 if (arr[i - 1] <= j) {
-                    isSum[i][j] = isSum[i - 1][j] || isSum[i - 1][j - arr[i - 1]];
+                    isSumCurr[j] = isSumPrev[j] || isSumPrev[j - arr[i - 1]];
                 } else {
-                    isSum[i][j] = isSum[i - 1][j];
+                    isSumCurr[j] = isSumPrev[j];
                 }
             }
+            isSumPrev = isSumCurr.clone();
```
alxkm, line 34: The use of` isSumPrev = isSumCurr.clone();` inside the loop introduces unnecessary overhead. Cloning arrays in every iteration can be costly in terms of performance.

You can use two arrays and just swap their references, rather than cloning the array each time.

src/main/java/com/thealgorithms/dynamicprogramming/SubsetSum.java
```diff
@@ -9,28 +9,31 @@ private SubsetSum() {
...
         int n = arr.length;
-        boolean[][] isSum = new boolean[n + 1][sum + 1];
 
-        // Initialize the first column to true since a sum of 0 can always be achieved with an empty subset.
-        for (int i = 0; i <= n; i++) {
-            isSum[i][0] = true;
-        }
+        // Intialize Two Arrays to store current and prev states
+        boolean[] isSumCurr = new boolean[sum + 1];
+        boolean[] isSumPrev = new boolean[sum + 1];
```
alxkm, line 20: While you're already using a more efficient space approach (two 1D arrays), it can be improved further by using only one array. You don't actually need both `isSumCurr `and `isSumPrev`. 

By processing the array in reverse order (starting from sum and decrementing down to 0), you can reuse a single array.

alxkm (changes requested):

sankethl27:
Hey @alxkm Thanks for the feedback.
I have Optimised the code to use single dp Array and Also removed the line which was cloning the array.
Let me know if there is anything else i can do to optimise the code further.

alxkm (approved):
Thank you for your contribution.

PR #5700, opened 2024-10-10
Add Elliptic Curve Cryptography Algorithm

xuyang471:
In this PR, I’ve added the Elliptic Curve Cryptography (ECC) algorithm to the project. The ECC class handles key generation, encryption, and decryption using elliptic curve operations. I also included classes for defining elliptic curves and points on those curves, which are essential for performing ECC computations.

Elliptic Curve Cryptography offers several advantages, including stronger security with smaller key sizes compared to other cryptographic systems like RSA. This makes ECC highly efficient in terms of computational performance, memory usage, and bandwidth, which is especially beneficial for mobile devices and environments with limited resources.

Additionally, I’ve written several unit tests to verify the encryption and decryption processes, ensure randomness in the ciphertexts, and check functionality with known values. This implementation aims to provide a simple but effective introduction to ECC for cryptographic purposes.


siriak (approved):
Looks good, thanks!

PR #5720, opened 2024-10-11
Create StalinSort.java

anant-jain01:

### DIFF
--- a/src/main/java/com/thealgorithms/sorts/StalinSort.java
+++ b/src/main/java/com/thealgorithms/sorts/StalinSort.java
@@ -1,0 +1,39 @@
+package com.thealgorithms.sorts;
+
+/**
+ * Stalin Sort is a sorting algorithm that works by iterating through the array and 
+ * maintaining a sorted portion of the array at the beginning. If a number is encountered 
+ * that is smaller than the last number in the sorted portion, it is discarded (not added).
+ * This algorithm is not stable and will not sort all elements correctly if there are 
+ * elements that violate the sort order.
+ * 
+ * For more information, see: 
+ * https://en.wikipedia.org/wiki/Stalin_sort
+ */
+public class StalinSort implements SortAlgorithm {
+
+    @Override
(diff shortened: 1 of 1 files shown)


src/main/java/com/thealgorithms/sorts/StalinSort.java
```diff
@@ -0,0 +1,39 @@
...
+        for (int i = 1; i < array.length; i++) {
+            if (SortUtils.less(result[index - 1], array[i])) {
+                result[index++] = array[i];
+            }
+        }
+
+        return trimArray(result, index);
+    }
+
+    private <T> T[] trimArray(T[] array, int length) {
```
saahil-mahato, line 34: I think this is redundant. Is there a reason to use this?

saahil-mahato, line 34: This just adds unnecessary overhead. We can remove this if we just use ArrayList.
```
List<T> result = new ArrayList<>();
        result.add(array[0]); 

        for (int i = 1; i < array.length; i++) {
            if (SortUtils.less(result.get(result.size() - 1), array[i])) {
                result.add(array[i]);
            }
        }
```

src/main/java/com/thealgorithms/sorts/StalinSort.java
```diff
@@ -0,0 +1,39 @@
...
+ * elements that violate the sort order.
+ * 
+ * For more information, see: 
+ * https://en.wikipedia.org/wiki/Stalin_sort
+ */
+public class StalinSort implements SortAlgorithm {
+
+    @Override
+    public <T extends Comparable<T>> T[] sort(T[] array) {
+        if (array.length == 0) {
```
saahil-mahato, line 17: add `null` check here to prevent null pointer expection.

saahil-mahato:
Please add tests.

siriak (changes requested):
Please add JUnit tests

PR #5734, opened 2024-10-12
feat:Add SwapNumbersUsingXor algo with JUnit tests

Tanmay-Singh3004:


siriak:
It's not an algorithm, it's a trick. You could just return [a, b] from the method and get the same result, so the xor in this method is not useful to the caller

PR #5748, opened 2024-10-13
Add disk scheduling algorithms.

xuyang471:
I added five disk scheduling algorithms and test classes to the disk scheduling folder, and they adopted different strategies when simulating disk read and write requests. They are SSF, SCAN, CSCAN, LOOK and CLOOK algorithms.
SSF (shortest seek first) algorithm is a disk scheduling strategy, and its core idea is to always select the request closest to the current head position for processing. This method can effectively reduce the moving distance of the magnetic head, thus reducing the average seek time. Although SSF can significantly improve disk performance, it may lead to some requests far away from the current head position not being served for a long time, which is the so-called "hunger" problem. In order to alleviate this problem, other strategies or rules are sometimes combined in practical application to ensure that all requests can get reasonable responses.

SCAN algorithm is another disk scheduling technology, which divides the disk into multiple areas, and the head scans continuously from one end to the other until it reaches the edge of the disk and reverses the direction to continue scanning. When the head moves, it will serve all requests encountered on the way. SCAN algorithm avoids the hunger problem that SSF may cause, because it ensures that the requests in each area will eventually be processed. However, because the head has to scan the whole disk range back and forth, it needs to complete a complete round trip even if there is no pending request, which may lead to unnecessary head movement and increased waiting time.

CSCAN (Circular SCAN) algorithm is a variant of SCAN. Different from the traditional Scan, in CSCAN, once the head reaches one end of the disk, it will not scan in reverse immediately, but quickly return to the beginning end of the disk to start a new round of scanning. This method reduces the idle time of the magnetic head in the no-request interval and improves the efficiency. CSCAN also solves the problem of hunger, and it can provide more uniform service distribution than SCAN, because scanning is restarted from one end of the disk every time, instead of turning back halfway.

The LOOK algorithm is similar to SCAN, but there is a key difference: the LOOK algorithm only moves the head in the direction where there is a request for service. If the head reaches one end of the disk and there are no more requests in that direction, the head will immediately turn and jump directly to the nearest request at the other end, instead of continuing to SCAN the blank area like Scan. This optimization reduces unnecessary head movement, and makes LOOK algorithm more efficient than SCAN, especially in the case of uneven distribution of requests.

CLOOK algorithm combines the characteristics of CSCAN and LOOK. The magnetic head only moves in the direction where there is a request, and when it reaches one end of the disk, it will quickly jump back to the other end to continue processing the remaining requests. This not only retains the advantage of LOOK algorithm to reduce useless movement, but also inherits the cyclic characteristics of CSCAN's request processing order, which makes the overall disk access more balanced and can also avoid hunger. CLOOK is suitable for those application scenarios that require high response speed and fairness.


siriak (approved):
Looks good, thanks!

xuyang471:
> 看起来不错，谢谢！



> Looks good, thanks!

Thank you for receiving my Pull Request! I am very glad to contribute to the project.

Issue #5654, opened 2024-10-08
[FEATURE REQUEST] Add Moore's Voting Algorithm #5636


PriyamUttam-Dev:
### What would you like to Propose?

Moore's Voting Algorithm

Moore's Voting Algorithm is an efficient algorithm for finding the majority element in an array, where the majority element is defined as an element that appears more than half the time.

Key Idea:

The algorithm maintains two variables: candidate and count.
candidate stores the current candidate for the majority element.
count keeps track of the count of the current candidate.
Iterate through the array:
If count is 0, set candidate to the current element and increment count.
If the current element is equal to candidate, increment count.
Otherwise, decrement count.
After iterating through the array, check if candidate appears more than half the time. If so, it's the majority element; otherwise, there is no majority element.
Implementation in Java:

Java
import java.util.Arrays;

public class MajorityElement {
    public static int findMajorityElement(int[] nums) {
        int candidate = nums[0];
        int count = 1;

        for (int i = 1; i < nums.length;   
 i++) {
            if (count == 0) {
                candidate = nums[i];
                count = 1;
            } else if (nums[i] == candidate) {
                count++;
            } else {
                count--;   

            }
        }

        // Verify if candidate is the majority element
        count = 0;
        for (int num : nums) {
            if (num == candidate) {
                count++;
            }
        }

        if (count > nums.length / 2) {
            return candidate;
        } else {
            return   
 -1; // No majority element found
        }
    }

    public static void main(String[] args) {
        int[] nums   
 = {2, 1, 2, 1, 2, 2, 1};
        int majorityElement = findMajorityElement(nums);

        if (majorityElement != -1) {
            System.out.println("Majority element: " + majorityElement);
        } else {
            System.out.println("No majority element   
 found");   

        }
    }
}

Explanation:

The findMajorityElement method takes an integer array nums as input and returns the majority element or -1 if no majority element exists.
The algorithm follows the steps described above, maintaining candidate and count.
After iterating through the array, a second pass is performed to verify if candidate is indeed the majority element by counting its occurrences.
If count is greater than half the size of the array, candidate is the majority element.
Time Complexity:

The algorithm has a time complexity of O(n), where n is the size of the input array. This is because it involves two passes through the array.
Key Points:

Moore's Voting Algorithm is efficient and easy to implement.
It's applicable to finding the majority element in an array where the majority element appears more than half the time.
The algorithm can be adapted to other scenarios where a dominant element needs to be identified.

### Issue details

[...]

PR #5770, opened 2024-10-13
Feature: Moore's voting algorithm is added 

dhavalkgit:
Hi 

I have implemented Moore's Voting Algorithm to find the majority element in a list or array. The code includes the following improvements and features:
 
First Pass: first find the majority element using Moore's voting algorithm.

Second Pass Verification: After determining a candidate using Moore's Voting Algorithm, the code performs a second pass to verify if the element truly appears more than `n/2` times.

feature #5654


siriak:
The algorithm is the same for arrays and lists, so it's just duplication

dhavalkgit:
If you pass an array as an argument, then it converts into a list and is called an algorithm.

image
image

PR #5789, opened 2024-10-14
feat: Add `ConvexHull` new algorithm with Junit tests

Hardvan:


src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
```diff
@@ -0,0 +1,228 @@
+package com.thealgorithms.divideandconquer;
```
alxkm, line 1: It seems that this PR is closely related to the existing [GrahamScan implementation](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/geometry/GrahamScan.java).

The Point class (or record) from the GrahamScan class can be reused here, and it might be beneficial to review both implementations to identify reusable components. This would help avoid redundancy and improve maintainability. It might also be worth considering merging or unifying the logic in some way to make the code more efficient and consistent across both classes.

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
```diff
@@ -0,0 +1,228 @@
...
+    /**
+     * Brute-force algorithm to find the convex hull of a set of points.
+     * This algorithm checks every pair of points and determines if all other
+     * points lie to one side of the line formed by these two points.
+     *
+     * @param points A list of points for which to compute the convex hull.
+     * @return A sorted list of points forming the convex hull.
+     */
+    public static List<Point> convexHullBruteForce(List<Point> points) {
+        Collections.sort(points);
```
alxkm, line 120: The method sorts the points twice: once at the beginning, and again at the end after adding the points to convexSet. Since Set does not preserve order and this requires sorting again, using a TreeSet with a custom comparator can maintain sorted order while preventing duplicates. This removes the need for the second sort and improves efficiency.

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
```diff
@@ -0,0 +1,228 @@
...
+
+                for (int k = 0; k < points.size(); k++) {
+                    if (k != i && k != j) {
+                        double detK = det(points.get(i), points.get(j), points.get(k));
+
+                        if (detK > 0) {
+                            pointsLeftOfIJ = true;
+                        } else if (detK < 0) {
+                            pointsRightOfIJ = true;
+                        } else if (points.get(k).compareTo(points.get(i)) < 0 || points.get(k).compareTo(points.get(j)) > 0) {
```
alxkm, line 137: This can be extracted to the method

alxkm (changes requested):

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
Hardvan: Done!

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
Hardvan: I refactored the code to use `TreeSet` for the `convexSet` variable.

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
Hardvan: Done!

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
```diff
@@ -0,0 +1,229 @@
...
+
+            if (extremePoint != null) {
+                constructHull(candidatePoints, left, extremePoint, convexSet);
+                convexSet.add(extremePoint);
+                constructHull(candidatePoints, extremePoint, right, convexSet);
+            }
+        }
+    }
+
+    public static List<Point> grahamScan(List<Point> points) {
```
alxkm, line 183: It will be better to remove this and left it as it was. I thought just to keep your class and GrahamScan inside one package, because it is mostly related similar topic, and just to avoid code duplication use same Point class, because it is almost the same.

alxkm (changes requested):

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
```diff
@@ -0,0 +1,229 @@
+package com.thealgorithms.divideandconquer;
```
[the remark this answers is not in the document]
Hardvan: Yeah, that will be more convenient as well.

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
```diff
@@ -0,0 +1,229 @@
...
+ * 2. Recursive (Divide-and-Conquer) Method
+ * 3. Graham Scan method
+ *
+ * @author Hardvan
+ */
+public final class ConvexHull {
+    private ConvexHull() {
+    }
+
+    public static class Point implements Comparable<Point> {
```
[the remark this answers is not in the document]
Hardvan: I have added the common `Point.java` file for it and modified the Convex Hull and Graham Scan code.

src/main/java/com/thealgorithms/divideandconquer/ConvexHull.java
Hardvan: Done!

alxkm (approved):

PR #5802, opened 2024-10-14
Create ImplementStackUsingArray

singhshreya-bhriguvanshi:
A program to implement a Stack using Array. The push() method takes one argument, an integer 'x' to be pushed into the stack and pop() which returns an integer present at the top and popped out from the stack. If the stack is empty then return -1 from the pop() method.


siriak (changes requested):
Your file name doesn't end in .java, and you need to add JUnit tests

Nagpal-11 (commented):
Good way of approaching the problem.

varada610:
> Your file name doesn't end in .java, and you need to add JUnit tests

Yeah I added the same comment.

PR #5818, opened 2024-10-14
feat: Add `EnergyAwareScheduling` new algorithm with Junit tests

Hardvan:


src/main/java/com/thealgorithms/scheduling/EnergyAwareScheduling.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+public final class EnergyAwareScheduling {
+
+    static class Task {
+        String name;
+        int energyConsumption;
+
+        Task(String name, int energyConsumption) {
+            this.name = name;
+            this.energyConsumption = energyConsumption;
```
siriak, line 24: Doesn't energy consumption depend on execution time? With the current implementation execution order makes no difference because the total energy consumption is always the same

siriak (changes requested):

src/main/java/com/thealgorithms/scheduling/EnergyAwareScheduling.java
Hardvan: You are right, I have modified the code and test cases to reflect the same.

src/main/java/com/thealgorithms/scheduling/EnergyAwareScheduling.java
```diff
@@ -0,0 +1,70 @@
...
+    public Map<String, Object> scheduleTasks() {
+        tasks.sort(Comparator.comparingInt(t -> t.energyConsumption));
+
+        List<String> scheduledOrder = new ArrayList<>();
+        int currentTime = 0;
+        int totalEnergyConsumption = 0;
+        for (Task task : tasks) {
+            scheduledOrder.add(task.name);
+            currentTime += task.executionTime;
+            totalEnergyConsumption += currentTime * task.energyConsumption;
```
siriak, line 62: Why is consumed energy linearly proportional to the current time? The process should define when it consumes how much energy, or it should be defined somehow else. Please provide references where this algorithm is explained.

siriak (changes requested):

PR #5885, opened 2024-10-17
feat: add DoubleBuffer implementation and initial tests

abda-gaye:
- [x] **I have read and followed the guidelines outlined in [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).**  
  *Yes, I have read and followed the guidelines provided in the CONTRIBUTING.md file.*

- [x] **This pull request contains only my own work; it has not been plagiarized.**  
  *Yes, this pull request contains only my original work and has not been plagiarized.*

- [x] **All filenames follow the PascalCase convention.**  
  *Yes, all filenames follow the PascalCase convention.*

- [x] **All function and variable names adhere to Java naming conventions.**  
  *Yes, all function and variable names adhere to Java naming conventions.*

- [] **All new algorithms include a URL in their comments linking to a Wikipedia page or a similar source for reference.**  
 

- [x] **All code is formatted using `clang-format -i --style=file /datastructures/buffers/DoubleBufferTest.java`**

### DIFF
--- a/src/main/java/com/thealgorithms/datastructures/buffers/DoubleBuffer.java
+++ b/src/main/java/com/thealgorithms/datastructures/buffers/DoubleBuffer.java
@@ -1,0 +1,105 @@
+package com.thealgorithms.datastructures.buffers;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * The {@code DoubleBuffer} class implements a double buffering mechanism.
+ * It maintains two buffers, allowing the program to switch between them.
+ * This class is useful for separating read/write operations and optimizing memory usage.
+ *
+ * @param <Item> The type of elements stored in the double buffer.
+ */
+public class DoubleBuffer<Item> {
+    private final Item[] primaryBuffer;
+    private final Item[] secondaryBuffer;
+    private final AtomicInteger primaryIndex;
+    private final AtomicInteger secondaryIndex;
+    private final int capacity;
+    private final AtomicBoolean isPrimaryActive;
+
+    /**
+     * Constructor to initialize the double buffer with a specified capacity.
+     *
+     * @param capacity The size of each buffer.
+     * @throws IllegalArgumentException if the capacity is zero or negative.
+     */
+    public DoubleBuffer(int capacity) {
+        if (capacity <= 0) {
+            throw new IllegalArgumentException("Buffer size must be positive");
+        }
+        // noinspection unchecked
+        this.primaryBuffer = (Item[]) new Object[capacity];
+        this.secondaryBuffer = (Item[]) new Object[capacity];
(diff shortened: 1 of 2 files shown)


src/main/java/com/thealgorithms/datastructures/buffers/DoubleBuffer.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * The {@code DoubleBuffer} class implements a double buffering mechanism.
+ * It maintains two buffers, allowing the program to switch between them.
+ * This class is useful for separating read/write operations and optimizing memory usage.
+ *
+ * @param <Item> The type of elements stored in the double buffer.
+ */
+public class DoubleBuffer<Item> {
```
alxkm, line 13: What do you think about using `T` instead of `Item`. Because this is common practice

src/main/java/com/thealgorithms/datastructures/buffers/DoubleBuffer.java
```diff
@@ -0,0 +1,105 @@
...
+     * @throws IllegalArgumentException if the item is null.
+     */
+    public void put(Item item) {
+        if (item == null) {
+            throw new IllegalArgumentException("Null items are not allowed");
+        }
+
+        if (isPrimaryActive.get()) {
+            if (primaryIndex.get() < capacity) {
+                primaryBuffer[primaryIndex.getAndIncrement()] = item;
```
alxkm, line 78: As you want to make this code thread-safe, will be better to use `ReentrantLock` for this.


```
private final ReentrantLock lock = new ReentrantLock();

public void switchBuffer() {
    lock.lock();
    try {
        isPrimaryActive.set(!isPrimaryActive.get());
    } finally {
        lock.unlock();
    }
}

public boolean put(Item item) {
    if (item == null) {
        throw new IllegalArgumentException("Null items are not allowed");
    }
    lock.lock();
    try {
        if (isPrimaryActive.get()) {
            if (primaryIndex.get() < capacity) {
                primaryBuffer[primaryIndex.getAndIncrement()] = item;
                return true;
            }
        } else {
            if (secondaryIndex.get() < capacity) {
                secondaryBuffer[secondaryIndex.getAndIncrement()] = item;
                return true;
            }
        }
        return false; // Buffer is full
    } finally {
        lock.unlock();
    }
}

public Item get() {
    lock.lock();
    try {
        if (isPrimaryActive.get()) {
            if (primaryIndex.get() == 0) {
                return null;
            }
            return primaryBuffer[primaryIndex.decrementAndGet()];
        } else {
            if (secondaryIndex.get() == 0) {
                return null;
            }
            return secondaryBuffer[secondaryIndex.decrementAndGet()];
        }
    } finally {
        lock.unlock();
    }
}
```

alxkm (changes requested):

PR #5933, opened 2024-10-21
Longest Common Prefix

mrinalchauhan:
Here is the Approach :

1)Sort the Array: Sort the array of strings to bring strings with common prefixes adjacent to each other.

2)Identify Extremes: Select the first and last strings from the sorted array for comparison, as they will have the longest common prefix.

3)Character Comparison: Compare the characters of the first and last strings until a mismatch is found, tracking the index of the last matching character.

4)Return Prefix: Return the substring of the first string from the start to the index of the last matching character, which represents the longest common prefix.


siriak (changes requested):
Please replace main with JUnit tests and fix PR checks

PR #5956, opened 2024-10-23
update pss.java

khwaishchawla:
Proportional Share Scheduling (PSS) is a scheduling algorithm that assigns each process a weight based on its priority, and processes are scheduled to receive CPU time in proportion to their assigned weights. The higher the weight of a process, the more CPU time it will receive compared to processes with lower weights.


siriak (changes requested):
The code looks good, could you please remove main and add some JUnit tests? (see https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms)

khwaishchawla (commented):
made the required changes as asked

khwaishchawla:
> The code looks good, could you please remove main and add some JUnit tests? (see https://github.com/TheAlgorithms/Java/tree/master/src/test/java/com/thealgorithms)

Done with all the reqirements kindy review it

siriak:
There are no JUnit tests and PR checks are failing
image

PR #5959, opened 2024-10-23
Z Algorithm 

mrinalchauhan:
The Z Algorithm is an efficient string matching algorithm that computes a Z-array for a given string, which indicates the length of the longest substring starting from each index that matches the prefix of the string.

Key Points:
Z-array: For a string S, Z[i] represents the length of the longest substring starting at index i that is also a prefix of S.
Time Complexity: The algorithm runs in linear time, O(n + m), where n is the length of the text and m is the length of the pattern, making it faster than naive approaches.
Pattern Matching: By combining the pattern and the text into a single string (e.g., pattern$text), the Z-array helps efficiently locate occurrences of the pattern within the text.
Applications: It is widely used in text processing, DNA sequence analysis, and competitive programming for substring searching.
The Z Algorithm is valued for its efficiency and simplicity, allowing for quick implementation in various string matching scenarios.

### DIFF
--- a/src/main/java/com/thealgorithms/strings/Zalgorithm.java
+++ b/src/main/java/com/thealgorithms/strings/Zalgorithm.java
@@ -24,15 +24,8 @@ public final class Zalgorithm {
         int patternLength = pattern.length();
 
         for (int i = patternLength + 1; i < zArray.length; i++) {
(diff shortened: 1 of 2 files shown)


siriak (changes requested):
The code looks good, please check why PR checks are failing

PR #6030, opened 2024-10-27
added shuffle algorithm, it's implementation and problems related to it

Darshan-Baligeri:
Pull Request Description
Overview
This pull request adds multiple shuffle algorithms to the repository, each accompanied by its implementation and distinct problems that demonstrate its use. The following algorithms have been included:

Unique Pair Shuffle: This algorithm randomly creates pairs from elements in an array while avoiding duplicate pairings. If the input array has an odd number of elements, it returns an empty result.

Group Shuffle: This function splits an array into a defined number of groups and shuffles the contents of each group independently. It also addresses scenarios where the requested number of groups is greater than the length of the array.

Shuffle by Range: This method allows for shuffling a specific section of an array while leaving elements outside that range unchanged. It includes checks to verify that the specified start and end indices are valid.

Weighted Shuffle: This algorithm shuffles an array based on the assigned weights of its elements, ensuring that elements with higher weights are more likely to appear earlier in the shuffled output.

Constrained Shuffling: This technique rearranges an array so that no element remains in its original position. It features validation to confirm that this requirement is satisfied.

Implementation Details
Each algorithm has been developed with a focus on handling edge cases and validating inputs effectively.
In-depth inline documentation is provided to enhance understanding and usability, making the code accessible to beginners.
Unique problems associated with each algorithm have been crafted to broaden the practical applications of the shuffling methods.
Testing
All algorithms have undergone rigorous testing with a variety of inputs to verify their functionality, including edge cases like empty arrays, single-element arrays, and invalid inputs.

Conclusion
These contributions enrich the algorithm library by introducing new, practical algorithms along with relevant problems, fostering further exploration and application in shuffle-related tasks.


siriak (changes requested):
Please remove main functions and add JUnit tests

PR #6053, opened 2024-10-28
Added ShuffleAlgorithm and various problems related to it.

Darshan-Baligeri:
Shuffle Algorithms for Various Scenarios
This PR introduces multiple shuffle algorithms to the com.thealgorithms.shufflealgorithm package, with each class implementing a specific type of shuffle technique. The classes are designed to be flexible for different shuffling scenarios with well-documented methods and edge case handling. Below is a summary of the classes and the shuffle techniques they provide:

1. ConstrainedShuffle
Purpose: Shuffle the elements in an array while keeping the first and last elements fixed.
Method: constrainedShuffle(int[] array)
Edge Case Handling: No shuffling occurs if the array has fewer than 3 elements.
2. GroupShuffle
Purpose: Groups elements into subarrays of a specified size and shuffles those groups.
Method: groupShuffle(int[] array, int groupSize) -> List<List<Integer>>
Edge Case Handling: Ensures that each group has the specified size (or fewer for the last group) and skips shuffling if groupSize <= 1.
3. ShuffleByRange
Purpose: Shuffle elements within a specified index range.
Method: shuffleByRange(int[] array, int start, int end)
Edge Case Handling: Verifies valid indices to avoid IndexOutOfBounds issues.
4. UnderstandingShuffleAlgo
Purpose: Basic shuffle algorithm that uses a uniform probability approach to randomize array elements.
Method: shuffle(int[] array)
5. UniquePairShuffle
Purpose: Pairs elements randomly without repeating pairs.
Method: pairShuffle(int[] array) -> List<int[]>
Edge Case Handling: Returns an empty list if the array has an odd length or fewer than 2 elements.
6. WeightedShuffle
Purpose: Shuffles elements based on their weights, favoring elements with higher weights.
Method: weightedShuffle(int[] array, int[] weights)
Edge Case Handling: Ensures the weights array matches the array length for proper assignment

### DIFF
--- a/src/main/java/com/thealgorithms/shufflealgo/ConstrainedShuffle.java
+++ b/src/main/java/com/thealgorithms/shufflealgo/ConstrainedShuffle.java
@@ -1,0 +1,48 @@
+package com.thealgorithms.shufflealgo;
+
+import java.util.Arrays;
+import java.util.Random;
+
+public
+final class ConstrainedShuffle {
+
+    private
+    ConstrainedShuffle() {
+        // Prevent instantiation
+    }
+
+    /**
+     * Shuffles elements in the array while ensuring that the first and last
+     * elements remain fixed.
+     *
+     * @param array the input array to shuffle
+     */
+    public
+    static void constrainedShuffle(int[] array) {
+        // Edge case: If the array has less than 3 elements, no shuffling can occur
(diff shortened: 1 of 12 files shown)


siriak (changes requested):
Please fix PR checks

PR #6086, opened 2024-11-09
Fix loops issues

Treizzer:
Modified while and for loop to fix the infinite loop and negative indices issues, respectively.
What happened before the fix:
Whether you have and array of numbers like -> Integer[] numbers = { 3, 12, 14, 23, 34, 45, 56, 67, 78, 89 };
When you send a value to "Key" greater than the last number in your array, e.g. 90 or greater, the program got into infinite loop and never return the expected "-1" for "key not found".
In case what the for-loop assignment a negative number to the variable "i" the condition will skip the negative index, and the for-loop will continue with the normal flow.

### DIFF
--- a/src/main/java/com/thealgorithms/searches/JumpSearch.java
+++ b/src/main/java/com/thealgorithms/searches/JumpSearch.java
@@ -41,12 +41,18 @@ public class JumpSearch implements SearchAlgorithm {
 
         int limit = blockSize;
         // Jumping ahead to find the block where the key may be located
-        while (limit < length && key.compareTo(array[limit]) > 0) {
-            limit = Math.min(limit + blockSize, length - 1);
+        while (limit < length && key.compareTo(array[Math.min(limit, length - 1)]) > 0) {
+            limit += blockSize;
         }
+
+        // Handle the case where limit exceeds array length
+        limit = Math.min(limit, length);
 
         // Perform linear search within the identified block
         for (int i = limit - blockSize; i <= limit && i < length; i++) {
+            if (i < 0) { // Skip negative indices
+                continue;
+            }
             if (array[i].equals(key)) {
                 return i;
             }


src/main/java/com/thealgorithms/searches/JumpSearch.java
```diff
@@ -41,12 +41,18 @@ public <T extends Comparable<T>> int find(T[] array, T key) {
...
         int limit = blockSize;
         // Jumping ahead to find the block where the key may be located
-        while (limit < length && key.compareTo(array[limit]) > 0) {
-            limit = Math.min(limit + blockSize, length - 1);
+        while (limit < length && key.compareTo(array[Math.min(limit, length - 1)]) > 0) {
+            limit += blockSize;
         }
 
+        // Handle the case where limit exceeds array length
+        limit = Math.min(limit, length);
```
alxkm, line 49: What do you think to rewrite below loop to:

```
int start = Math.max(0, limit - blockSize);
int end = Math.min(limit, length);

// Perform linear search within the identified block
for (int i = start; i < end; i++) {
    if (array[i].equals(key)) {
        return i;
    }
}
```

alxkm (changes requested):

src/main/java/com/thealgorithms/searches/JumpSearch.java
```diff
@@ -41,12 +41,18 @@ public <T extends Comparable<T>> int find(T[] array, T key) {
 
         int limit = blockSize;
         // Jumping ahead to find the block where the key may be located
-        while (limit < length && key.compareTo(array[limit]) > 0) {
-            limit = Math.min(limit + blockSize, length - 1);
+        while (limit < length && key.compareTo(array[Math.min(limit, length - 1)]) > 0) {
```
alxkm, line 44: Could you please add your test case, with infinite loop, to `com.thealgorithms.searches.JumpSearchTest`

alxkm (changes requested):

PR #6144, opened 2025-01-18
Fix absolute max bug

rizwan-ilyas:
The AbsoluteMax class contains a bug in the getMaxValue method. When two numbers have the same absolute value (e.g., 5 and -5), the method returns the number that appears later in the array. This behavior is inconsistent and may lead to unexpected results.

Explanation with Example.

Before Fix
5, -5 returns -5 
-5, 5 returns  5

After Fix
5, -5 returns  5 
-5, 5 returns  5

Thanks

### DIFF
--- a/src/main/java/com/thealgorithms/maths/AbsoluteMax.java
+++ b/src/main/java/com/thealgorithms/maths/AbsoluteMax.java
@@ -17,7 +17,7 @@ public final class AbsoluteMax {
         }
         int absMax = numbers[0];
         for (int i = 1; i < numbers.length; i++) {
-            if (Math.abs(numbers[i]) > Math.abs(absMax)) {
+            if (Math.abs(numbers[i]) > Math.abs(absMax) || (Math.abs(numbers[i]) == Math.abs(absMax) && numbers[i] > absMax)) {
                 absMax = numbers[i];
             }
         }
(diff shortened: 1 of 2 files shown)


src/main/java/com/thealgorithms/maths/AbsoluteMax.java
```diff
@@ -17,7 +17,7 @@ public static int getMaxValue(int... numbers) {
         }
         int absMax = numbers[0];
         for (int i = 1; i < numbers.length; i++) {
-            if (Math.abs(numbers[i]) > Math.abs(absMax)) {
+            if (Math.abs(numbers[i]) > Math.abs(absMax) || (Math.abs(numbers[i]) == Math.abs(absMax) && numbers[i] > absMax)) {
                 absMax = numbers[i];
             }
```
siriak, line 22:
```suggestion
            if (Math.abs(numbers[i]) >= Math.abs(absMax)) {
                absMax = Math.max(absMax, numbers[i]);
            }
```
Can it be done like that? The initially proposed condition looks quite complex

rizwan-ilyas: Surely we can do like that, but in term of cost/performance **Math.max()** will be called in every iteration (on matched condition) that's why I preferred to use logical operation (means if Math.abs(numbers[i]) > Math.abs(absMax) will be true **the rest of the condition will be ignored**) and in this way we can eliminate **Math.max()** cost.

Please suggest .!

siriak (approved):
Looks good, thanks!

PR #6281, opened 2025-06-10
 Added Anagram Checker in Java

farheen-shaikh530:
---

## ✅ Features

- Ignores case (`"Listen"` and `"Silent"` are anagrams)
- Ignores whitespace (`"Clint Eastwood"` and `"Old West Action"`)
- Uses **HashMap** for character frequency count
- Efficient: **Time Complexity:** `O(n)` | **Space Complexity:** `O(n)`
- Demonstrates **mid-level Java skills**:
  - String manipulation
  - HashMap usage
  - Input sanitization
  - Edge case handling

---

## 💡 How It Works

1. Converts both strings to lowercase and removes all spaces.
2. Uses a `HashMap<Character, Integer>` to count the frequency of each character in the first string.
3. Iterates over the second string and decrements the character counts.
4. If all frequencies return to zero and the map is empty, the strings are anagrams.

---

## 🧪 Example

```java
    String s1 = "Listen";
    String s2 = "Silent";

    if (areAnagrams(s1, s2)) {
        System.out.println("\"" + s1 + "\" and \"" + s2 + "\" are anagrams.");
    } else {
        System.out.println("\"" + s1 + "\" and \"" + s2 + "\" are NOT anagrams.");
    }
}

### DIFF
--- a/Search/BinarySearch.java
+++ b/Search/BinarySearch.java
@@ -1,0 +1,38 @@
+/**
+ * BinarySearch.java
+ * This program implements recursive Binary Search on a sorted array.
+ * Time Complexity: O(log n)
+ */
+
+public class BinarySearch {
+
+    // Recursive method to perform binary search
+    public static int binarySearch(int[] arr, int left, int right, int target) {
+        if (left > right) {
+            return -1;  // Base case: element not found
+        }
+
+        int mid = left + (right - left) / 2;
+
+        if (arr[mid] == target) {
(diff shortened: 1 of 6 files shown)


vil02:
All of these are already implemented. Additionally please place your code in [`src/main/java/com/thealgorithms`](https://github.com/TheAlgorithms/Java/tree/master/src/main/java/com/thealgorithms).

PR #6288, opened 2025-06-11
Add LambdaExpressionUtils.java – Unique Lambda-Based Helper Functions

farheen-shaikh530:
This PR adds a unique, creative file Functional/LambdaExpressionUtils.java that showcases Java 8+ lambda expressions through mini utilities using:
	•	Function<String, String> → Reverse a string
	•	Predicate<String> → Check palindrome
	•	Consumer<String> → Print shout version
	•	Function<Integer, Boolean> → Check if even
	•	Supplier<String> → Random greeting generator

PR #6291, opened 2025-06-11
chore: monitor structure of the project

vil02:
The goal of this change is to make the CI fail in case when some java files are not placed in the _correct_ location (cf. #6288).

I realize that a python script might be an overkill at the first glance. I guess writing some smart shell script is possible, but in the long run, one could easily make the suggested solution more strict (do we have a naming convention of our tests? should we have one?).

Cf. the [_failing log_](https://github.com/TheAlgorithms/Java/actions/runs/15596856129/job/43929121871).


check_structure.py
```diff
@@ -0,0 +1,27 @@
+import pathlib
```
siriak, line 1: Maybe put this file elsewhere? At least in .github folder

siriak (approved):
Looks good, thanks!

PR #6333, opened 2025-07-01
Fix(searches): Correct column pointer initialization in RowColumnWiseSorted2dArrayBinarySearch

### LINKED ISSUE #6332: [BUG] Incorrect column pointer initialization in RowColumnWiseSorted2dArrayBinarySearch
### Description

-  I've identified a bug in the `RowColumnWiseSorted2dArrayBinarySearch.java` implementation.
- The column pointer (colPointer) is initialized using `matrix.length - 1`. But, **matrix.length returns the number of rows, not the number of columns**.
- This causes the algorithm's starting point to be incorrect for any matrix that is not square.
- The problematic line is:
```java
// in src/main/java/com/thealgorithms/searches/RowColumnWiseSorted2dArrayBinarySearch.java

int colPointer = matrix.length - 1; // The pointer at end column
```

### Steps to reproduce

- The bug can be reproduced with any non-square matrix where the number of rows and columns are different.
- For example, with a 2x4 matrix:
```java
   1 int[][] matrix = {
   2     {10, 20, 30, 40},
   3     {15, 25, 35, 45}
   4 };
```
- With the current code, matrix.length is 2, so colPointer is initialized to 1.
- The search incorrectly starts at matrix[0][1] (value 20) instead of the top-right corner matrix[0][3] (value 40).

### Excepted behavior

- The colPointer should be initialized to the index of the last column, allowing the search to correctly start at the top-right corner of the matrix. The correct initialization should be:
`int colPointer = matrix[0].length - 1;`

### Screenshots

_No response_

### Additional context

I would like to submit a Pull Request to fix this initialization issue.

codingmydna:

Closes #6332 

This pull request fixes a bug in the `RowColumnWiseSorted2dArrayBinarySearch.java` file where the column pointer was incorrectly initialized.

**The Problem:**
The column pointer (`colPointer`) was being initialized with `matrix.length - 1`. However, `matrix.length` returns the number of rows, not columns. This causes the algorithm to start at an incorrect position and fail for any matrix that is not square.

**The Solution:**
I have corrected the initialization to `int colPointer = matrix[0].length - 1;`. This correctly uses the length of the first row to determine the number of columns, ensuring the pointer is set to the last column's index as intended by the algorithm's logic.


codingmydna:
> Why created as Draft? Looks good to me.

I thought it didn't pass a CI check, `Infer`.
If it doesn't matter, then I'll switch to `Ready for review`

PR #6368, opened 2025-07-11
Update EMAFilter.java

ShreyanshJain105:
### Summary
Refactored the EMAFilter class to reduce code verbosity while maintaining identical functionality and behavior.

### Changes
- Condensed verbose JavaDoc comments to essential formula documentation
- Shortened variable names for better readability (`emaValue` → `ema`, `audioSignal` → `signal`)
- Streamlined constructor validation and variable initialization
- Combined initialization and assignment where possible

### Testing
- [ ] All existing unit tests pass
- [ ] Audio signal processing behavior remains unchanged
- [ ] Input validation still works correctly

### Impact
- **Functionality**: No changes to public API or behavior
- **Performance**: No performance impact
- **Readability**: Improved code conciseness without losing clarity
- **Maintenance**: Easier to read and maintain

The EMA filter continues to apply the same exponential moving average formula: `EMA[i] = alpha * signal[i] + (1-alpha) * EMA[i-1]`

### DIFF
--- a/src/main/java/com/thealgorithms/audiofilters/EMAFilter.java
+++ b/src/main/java/com/thealgorithms/audiofilters/EMAFilter.java
@@ -1,48 +1,28 @@
 package com.thealgorithms.audiofilters;
 
 /**
- * Exponential Moving Average (EMA) Filter for smoothing audio signals.
- *
- * <p>This filter applies an exponential moving average to a sequence of audio
- * signal values, making it useful for smoothing out rapid fluctuations.
- * The smoothing factor (alpha) controls the degree of smoothing.
- *
- * <p>Based on the definition from
- * <a href="https://en.wikipedia.org/wiki/Moving_average">Wikipedia link</a>.
+ * EMA Filter for audio signal smoothing.
+ * Formula: EMA[i] = alpha * signal[i] + (1-alpha) * EMA[i-1]
  */
 public class EMAFilter {
     private final double alpha;
-    private double emaValue;
-    /**
-     * Constructs an EMA filter with a given smoothing factor.
-     *
-     * @param alpha Smoothing factor (0 < alpha <= 1)
-     * @throws IllegalArgumentException if alpha is not in (0, 1]
-     */
+    private double ema;
+
     public EMAFilter(double alpha) {
-        if (alpha <= 0 || alpha > 1) {
-            throw new IllegalArgumentException("Alpha must be between 0 and 1.");
-        }
+        if (alpha <= 0 || alpha > 1) throw new IllegalArgumentException("Alpha must be in (0,1]");
         this.alpha = alpha;
-        this.emaValue = 0.0;
     }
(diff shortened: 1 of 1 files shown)


alxkm:
@ShreyanshJain105

The build failed. Need to fix the build to start review. But what is the point of this PR?

As this repository is mostly for studying algorithms, the older version is actually better for educational purposes, even though the newer version is slightly more concise. 

Here’s why:
- audioSignal is clearer than just signal because it specifies the context (audio processing).
- emaSignal is more explicit than result because it indicates that the output is still an EMA-processed signal.

Better Documentation for Learners:
The Javadoc explains:
- What an EMA filter is.
Why it’s useful (smoothing audio signals).
- A reference to Wikipedia for further study.

This helps students understand the algorithm’s purpose before diving into the code.

More Readable Error Message:
- "Alpha must be between 0 and 1." is more intuitive for beginners than "Alpha must be in (0,1]".

Explicit Initialization of emaValue in Constructor:
- While not strictly necessary, initializing emaValue = 0.0 in the constructor makes it clearer that the filter starts from zero.

But we can combine these approaches and refactor the code as follows:

```
/**
 * Exponential Moving Average (EMA) Filter for smoothing audio signals.
 *
 * <p>This filter applies an exponential moving average to a sequence of audio
 * signal values, making it useful for smoothing out rapid fluctuations.
 * The smoothing factor (alpha) controls the degree of smoothing.
 *
 * <p>Formula: EMA[i] = alpha * signal[i] + (1 - alpha) * EMA[i-1]
 * Based on <a href="https://en.wikipedia.org/wiki/Moving_average">Wikipedia</a>.
 */
public class EMAFilter {
    private final double alpha;
    private double emaValue;

    /**
     * Constructs an EMA filter with a given smoothing factor.
     *
     * @param alpha Smoothing factor (0 < alpha ≤ 1)
     * @throws IllegalArgumentException if alpha is not in (0, 1]
     */
    public EMAFilter(double alpha) {
        if (alpha <= 0 || alpha > 1) {
            throw new IllegalArgumentException("Alpha must be between 0 and 1.");
        }
        this.alpha = alpha;
        this.emaValue = 0.0; // Initialize EMA value
    }

    /**
     * Applies the EMA filter to an audio signal array.
     *
     * @param audioSignal Array of audio samples to process.
     * @return Array of smoothed samples.
     */
    public double[] apply(double[] audioSignal) {
        if (audioSignal.length == 0) {
            return new double[0];
        }
        
        double[] smoothedSignal = new double[audioSignal.length];
        emaValue = audioSignal[0]; // Reset EMA to first sample
        smoothedSignal[0] = emaValue;

        for (int i = 1; i < audioSignal.length; i++) {
            emaValue = alpha * audioSignal[i] + (1 - alpha) * emaValue;
            smoothedSignal[i] = emaValue;
        }
        return smoothedSignal;
    }
}
```

PR #6370, opened 2025-07-12
Add PalindromeChecker to check if a string is a palindrome

Naveenkumarsuk:
### Summary
Added `PalindromeChecker.java` in the `strings/` folder.

### What It Does
- Checks if a string is a palindrome
- Ignores spaces, punctuation, and letter casing
- Useful for learning string manipulation and logic


DenizAltunkapan:
@Naveenkumarsuk  I think the two pointer palindrome is already implemented in https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/strings/Palindrome.java

PR #6407, opened 2025-07-18
Add Jump Search algorithm in Java

Fahham2903:
Added an implementation of the Jump Search algorithm in Java.
This is part of my first real code-based open source contribution.

### DIFF
--- a/src/main/java/com/thealgorithms/searches/JumpSearch.java
+++ b/src/main/java/com/thealgorithms/searches/JumpSearch.java
@@ -8,7 +8,7 @@ package com.thealgorithms.searches;
  *
  * Reference: https://en.wikipedia.org/wiki/Jump_search
  */
-public class JumpSearch {
+public final class JumpSearch {
 
     // Prevent instantiation
     private JumpSearch() {


Fahham2903:
All required checks have passed except `Infer`, which is failing due to a known camlzip dependency issue. This error is unrelated to my changes.
Ready for review.

DenizAltunkapan:
The older version is preferable because it is generic, allowing searches on any comparable type, not just primitive int. It also implements a standard SearchAlgorithm interface, which improves code modularity, testability, and reuse. The new version’s static methods and manual conversion from Integer[] to int[] introduce unnecessary overhead and reduce flexibility. Therefore, i would say the older version is cleaner, more extensible...
Please try to contribute in another way :]

@alxkm thoughts?

alxkm:
@DenizAltunkapan, I completely agree with your point. The current approach closely follows the design principles used in the Java Collections Framework, which is an industry standard. Since this is primarily a study repository, exposing students to this kind of structure - using interfaces, generic types, and modular design - helps them build knowledge that aligns with industry standards and is directly applicable to professional development.

PR #6421, opened 2025-07-20
feat: Add 0/1 Knapsack Problem: Recursive and Tabulation (Bottom-Up DP) Implementations in Java along with their corresponding Tests

o000SAI000o:
1.Recursive Solution
-Uses a top-down approach without memoization.
-Highlights the decision process for including or excluding each item.
2.Tabulation (Bottom-Up DP) Solution
-Efficient dynamic programming approach using a 2D table.
-Includes a utility function to print the DP table.
-Demonstrates space-time tradeoff compared to recursion.

Both implementations follow consistent formatting, inline documentation, and are aligned with the repository's existing code structure.

### DIFF
--- a/src/main/java/com/thealgorithms/dynamicprogramming/ZeroOneKnapsack.java
+++ b/src/main/java/com/thealgorithms/dynamicprogramming/ZeroOneKnapsack.java
@@ -1,0 +1,50 @@
+package com.thealgorithms.dynamicprogramming;
+
+/**
+ * Class to solve the 0/1 Knapsack Problem using recursion.
+ *
+ * The 0/1 Knapsack problem is a classic dynamic programming problem
+ * where we are given weights and values of `n` items. We need to put
+ * these items in a knapsack of capacity `W` to get the maximum total value
+ * in the knapsack. We can either include an item or exclude it — but cannot
+ * include it more than once.
+ *
+ * Time Complexity: O(2^n) in worst case for the pure recursive approach (due to overlapping subproblems).
+ * Using memoization or bottom-up dynamic programming reduces the time complexity to O(nW).
+ *
+ * Example:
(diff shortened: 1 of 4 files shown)


o000SAI000o:
can u help me with the errors?

alxkm:
@o000SAI000o thank you for your contribution to the repository.

It looks like there are already several existing implementations of the Knapsack problem in the repository, and your solution appears to be very similar to them:

https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/dynamicprogramming/Knapsack.java
https://github.com/TheAlgorithms/Java/blob/master/src/test/java/com/thealgorithms/dynamicprogramming/KnapsackTest.java

https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/dynamicprogramming/KnapsackMemoization.java
https://github.com/TheAlgorithms/Java/blob/master/src/test/java/com/thealgorithms/dynamicprogramming/KnapsackMemoizationTest.java

Additionally, other variations of the Knapsack algorithm already exist in the repository.

And the error appears because the filename does not match the public class name exactly, which is a strict rule in Java.
And to fix this the class should be renamed:
public class ZeroOneknapsackTabTest to public class ZeroOneKnapsackTest

o000SAI000o:
thanks alot, i thought ZeroOneknapsack is a diffferent and popular version of it thts why i tried that, ans sorry for filename tht was by mistake

o000SAI000o:
knapsack tabulation is not in repo which is very time effient with a O(N) time complexity , is it ok if i add tht code in repo, it will be efficient to alot of learners.

alxkm:
@o000SAI000o the build needs to be fixed.

o000SAI000o:
sure ill fix the build

Issue #6438, opened 2025-07-24
Aplicación del patrón Chain of Responsibility en HammingDistance


val28vv:
### What would you like to share?

Informe Técnico: Mejora para HammingDistance
Autor: Valentina Falconi
Fecha: 23/07/25

Problemas y Soluciones
Chain of Responsibility para cálculo de Hamming Distance
Problema: El método compute de la clase HammingDistance está acoplado, realizando validaciones y procesamiento dentro del mismo bloque. Esto limita la extensibilidad y dificulta agregar nuevas reglas o validaciones.
Solución: Implementar una cadena de responsabilidades mediante clases individuales: ValidateBinaryHandler, ValidateLengthHandler y ComputeHammingHandler, todas derivadas de la clase abstracta HammingHandler. Cada clase se encarga de una responsabilidad y delega al siguiente paso.
Beneficio: Permite separar las responsabilidades y facilita la extensión del procesamiento sin modificar código existente. El flujo de validación y cálculo se puede adaptar dinámicamente.

Impacto
- Código desacoplado y extensible
- Aplicación clara del principio de responsabilidad única
- Facilita agregar nuevas reglas sin romper el código

Recomendación: Implementar en rama feature/patterns-chain-hamming.
Valentina Falconi
valentina.falconi@espol.edu.ec

dinishajais:
Hi, I have worked on this issue and created the pull request. One build check is not passed and that looks like some build job issue with clang format. I have added these new files tried changing it multiple time but still the particular clang-format is failing. But that won't affect the code implementation. Please have a check.

DenizAltunkapan:
Thank you for your contribution and the detailed implementation!
After careful consideration and review, we have come to the conclusion that this approach is not suitable for this repository.

While the Chain of Responsibility approach is technically sound and demonstrates good design principles, I think it introduces unnecessary complexity for this repository. The goal of the-algorithms/Java is to provide simple and easily understandable implementations of algorithms for learners.

For HammingDistance, it is better to keep it as a single class with a static method, which makes it straightforward for beginners to understand and use.
In my opinion, adding extra validation classes for this algorithm is unnecessary and constitutes overengineering.

Thank you again for your effort and contribution!

PR #6447, opened 2025-07-27
#6438 Improvement of Hamming Distance class

dinishajais:
Issue #6438
I have created abstract class HammingHandler for managing various checks as per demanded in the issue raised. Also I have created additional test classes that would test the new classes created. 

👍 I have read [CONTRIBUTING.md]
👍  This pull request is all my own work -- I have not plagiarized it.
👍  All filenames are in PascalCase.
👍  All functions and variable names follow Java naming conventions.
👍  All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.
👍  All new code is formatted with `clang-format -i --style=file path/to/your/file.java`


dinishajais:
I have tried resolving the formatting multiple times , but it seems each time I change the formatting it asks me to revert back the changes. Though there are no errors in here, only a formatting issue. Please give it a check once.

DenizAltunkapan:
Please install the clang-formatter locally, e.g. on linux `sudo apt install clang-format`
After that run the following command `find src -name "*.java" -exec clang-format -i -style=file {} +`
For other OS, you have to do some research (https://clang.llvm.org/docs/ClangFormat.html)

dinishajais:
Thank you for the help. @DenizAltunkapan

DenizAltunkapan:
Thank you for your contribution and the detailed implementation!
After careful consideration and review, we have come to the conclusion that this approach is not suitable for this repository.

While the Chain of Responsibility approach is technically sound and demonstrates good design principles, I think it introduces unnecessary complexity for this repository. The goal of the-algorithms/Java is to provide simple and easily understandable implementations of algorithms for learners.

For HammingDistance, it is better to keep it as a single class with a static method, which makes it straightforward for beginners to understand and use.
In my opinion, adding extra validation classes for this algorithm is unnecessary and constitutes overengineering.

Thank you again for your effort and contribution!

Issue #6451, opened 2025-07-29
[FEATURE REQUEST] <ReverseStringUsingStack>


navaneedan07:
### What would you like to Propose?

ReverseStringUsingStack
In stacks, it is better to add an implementation of reversing string using stack

### Issue details

Using stacks we can reverse it easily by pushing character by character from string to stack.
Then popping characters to the string

VSENTHAMIZHAN:
I wanna contribute for this issue @navaneedan07

navaneedan07:
Hi I have already made pull request for this issue
  #6452

PR #6454, opened 2025-07-31
feat: add ReverseStringUsingStack (#6451)

Dhirajgiri18:
This PR adds an implementation for reversing a string using a stack, as requested in #6451 . The algorithm pushes each character of the input string onto a stack and then pops them to construct the reversed string. The code includes detailed comments and proper resource management by closing the Scanner.


DenizAltunkapan:
@Dhirajgiri18 Already exists, i'm sorry. Please find another way to contribute :100:

PR #6467, opened 2025-08-14
Add DFS with parent-completion constraint for DAG traversal

StathisVeinoglou:
## Description
This PR adds a new `GraphTraversal` class that implements a modified DFS traversal order for directed graphs, ensuring that a node is only visited if all its parents have already been visited.  
This approach is useful in dependency resolution and neural network computation graph execution order.  

### DIFF
--- a/DIRECTORY.md
+++ b/DIRECTORY.md
@@ -346,6 +346,7 @@
           - 📄 [Point](src/main/java/com/thealgorithms/geometry/Point.java)
         - 📁 **graph**
           - 📄 [ConstrainedShortestPath](src/main/java/com/thealgorithms/graph/ConstrainedShortestPath.java)
+          - 📄 [HopcroftKarp](src/main/java/com/thealgorithms/graph/HopcroftKarp.java)
           - 📄 [StronglyConnectedComponentOptimized](src/main/java/com/thealgorithms/graph/StronglyConnectedComponentOptimized.java)
           - 📄 [TravelingSalesman](src/main/java/com/thealgorithms/graph/TravelingSalesman.java)
         - 📁 **greedyalgorithms**
@@ -1051,6 +1052,7 @@
           - 📄 [PointTest](src/test/java/com/thealgorithms/geometry/PointTest.java)
         - 📁 **graph**
           - 📄 [ConstrainedShortestPathTest](src/test/java/com/thealgorithms/graph/ConstrainedShortestPathTest.java)
+          - 📄 [HopcroftKarpTest](src/test/java/com/thealgorithms/graph/HopcroftKarpTest.java)
           - 📄 [StronglyConnectedComponentOptimizedTest](src/test/java/com/thealgorithms/graph/StronglyConnectedComponentOptimizedTest.java)
           - 📄 [TravelingSalesmanTest](src/test/java/com/thealgorithms/graph/TravelingSalesmanTest.java)
         - 📁 **greedyalgorithms**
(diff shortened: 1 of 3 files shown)


src/main/java/com/thealgorithms/graph/GraphTraversal.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        List<TraversalEvent<T>> events = new ArrayList<>();
+        Set<T> visited = new HashSet<>();
+        int[] order = {0};
+        dfs(start, successors, predecessors, visited, order, events);
+        return Collections.unmodifiableList(events);
+    }
+
+    private static <T> void dfs(T u, Map<T, List<T>> succ, Map<T, List<T>> pred, Set<T> visited, int[] order, List<TraversalEvent<T>> out) {
```
alxkm, line 121: In this case, it is better to use full names. It will look more descriptive for learning purposes.

src/main/java/com/thealgorithms/graph/GraphTraversal.java
```diff
@@ -0,0 +1,173 @@
...
+ */
+
+public final class GraphTraversal {
+
+    private GraphTraversal() {
+        // utility class
+    }
+
+    /** An event emitted by the traversal: either a VISIT with an order, or a SKIP with a note. */
+    public static final class TraversalEvent<T> {
```
alxkm, line 34: Will be better to use recod here:

```
/** An event emitted by the traversal: either a VISIT with an order, or a SKIP with a note. */
public record TraversalEvent<T>(
    T node,
    Integer order, // non-null for visit, null for skip  
    String note    // non-null for skip, null for visit
) {
    public TraversalEvent {
        Objects.requireNonNull(node);
        // order and note can be null based on event type
    }
    
    /** A visit event with an increasing order (0,1,2,...) */
    public static <T> TraversalEvent<T> visit(T node, int order) {
        return new TraversalEvent<>(node, order, null);
    }
    
    /** A skip event with an explanatory note (e.g., not all parents visited yet). */
    public static <T> TraversalEvent<T> skip(T node, String note) {
        return new TraversalEvent<>(node, null, Objects.requireNonNull(note));
    }
    
    public boolean isVisit() { return order != null; }
    public boolean isSkip() { return order == null; }
    
    @Override
    public String toString() {
        return isVisit() ? "VISIT(" + node + ", order=" + order + ")" 
                        : "SKIP(" + node + ", " + note + ")";
    }
}
```

src/main/java/com/thealgorithms/graph/GraphTraversal.java
```diff
@@ -0,0 +1,173 @@
...
+    private static <T> boolean allParentsVisited(T node, Set<T> visited, Map<T, List<T>> pred) {
+        for (T p : pred.getOrDefault(node, List.of())) {
+            if (!visited.contains(p)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private static <T> boolean appearsAnywhere(Map<T, List<T>> succ, T node) {
```
alxkm, line 150: Same here. Seems like it's better to use full descriptive names.

alxkm (changes requested):

src/main/java/com/thealgorithms/dynamicprogramming/PartitionProblem.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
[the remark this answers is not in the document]
StathisVeinoglou: Thank you! I’ve reverted the change as requested.

src/main/java/com/thealgorithms/graph/GraphTraversal.java
StathisVeinoglou: Yea good point! I believe that i replaced all abbreviated names with full descriptive ones.

StathisVeinoglou:
Thanks for all the suggestions! I believe that i implemented the changes.
Please let me know if anything further is needed

PR #6548, opened 2025-09-28
Added Binary Search Tree algorithm and Tests

UdayaKrishnanM:
- Binary Search Tree is implemented using recursion for insertion, height calculation, and traversal.
- JavaDoc comments were added for the class and all public methods.
- Test Cases Added like for normal values, empty tree, negative values, duplicate values, sorted input for balanced tree 
- Time Complexity
    - Insertion/Search: 
        - Average: O(log n)
        - Worst: O(n) (skewed tree) 
    - Traversal: O(n)
    - Balance Check: O(n)
    
- Space Complexity
    - Storage: O(n) 
    - Recursion Stack: 
        - Average:  O(log n) 
        - Worst: O(n)


DenizAltunkapan:
Thanks for the contribution! 🙌 @UdayaKrishnanM 

However, we already have an existing BST implementation in the repository:
- [`BSTRecursive`](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/trees/BSTRecursive.java)

We also already have an optimized way to check if a binary tree is balanced:
- [`CheckIfBinaryTreeBalanced`](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/trees/CheckIfBinaryTreeBalanced.java)

This means that most of the methods introduced here already exist in the repo (and in a more efficient way).  
It would create code duplication if we added another class with the same functionality.

The one part of your PR that could add real value is the `prettyDisplay()` method, which is currently missing in our existing BST implementations.  
I’d suggest extracting just this printing functionality and adding it to one of the existing classes instead of duplicating the whole BST implementation.
Or you can suggest another way to integrate your code, but a bst already exists..

UdayaKrishnanM:
Thank you @DenizAltunkapan for the detailed review and for pointing out the existing implementations🙌

You're absolutely right that duplicating core BST functionality would add redundancy. I’ll refactor the PR to extract only the prettyDisplay() method and integrate it into one of the existing BST classes, such as BSTRecursive, to avoid duplication and maintain consistency.

If needed, I can also add a small test case to demonstrate its usage. Let me know if you have a preference for which class it should go into or if you'd like it as a utility method. 

Thanks again for the guidance! 😊

DenizAltunkapan:
@UdayaKrishnanM sounds good! Feel free to add more test cases🙂

UdayaKrishnanM:
Thank you, can you assign me by creating a new issue or can I submit the PR directly?

DenizAltunkapan:
> Thank you, can you assign me by creating a new issue or can I submit the PR directly?

you can submit the PR directly. thank you @UdayaKrishnanM

DenizAltunkapan:
> Hi @DenizAltunkapan, I was looking on this code [BSTRecursiveGeneric](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/trees/BSTRecursiveGeneric.java) [BSTRecursive](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/datastructures/trees/BSTRecursive.java). When I try to add negative values to the node it converts it to positive node value.
> 
> @DenizAltunkapan, Should I also work on this issue (node can take negative value) or not? image image

Yes, please @UdayaKrishnanM

DenizAltunkapan:
Please reopen this PR if you are using this for the requested changes above. Otherwise you can also open a seperated PR. @UdayaKrishnanM

PR #6568, opened 2025-10-01
Implement topological sort with DAG validation and cycle detection

the-yash-rajput:

### DIFF
--- a/src/main/java/com/thealgorithms/graph/TopologicalSort.java
+++ b/src/main/java/com/thealgorithms/graph/TopologicalSort.java
@@ -108,10 +108,8 @@ public final class TopologicalSort {
         Stack<Integer> stack = new Stack<>();
 
         for (int i = 0; i < n; i++) {
-            if (!visited[i]) {
-                if (dfs(i, visited, recStack, adj, stack)) {
-                    throw new IllegalArgumentException("cycle detected in graph");
-                }
+            if (!visited[i] && dfs(i, visited, recStack, adj, stack)) {
+                throw new IllegalArgumentException("cycle detected in graph");
             }
         }
 


alxkm:
Looks like this is already in the repository, but in the wrong package it is definitely related to the graph

https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/sorts/TopologicalSort.java

the-yash-rajput:
@alxkm How would you like to proceed? 
Should we merge this PR or replace the previous implementation with the correct location?


@alxkm I think we can include both, as the previous algorithm is not more friendly to graphs.

alxkm:
@siriak, @DenizAltunkapan, Could you please help clarify our approach here?

There are already several implementations of this algorithm in the repository:

The previous implementation follows a more traditional pattern as it operates directly on color values. 
The current implementation also good and it has test coverage.

From my perspective, the optimal solution would be to:

- Migrate the old implementation here
- Make it generic
- Clean up the code (e.g., implement specialized exceptions which already presented here)
- Maintain test suites for both variants

What are your thoughts on this approach?

the-yash-rajput:
@alxkm  Let’s first align on the right strategy and once confirmed, we can proceed with the necessary implementation.

alxkm:
@the-yash-rajput, you can proceed with the following approach as I described - creating a generic implementation of the algorithm.

- Migrate the existing implementation to the correct package location
- Refactor it to be generic and reusable
- Implement the specialized exceptions that are already present in the your version
- And generic tests for it with Integer and String values

This approach will provide the traditional algorithm that uses color values.

What do you think about this?

siriak:
@alxkm I agree with the proposed approach, I think both implementations should be stored together.

P.S. I reviewed this 2 days ago, but forgot to write a reply and then lost the tab -_-

DenizAltunkapan:
@the-yash-rajput @alxkm @siriak I also agree - and by the way sorry for the late reply... I was on vacation

the-yash-rajput:
@alxkm @siriak @DenizAltunkapan Okay, I am starting on this.

PR #6584, opened 2025-10-02
Feature iterative flood fill

CrodiYa:
#### Project already has Flood Fill
- This project already has a Flood Fill algorithm, but it is DFS.
For any big "image" DFS version won`t work because of stack overflow.
The BFS version, on the other hand, handles large "images" without any problems.

- My first "big" project on Java was simple pixel paint app and that was my first time encountering _StackOverflowException_ and I was very confused: it took me some time to figure out the problem and solution.
So, it would be useful to have this example of this algorithm here.

- A similar version of this algorithm can be found here: https://www.geeksforgeeks.org/dsa/flood-fill-algorithm/
Main difference - this version looks for diagonals too

#### Technical details

This version uses queue and helper class ```Point``` to traverse. 
While queue is not empty, the "pixel" is painted and "pixels" around it are added to queue (if coordinates are in boundaries and colour is not the same).
Directions for new "pixels" stored in arrays ```dx``` and ```dy```.  First four - horizontal and vertical, second four - diagonals.

It is achievable to store coordinates (x, y) in long, using bit mask, but it`s too much and possibly can lead to collision.

Tests were copied from original FloodFill and one test was added to demonstrate that BFS handles "big images". 

To be honest, I am not entirely sure in what package to add this, so "others" was picked.


src/main/java/com/thealgorithms/others/IterativeFloodFill.java
DenizAltunkapan, line 1: Please move the private methods below the public methods for better readability. Also, consider adding a class-level JavaDoc comment with a detailed description of the algorithm and its purpose. Maybe including a reference link (e.g., to a BFS Flood Fill article) would be very helpful for context. @CrodiYa

PR #6587, opened 2025-10-02
Added the Huffman Coding using java

PriyanshuPaul79:
The uploaded Java file implements the **Huffman Coding** algorithm, which is a **lossless data compression** technique.

---

## Core Components and Functionality

The code defines the Huffman Coding process, including building the compression tree, generating the binary codes, and performing both encoding and decoding.

### `HuffmanNode` Class
This inner class represents a node in the Huffman Tree.

* **Attributes**: Stores the `frequency` (how many times the character appears), the `character` itself (only for leaf nodes), and pointers to the `left` and `right` child nodes.
* **`compareTo` Method**: Implements the `Comparable` interface to allow a **PriorityQueue** to order nodes based on their **frequency**. Lower frequency nodes have higher priority for extraction.

#### 1. Building the Tree (`buildTree` method)
The main goal of this method is to construct the optimal binary tree for encoding.

1.  **Calculate Frequencies**: It first iterates through the input `text` to count the occurrences of every character, storing them in a `Map<Character, Integer>`.
2.  **Create Leaf Nodes**: Each character and its frequency are turned into a `HuffmanNode` (a leaf node), and all are added to a **PriorityQueue** (`pq`).
3.  **Merge Nodes**: It repeatedly extracts the **two nodes with the lowest frequency** from the priority queue. A new **internal node** is created, whose frequency is the sum of the two extracted nodes, and the two extracted nodes become its left and right children. This new internal node is then re-inserted into the priority queue.
    * This process continues until only one node remains in the queue, which is the **root** of the final Huffman Tree.
    * This greedy strategy ensures that less frequent characters end up deeper in the tree, receiving longer codes, while more frequent characters stay higher up, receiving shorter codes.
4.  **Generate Codes**: The final tree is passed to the `generateCodes` helper method.

#### 2. Generating Codes (`generateCodes` method)
This is a recursive method that performs a **traversal** (typically a Depth First Search) of the Huffman Tree:

* It assigns **'0'** for a step to the **left** child and **'1'** for a step to the **right** child.
* When a **leaf node** (a character) is reached, the path taken from the root (the accumulated binary string) is stored as the character's unique Huffman code in the `huffmanCodes` map.

#### 3. Encoding (`encode` method)
This method iterates through the input `text` and replaces each character with its corresponding binary code retrieved from the `huffmanCodes` map.

#### 4. Decoding (`decode` method)
This method reconstructs the original text from the binary string:

* It starts at the **root** of the Huffman Tree.
* For each bit in the encoded text:
    * If the bit is **'0'**, it moves to the **left** child.
    * If the bit is **'1'**, it moves to the **right** child.
* When a **leaf node** is reached, the character is appended to the decoded text, and the traversal restarts from the **root** for the next character.

---

## Example Execution (from `main` method)

The example uses the string: `"huffman coding is a lossless data compression algorithm"`

1.  **Frequencies** are calculated for each character.
2.  The **Huffman Tree** is built based on these frequencies.
3.  **Huffman Codes** are generated (e.g., 'f' might get a short code, while 'z' if present, would get a long code).
4.  The original text is **Encoded** into a binary string.
5.  The binary string is **Decoded** back to the original text.
6.  A **Verification** check confirms that `Original Text == Decoded Text`.

### DIFF
--- a/src/main/java/com/thealgorithms/greedyalgorithms/HuffmanCoding.java
+++ b/src/main/java/com/thealgorithms/greedyalgorithms/HuffmanCoding.java
@@ -17,12 +17,11 @@ class HuffmanNode implements Comparable<HuffmanNode> {
     }
 }
 
-public class HuffmanCoding {
+public final class HuffmanCoding {
 
     private static Map<Character, String> huffmanCodes = new HashMap<>();
     private static HuffmanNode root;
-    
-     private HuffmanCoding() {
+    private HuffmanCoding() {
         throw new UnsupportedOperationException("Utility class");
     }
     /**


DenizAltunkapan:
@PriyanshuPaul79 Please fix the failing workflows so I can start the review.

PR #6665, opened 2025-10-06
Feat/add postman sort

PauLopNun:


DenizAltunkapan:
@PauLopNun 
The current implementation is named `PostmanSort`, but the code actually implements a Shellsort-like algorithm with a gap factor of 1.3. While it correctly sorts the array, it does **not** implement the classical Postman Sort as described in the reference link, which is typically a digit- or bucket-based sorting algorithm (similar to counting sort or radix sort for postal codes).

As a result, the name `PostmanSort` is misleading. To address this, you could either:

1. **Rename the class** to `ShellSort` (or another appropriate name), since a Shellsort implementation already exists [here](https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/sorts/ShellSort.java).
2. **Implement a true Postman Sort**, following a digit/bucket-based approach.

Nevertheless, this PR cannot be merged, because the implementation does not match the algorithm name or description.

PR #6669, opened 2025-10-06
feat: added rotating calipers

yadhu-vipin:
Title: Fix Rotating Calipers: Correct Diameter and Width Calculations

Description:
This PR fixes the `RotatingCalipers` implementation for calculating the diameter and minimum width of convex polygons.

Changes made:

1. Updated the `diameter` method to correctly compute the farthest pair of points using the rotating calipers technique. Previous calculations returned incorrect distances for simple polygons.
2. Corrected the `width` method to compute the true minimum width of the convex polygon, based on perpendicular projections along each edge.
3. Updated the test cases for `testDiameterSimpleTriangle` and `testWidthSimpleTriangle` to match the mathematically correct expected values.
4. Ensured all other geometry calculations, including `minAreaBoundingRectangle`, remain consistent with correct convex hull ordering.

Impact:

Triangles and other simple polygons now return the correct diameter and width.
Unit tests pass with accurate geometric values.
References:

 Rotating Calipers algorithm (Shamos, 1978)
 Convex hull computation for ordered points

### DIFF
--- a/src/main/java/com/thealgorithms/geometry/RotatingCalipers.java
+++ b/src/main/java/com/thealgorithms/geometry/RotatingCalipers.java
@@ -1,6 +1,5 @@
 package com.thealgorithms.geometry;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


DenizAltunkapan:
@yadhu-vipin Please fix the failing workflows so I can start the review.

PR #6682, opened 2025-10-06
Implemented Queue using array

alka-tripathi:
1) add(int data)

Purpose: Add an element at the end of the queue.

Behavior:

If the queue is full (isOverflow() returns true), it prints a message and does not add the element.

Otherwise, increments rear and stores the new element in the array.

Time Complexity: O(1)

2) remove()

Purpose: Remove the element from the front of the queue.

Behavior:

If the queue is empty (isEmpty() returns true), it prints a message and returns -1.

Otherwise, it removes the front element, shifts all remaining elements one step forward, and decrements rear.

Time Complexity: O(n) — due to shifting elements.

 3) peek()

Purpose: Get the element at the front of the queue without removing it.

Behavior:

If the queue is empty, prints a message and returns -1.

Otherwise, returns the first element of the array.

Time Complexity: O(1)

4) isEmpty()

Purpose: Check whether the queue has any elements.

Behavior:

Returns true if rear == -1 (no elements in the queue), otherwise false.

Use Case: Helps prevent errors when removing or peeking from an empty queue.

5) isOverflow()

Purpose: Check whether the queue is full.

Behavior:

Returns true if rear == size - 1 (no more space to add elements), otherwise false.

Use Case: Helps prevent adding elements beyond the array size.

#5959 #queue #165

### DIFF
--- a/src/test/java/com/thealgorithms/datastructures/queues/ImplementQueue.java
+++ b/src/test/java/com/thealgorithms/datastructures/queues/ImplementQueue.java
@@ -1,0 +1,82 @@
+public class ImplementQueue {
+    // implementation of queue using array
+    // Queue Algorith type:FIFO (First Come First Out)
+
+    // create Queue class
+    public static class Queue {
+        int arr[];
+        int size;
+        int rear;
+
+        public Queue(int n) {
+            arr = new int[n];
+            size = n;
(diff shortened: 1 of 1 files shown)


DenizAltunkapan:
@alka-tripathi  Please fix the failing workflows so I can start the review.

PR #6761, opened 2025-10-12
Add bipartite graph algo using DFS

Ilmakram:
Added a Java implementation to check whether a graph is bipartite using DFS.
The program supports user input, handles disconnected graphs, and includes detailed JavaDocs.
Tested with sample graphs to ensure correctness.


DenizAltunkapan:
@Ilmakram  Please fix the failing workflows so I can start the review.

PR #6902, opened 2025-10-22
Added surface area calculation for pyramid - Aly Anany

alyanany:
Added a function that calculated the surface area for a Pyramid at the bottom of the Area.java file within the src code.


DenizAltunkapan:
solved in https://github.com/TheAlgorithms/Java/pull/6853

PR #6914, opened 2025-10-23
Add optimized Bubble Sort using last swap index

MuhammedHamzatkt:
This PR adds an optimized version of the Bubble Sort algorithm in Java.

Changes:
- Uses the last swap index to reduce unnecessary comparisons.
- Works on partially sorted arrays more efficiently.
- Added test case in main() to demonstrate sorting.

This contributes to the sorting algorithms collection by improving performance for nearly sorted arrays.


DenizAltunkapan:
@MuhammedHamzatkt Please change the existing file if there are any improvements

PR #7069, opened 2025-11-15
Optimize power() using exponentiation by squaring (O(log n)) 

Vinayak-v12:
This PR optimizes the power() method by replacing the linear recursive
implementation with an exponentiation-by-squaring approach.

✔ Time complexity reduced from O(n) → O(log n)
✔ Handles negative exponents
✔ Keeps same behavior as original implementation
✔ Fully compatible with existing test cases
✔ No API changes

This improves performance and reduces recursion depth while maintaining clean readability.


DenizAltunkapan:
whats the difference between your implementation and https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/divideandconquer/BinaryExponentiation.java ? @Vinayak-v12

PR #6617, opened 2025-10-04
Add FindMiddleOfSinglyLinkedList algorithm and tests

Hirannya-Mhaisbadwe:
- [x] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [x] This pull request is all my own work -- I have not plagiarized it.
- [ ] All filenames are in PascalCase.
- [x] All functions and variable names follow Java naming conventions.
- [ ] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.
- [ ] All new code is formatted with `clang-format -i --style=file path/to/your/file.java`

This PR adds an algorithm to find the middle of a singly linked list, along with unit tests.

Changes include:
- `FindMiddleOfSinglyLinkedList.java` — implements slow/fast pointer technique to find the middle.
- `FindMiddleOfSinglyLinkedListTest.java` — JUnit tests covering odd/even lists, single-node, two-node, empty list, and longer lists.

Motivation:
- Finding the middle of a linked list is a common operation for merge sort, palindrome checks, and other list manipulations.
- Adding this improves the completeness of the linked list module.

Fixes #6615

PR #7166, opened 2025-12-15
Add algorithm to find middle of singly linked list with tests

koushik-sai:
This PR adds an enhancement to the linked list data structures by:

- Implementing an algorithm to find the middle node of a singly linked list using the slow and fast pointer technique
- Documenting behavior for even-length lists (returns second middle)
- Adding unit tests covering empty, single-node, odd-length, and even-length lists

Fixes #6617


koushik-sai:
Thanks for reviewing this! 🙌  
After further inspection, I noticed that the middle-finding logic already exists in SinglyLinkedList and is already covered by unit tests. Since this enhancement is already addressed, I’m closing the PR to avoid duplication.

PR #7170, opened 2025-12-17
[FEAT] Add Thin Lens formula for ray optics

koushik-sai:
### Description
This pull request adds an implementation of the Thin Lens formula used in ray optics.

The class provides:
- Image distance calculation using the thin lens equation
- Magnification calculation
- Real vs virtual image determination

### Notes
- Uses Cartesian sign convention
- Includes input validation for physical correctness

### Reference
- https://en.wikipedia.org/wiki/Thin_lens


koushik-sai:
All CI checks are passing. 
Thank you for reviewing this contribution.

PR #7178, opened 2025-12-19
Add recursive factorial implementation with tests

koushik-sai:
- Added a recursive implementation of factorial in the recursion package
- Followed utility-class pattern used across the repository
- Included JUnit tests for base cases, positive input, and negative input

### DIFF
--- a/src/main/java/com/thealgorithms/sorts/QuickSort.java
+++ b/src/main/java/com/thealgorithms/sorts/QuickSort.java
@@ -1,24 +1,36 @@
 package com.thealgorithms.sorts;
 
 /**
- * @author Varun Upadhyay (https://github.com/varunu28)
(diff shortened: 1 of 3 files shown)


DenizAltunkapan:
there already exists a recursive implementation @koushik-sai https://github.com/TheAlgorithms/Java/blob/master/src/main/java/com/thealgorithms/maths/FactorialRecursion.java .
Would you suggest moving the existing class into the recursive package?
If so, please move it there and remove your implementation.

koushik-sai:
Thanks for the suggestion 👍
I’ve moved the existing FactorialRecursion from the maths package to the recursion package, removed the duplicate implementation introduced in this PR, updated the corresponding tests, and applied clang-format. All tests pass locally.

PR #7207, opened 2026-01-12
Handle negative input in perfect square checks

asmitha-16:
This PR improves the perfect square validation by:
- Returning false for negative input values
- Preventing invalid calculations in Math.sqrt and Math.pow


alxkm (approved):
Looks good. Thank you for the contribution.

PR #7221, opened 2026-01-19
Add BinarySearchStrings algorithm with comprehensive tests

JeevanYewale:

### DIFF
--- a/src/main/java/com/thealgorithms/searches/BinarySearchStrings.java
+++ b/src/main/java/com/thealgorithms/searches/BinarySearchStrings.java
@@ -7,6 +7,7 @@ package com.thealgorithms.searches;
  * Time Complexity: O(log n * m) where n is array length and m is average string length
  * Space Complexity: O(1)
  * 
+ * @see <a href="https://en.wikipedia.org/wiki/Binary_search_algorithm">Binary Search Algorithm</a>
  * @author Jeevan Yewale (https://github.com/JeevanYewale)
  */
 public final class BinarySearchStrings {


src/main/java/com/thealgorithms/searches/BinarySearchStrings.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Performs case-insensitive binary search on a sorted string array
+     * 
+     * @param array sorted array of strings (must be sorted in lexicographical order, case-insensitive)
+     * @param target the string to search for
+     * @return index of target string if found, -1 otherwise
+     */
+    public static int searchIgnoreCase(String[] array, String target) {
```
singhc7: The logic here is 100% identical to the search method above.

Please refactor this to use a private helper method that accepts a Comparator. That way you don't duplicate the binary search code twice.

singhc7 (commented):
The clang-format check is failing, probably just some whitespace or indentation issues. Run clang-format command on your files to auto-fix the style.

Also, left some comments on the implementation.

src/main/java/com/thealgorithms/searches/BinarySearchStrings.java
JeevanYewale: @singhc7 Thank you for the detailed review! You're absolutely right on all points:

1. ✅ **Memory optimization**: Replaced `targetLower.compareTo(array[mid].toLowerCase())` with `target.compareToIgnoreCase(array[mid])` to avoid unnecessary String object creation.

2. ✅ **Code deduplication**: Refactored both methods to use a private `binarySearch(String[] array, String target, Comparator<String> comparator)` helper method. Now `search()` uses `String::compareTo` and `searchIgnoreCase()` uses `String::compareToIgnoreCase`.

3. ✅ **Formatting**: Will run clang-format before the next push.

The refactored code is much cleaner and more efficient. Thanks for the guidance!

PR #7275, opened 2026-02-18
Add string algorithms: RemoveStars and ComplexNumberMultiply

SYEDMDSAAD:
### Feature Added

Added two string algorithms:

1. RemoveStars
Removes characters affected by '*' in a string.

Example:
Input: leet**cod*e
Output: lecoe

2. ComplexNumberMultiply
Multiplies two complex numbers represented as strings.

Example:
Input: 1+1i , 1+1i
Output: 0+2i

Both implementations follow repository coding standards and compile successfully.

### DIFF
--- a/spotbugs-exclude.xml
+++ b/spotbugs-exclude.xml
@@ -9,16 +9,10 @@
         <Bug pattern="DMI_RANDOM_USED_ONLY_ONCE" />
     </Match>
     <Match>
-        <Bug pattern="SF_SWITCH_NO_DEFAULT" />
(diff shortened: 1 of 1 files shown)


src/main/java/com/thealgorithms/strings/ComplexNumberMultiply.java
DenizAltunkapan: please move this algorithm and its corresponding test class to the math package. also please make sure to validate the input and cove invalid inputs in your tests cases

PR #7318, opened 2026-03-17
Add MaximumElement implementation with documentation

AbhiramSakha:
### 📌 Description
Added a new implementation to find the maximum element in an array.

### ✅ Changes Made
- Implemented `findMax` method using iteration
- Added proper JavaDoc comments
- Ensured clean and readable code

### 🧪 Testing
- Verified logic with sample inputs

### 📋 Checklist
- [x] I have read CONTRIBUTING.md
- [x] This is my own work (no plagiarism)
- [x] Code follows Java naming conventions
- [x] Code is properly formatted

### DIFF
--- a/src/main/java/com/thealgorithms/arrays/MaximumElement.java
+++ b/src/main/java/com/thealgorithms/arrays/MaximumElement.java
@@ -1,17 +1,21 @@
 package com.thealgorithms.arrays;
 
 /**
- * This class provides a method to find the maximum element in an array.
+ * Finds the maximum element in an array.
  */
-public class MaximumElement {
+public final class MaximumElement {
+
+    private MaximumElement() {
+        // utility class
+    }
 
     /**
-     * Finds the maximum value in the given array.
+     * Returns the maximum element in the array.
      *
-     * @param arr the input array
-     * @return the maximum element in the array
+     * @param arr input array
+     * @return maximum value
      */
-    public static int findMax(int[] arr) {
+    public static int max(int[] arr) {
         int max = Integer.MIN_VALUE;
 
         for (int num : arr) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/thealgorithms/arrays/MaximumElement.java
```diff
@@ -0,0 +1,29 @@
...
+        // utility class
+    }
+
+    /**
+     * Returns the maximum element in the array.
+     *
+     * @param arr input array
+     * @return maximum value
+     */
+    public static int max(int[] arr) {
```
singhc7: While it is always great to see new contributions, there are a few critical edge cases and efficiency improvements needed here for a robust utility class:

* The method does not validate against `null` inputs. Passing a `null` array will throw an unhandled `NullPointerException` when the enhanced for-loop attempts to execute.
* If an empty array (`new int[0]`) is passed, the loop is bypassed entirely and the method returns `Integer.MIN_VALUE`. Since an empty set technically has no maximum element, returning an arbitrary extreme integer can cause silent downstream bugs. This should throw an `IllegalArgumentException` instead.
* Initializing `max` to `Integer.MIN_VALUE` relies on a magic constant. A cleaner, more efficient algorithmic standard is to validate the array length, set `max = arr[0]`, and iterate starting from the second element.

singhc7 (commented):
Hey! Thanks for taking the time to put this together. 

While reviewing this, I noticed that we actually already have established implementations handling this exact logic in the repository (such as `FindMax.java`and `AbsoluteMax.java` in the maths package). 

To keep the codebase lean and avoid unnecessary redundancy, it might be best to withdraw this specific PR. If you are looking to contribute, I'd highly recommend checking out the open issues to see where help is needed most right now!

Also left some general suggestions for your code for future reference.

PR #7359, opened 2026-04-01
Add iterative implementation alongside recursive BinarySearch

Mudita-Singh:
🧩 Description
Added an iterative implementation of Binary Search alongside the existing recursive approach.

🚀 Changes Made
Implemented findIterative method using an iterative approach
Maintains consistency with existing generic structure (T extends Comparable<T>)
Uses bitwise shift (>>> 1) for midpoint calculation, consistent with existing implementation

✅ Benefits
Avoids recursion overhead
Uses constant space complexity O(1)
Helps beginners understand both approaches

🧪 Testing
Tested with sample inputs to ensure correctness.


Mudita-Singh:
Added an iterative implementation of Binary Search alongside the existing recursive approach.

PR #7420, opened 2026-05-12
NeonNumber algorithm in maths package

AshishKumar1297:
Added NeonNumber algorithm that checks if a 
number is a Neon Number.
Example: 9 → 9²=81 → 8+1=9 ✓

- [x] I have read CONTRIBUTING.md
- [x] This pull request is my own work
- [x] All filenames are in PascalCase
- [x] All functions and variable names follow Java naming conventions
- [x] All new algorithms have a URL in their comments
- [x] All new algorithms include a corresponding test class


src/main/java/com/thealgorithms/maths/NeonNumber.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    /**
+     * Check if a number is a Neon number.
+     *
+     * @param number the input number
+     * @return true if neon number, false otherwise
+     */
+    public static boolean isNeon(final int number) {
+        int square = number * number;
```
prashantpiyush1111, line 23: Integer overflow risk! For large inputs (e.g. number = 50000), 
`number * number` exceeds Integer.MAX_VALUE and gives wrong result.

Fix:
long square = (long) number * number;

src/main/java/com/thealgorithms/maths/NeonNumber.java
```diff
@@ -0,0 +1,32 @@
...
+        int digitSum = 0;
```
prashantpiyush1111, line 24: No validation for negative numbers. isNeon(-5) won't crash 
but behavior is unintended.

Fix: Add at the start of method:
if (number < 0) {
    throw new IllegalArgumentException("Input must be non-negative");
}

src/main/java/com/thealgorithms/maths/NeonNumber.java
```diff
@@ -0,0 +1,32 @@
...
+        int temp = square;
+        while (temp > 0) {
```
prashantpiyush1111, line 26: When number = 0, temp = 0, so this loop never runs.
isNeon(0) returns true only by coincidence.

Fix:
if (number == 0) return true;

src/test/java/com/thealgorithms/maths/NeonNumberTest.java
```diff
@@ -0,0 +1,23 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+public class NeonNumberTest {
+
+    @Test
+    public void testIsNeonTrue() {
```
prashantpiyush1111, line 11: assertTrue(NeonNumber.isNeon(0)) passes but for wrong reason —
loop never executes for 0. Consider adding explicit test 
that verifies the logic, not just the output.

src/test/java/com/thealgorithms/maths/NeonNumberTest.java
```diff
@@ -0,0 +1,23 @@
...
+        assertTrue(NeonNumber.isNeon(0));
+        assertTrue(NeonNumber.isNeon(1));
+        assertTrue(NeonNumber.isNeon(9));
+    }
+
+    @Test
+    public void testIsNeonFalse() {
```
prashantpiyush1111, line 18: Missing edge case tests:

1. Negative input:
assertThrows(IllegalArgumentException.class, 
    () -> NeonNumber.isNeon(-1));

2. Large number to catch overflow:
assertFalse(NeonNumber.isNeon(50000));

prashantpiyush1111 (commented):
Overall structure is good — final class, private constructor, 
and JavaDoc are all in place. 

However, there are a few issues that need to be fixed 
before merging:

1. Integer overflow in `isNeon()` method
2. No handling for negative inputs
3. `isNeon(0)` works by coincidence, not by logic
4. Missing edge case tests

Please address the inline comments.

prashantpiyush1111 (changes requested):
Comment box mein ye likho:

Please address the inline comments 
before this can be merged:

1. Integer overflow: use 
   long square = (long) number * number

2. Add negative input validation

3. Handle isNeon(0) explicitly

4. Add missing edge case tests

DenizAltunkapan:
Thanks for the contribution, @AshishKumar1297 ,  the algorithm itself is correct and the tests pass.

I'm going to close this PR, for two reasons:

- It accidentally adds an empty ashish-contribution file to the repository root, which shouldn't be part of any change.
- More importantly, a Neon number check is a trivial recreational-number property that overlaps heavily with algorithms we already have (Armstrong, AutomorphicNumber, HappyNumber). We're trying to avoid accumulating more one-off digit-property checks of this kind.

Nothing wrong with your code quality.
Please look into our [issues](https://github.com/TheAlgorithms/Java/issues) if you need other contribution ideas

PR #7499, opened 2026-06-27
Fix: Resolved string bounding bug in UI (Fixes #1234)

### LINKED ISSUE #1234: Can I add convex hull algorithms, some missing graph algos and hashing?
I would like to add the following algorithms:
- Convex hull:
  - Graham scan
  - Jarvis march
  - Chan's algorithm
  - Kirkpatrick-Seidel algorithm
- Johnson's shortest paths algorithm
- Vertex coloring
- HashMap variants using open addressing

Abhi13shek:
- [ ] All new algorithms include a corresponding test class that validates their functionality.
- [ ] All new code is formatted with `clang-format -i --style=file path/to/your/file.java`

### DIFF
--- a/src/main/java/com/thealgorithms/backtracking/NQueens.java
+++ b/src/main/java/com/thealgorithms/backtracking/NQueens.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.logging.Logger;
 
 /**
  * Problem statement: Given a N x N chess board. Return all arrangements in
@@ -41,6 +42,8 @@ import java.util.Set;
  */
 public final class NQueens {
 
+    private static final Logger LOGGER = Logger.getLogger(NQueens.class.getName());
+
     // Store occupied rows for constant time safety check
     private static final Set<Integer> OCROWS = new HashSet<>();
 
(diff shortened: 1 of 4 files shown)


prashantpiyush1111:
Hi @Abhi13shek , thanks for the contribution! I reviewed the diff and have a few observations:
1. Issue reference mismatch

The PR title says "Fixes #1234", but issue #1234 ("Can I add convex hull algorithms, some missing graph algos and hashing?") is a 2020 feature request that was auto-closed as stale in 2021. It has no connection to the changes in this PR — no convex hull, graph algorithm, or hashing code is touched here. Could you confirm the correct issue number this PR is meant to address?
2. Unrelated changes bundled together

This PR mixes two separate concerns:

NQueens.java: replacing System.out.println with Logger.info
StackPostfixNotation.java: refactoring Stack to ArrayDeque/Deque, plus try-with-resources for the Scanner

Per the contributing guidelines, it's usually cleaner to split unrelated changes into separate PRs — this makes review and rollback easier if something needs to be reverted later.
3. Test coverage gap

Codecov flags 4 missing lines in NQueens.java (63.6% patch coverage). It might be worth adding a test case that covers the logger branches (the "no solution found" and "arrangement found" paths) to close that gap.
4. Minor: empty log line

In NQueens.java, LOGGER.info("") (replacing the old blank System.out.println()) doesn't really add value — loggers print structured lines (timestamp/level), so an empty call just creates a noisy blank entry rather than visual spacing. Might be worth removing it.
cc @DenizAltunkapan  @alxkm  @yanglbme  — flagging in case this needs a second look before review, since the issue reference doesn't seem to match the actual changes.

DenizAltunkapan (changes requested):
I agree with @prashantpiyush1111 — the same points block this for me. Short summary of what to change:

1. **Fix the issue reference.** "Fixes #1234" points to an unrelated closed issue. Remove it or use the correct number.
2. **Split the PR.** NQueens (logger) and StackPostfixNotation (Deque) are unrelated and should be two PRs.
3. **Remove `LOGGER.info("")`.** An empty log line only adds noise.
4. **Add a test** for the NQueens logger branches to cover the 4 missing lines.

Requesting changes until these are addressed.

DenizAltunkapan:
The StackPostfixNotation change is good: java.util.Stack extends Vector and is effectively obsolete, ArrayDeque is the right replacement, and try-with-resources fixes a real leak on the exception path. I verified the push insertion-end difference doesn't change results. SudokuSolver is done well too.

One blocker:
LOGGER.info("");
in NQueens.java — as System.out.println() this was a bare newline, as a log call it emits a full record with timestamp and class name. Please drop it.

Nit, non-blocking: LOGGER.info("..." + queens + ...) builds the string even when the level is off; a supplier avoids that.
Fix the empty log line and the title/issue ref and I'll approve.

@Abhi13shek

PR #7505, opened 2026-07-01
feat: Add Bitmask DP TSP implementation

### LINKED ISSUE #7503: ADD BITWISE DP / SOS DP

Add implementation(s) of dynamic programming on bitmasks, where DP state includes a bitmask representing a subset of elements already used/visited (e.g. dp[mask][i]). Useful for problems like TSP, optimal assignment, and subset-based counting where n is small (≤ ~20) and the order/combination of included elements matters. Covers core bit tricks (set/check/clear bit, iterate submasks) along with example problems.

### Issue details

Bitmask DP represents a subset of elements as bits in an integer, where bit i being 1 means element i is included. The DP state dp[mask] (or dp[mask][i]) stores the best result for that subset, and transitions move from one mask to another by adding an element via mask | (1 << i). It's used when problems involve small sets (n ≤ ~20) and require tracking exactly which elements have been chosen, e.g., TSP, optimal assignment, and subset partitioning.

### Additional Information

_No response_

YashKumar-404:
Bitmask Dynamic Programming: Traveling Salesperson Problem (TSP)

Overview

This implementation demonstrates how to use Bitmask Dynamic Programming to solve the Traveling Salesperson Problem (TSP).

The Traveling Salesperson Problem asks: "Given a list of cities and the distances between each pair of cities, what is the shortest possible route that visits each city exactly once and returns to the origin city?"

What is Bitmask DP?

When solving optimization problems that involve subsets or combinations of elements (where the order/inclusion matters), tracking the "visited" state using standard arrays or objects can be highly inefficient.

Bitmask DP optimizes this by representing a subset of elements as bits in an integer:

If the $i$-th bit is 1, the $i$-th element is included in the subset.

If the $i$-th bit is 0, the $i$-th element is excluded.

Core Bitwise Operations Used:

Add an element to the subset: mask | (1 << i)

Check if an element is in the subset: (mask & (1 << i)) != 0

Because an integer is typically 32 bits, this technique is incredibly fast and memory-efficient but is generally restricted to problem sizes where $N \le 20$.

Algorithm Complexity

For a graph with $N$ vertices:

Time Complexity: $O(N^2 \cdot 2^N)$

There are $2^N$ possible subsets. For each subset, we iterate through $N$ currently visited nodes, and for each, we try to visit $N$ unvisited nodes. This is a massive improvement over the naive $O(N!)$ brute-force approach.

Space Complexity: $O(N \cdot 2^N)$

We use a 2D array dp[1 << N][N] to store the minimum cost for each subset and ending node.

Example

If we have 4 cities (0, 1, 2, 3), a subset containing cities 0, 2, and 3 is represented by the binary number 1101 (which is the integer 13).

The DP state dp[13][2] stores the minimum distance traveled to visit cities {0, 2, 3} where the last visited city is 2.

Transitions are made by checking unvisited cities (like city 1), adding them to the mask (1101 | 0010 = 1111), and updating the minimum cost.

How to Run & Test

If you have cloned the repository locally, you can run the JUnit tests for this specific implementation using Maven:

mvn test -Dtest=BitmaskDPTest

Resolves #7503


DenizAltunkapan (changes requested):
Thanks for the PR! The TSP logic looks correct, but a few things should be fixed before this can be merged:

1. **Add a test class.** New algorithms need a `BitmaskDPTest.java` under `src/test/...`. Right now there is no test, so the code is not covered.
2. **Remove `BitmaskDP_README.md` from `src/main/java`.** Source folders should only contain `.java` files. Please move the explanation into the class Javadoc instead.
3. **Return type / empty input.** `tsp` returns `int` and uses `Integer.MAX_VALUE`. For `n == 1` or disconnected costs the sum can overflow. Please guard against this (e.g. skip unreachable states) and document the expected input.
4. **Naming.** `distance` is fine, but please make the class name match the algorithm (e.g. keep `BitmaskDP` only if it stays generic; since it only does TSP, a clearer name or a note in the Javadoc would help).

Also note the repo already has `graph/TravelingSalesman.java` — please mention in the description why this DP version is a useful addition.

PR #7508, opened 2026-07-03
fix: Caesar cipher produces non-alphabetic output for negative shifts

### LINKED ISSUE #7506: Caesar cipher produces non-alphabetic garbage output for negative shifts
### Description

`Caesar.encode` and `Caesar.decode` in `src/main/java/com/thealgorithms/ciphers/Caesar.java` produce incorrect, non-alphabetic output when called with a **negative shift** value. The `encode(decode(x)) == x` round-trip is also broken for negative shifts in some cases.

### Root cause

`normalizeShift` computes the shift as:

```java
private static char normalizeShift(final int shift) {
    return (char) (shift % 26);
}
```

In Java, `%` preserves the sign of the dividend, so for a negative shift that is not an exact multiple of 26 (e.g. `-1`, `-2`, `-5`), `shift % 26` is itself negative (e.g. `-1 % 26 == -1`). Casting that negative `int` directly to `char` does **not** produce a small negative number — `char` is an unsigned 16-bit type in Java, so `(char) -1` wraps around to `65535` (`0xFFFF`).

That huge value is then added to the letter's `char` code (`current += shiftChar`), overflows the 16-bit `char` range, and truncates back into a seemingly "normal" but wrong character. The subsequent bounds check (`current > 'Z'` / `current < 'A'`) does not catch this because the truncated value can land anywhere in the `char` range, not just outside the alphabet.

### Steps to reproduce

```java
Caesar caesar = new Caesar();
System.out.println(caesar.encode("A", -1));
```

**Expected:** `"Z"` (a Caesar shift of -1 should wrap `'A'` back to `'Z'`, just like a shift of +25 would).

**Actual:** `"@"` (ASCII 64) — a non-alphabetic character is silently emitted.

Verified locally by extracting the exact `normalizeShift`/`encode`/`decode` logic and running it:

```
shift=-1   encode("A") = "@"   (char code 64, not a letter)
shift=-2   encode("A") = "?"   (char code 63)
shift=-5   encode("A") = "<"   (char code 60)
shift=-25  encode("A") = "("   (char code 40)
```

The round-trip is also broken for edge cases:

```java
String encoded = caesar.encode("A", -1); // "@"
String decoded = caesar.decode(encoded, -1); // "@" (not "A")
```

### Why this matters

- This isn't just an edge case — **any negative shift** that isn't an exact multiple of 26 triggers it (e.g. `-1`, `-2`, ..., `-25`, `-27`, etc.), and negative shifts are a normal, documented way to use a Caesar cipher (shifting backward).
- The bug is completely uncovered by the existing tests: `CaesarTest.java` only exercises `shift = 5` and the `bruteforce()` method (which internally only iterates shifts `0..26`), so no negative shift is ever tested.
- The failure mode is silent — it doesn't throw an exception, it just returns wrong/corrupted data, which is worse for a class users might trust for correctness.

### Suggested fix

Normalize the shift as a plain `int` using a formula that is always non-negative, instead of casting a possibly-negative `int` to `char`:

```java
private static int normalizeShift(final int shift) {
    return ((shift % 26) + 26) % 26;
}
```

and use `int shiftChar` in `encode`/`decode` instead of `char shiftChar`. Happy to submit a PR with this fix plus a test case covering negative shifts if that's welcome.

herley-shaori:
Fixes #7506

### Problem

`Caesar.encode`/`Caesar.decode` produced corrupted, non-alphabetic output for any negative shift that is not an exact multiple of 26:

```java
new Caesar().encode("A", -1); // returned "@" (ASCII 64) instead of "Z"
```

Root cause: `normalizeShift` did `(char) (shift % 26)`. Java's `%` preserves the dividend's sign, so `-1 % 26 == -1`, and casting a negative `int` to the unsigned 16-bit `char` type wraps it to `65535`. The following char arithmetic then overflowed and truncated into an arbitrary wrong character, which the `> 'Z'` / `< 'A'` wraparound checks could not catch. This also broke the `decode(encode(x, s), s) == x` round-trip for negative shifts.

### Fix

- `normalizeShift` now returns an `int` normalized into `0..25` via `((shift % 26) + 26) % 26`, so a negative shift is mapped to its equivalent positive rotation (e.g. `-1` → `25`).
- The shift arithmetic in `encode`/`decode` is performed in `int` and cast back to `char` only when appending, which also removes the `lossy-conversions` compound-assignment pattern.

### Tests

Added three regression tests to `CaesarTest`:

- `caesarEncryptWithNegativeShiftTest` — `encode("A", -1)` must wrap to `"Z"` (upper and lower case).
- `caesarDecryptWithNegativeShiftTest` — decoding with a negative shift.
- `caesarNegativeShiftRoundTripTest` — encode→decode round-trip for shifts `-1, -5, -25, -26, -27, -52`.

Verified locally: `mvn test -Dtest=CaesarTest` → 6/6 pass, `mvn checkstyle:check` clean, `clang-format --dry-run --Werror` clean. All existing tests still pass unchanged.

### DIFF
--- a/src/main/java/com/thealgorithms/ciphers/Caesar.java
+++ b/src/main/java/com/thealgorithms/ciphers/Caesar.java
@@ -9,8 +9,8 @@ package com.thealgorithms.ciphers;
  * @author khalil2535
  */
 public class Caesar {
-    private static char normalizeShift(final int shift) {
-        return (char) (shift % 26);
+    private static int normalizeShift(final int shift) {
+        return ((shift % 26) + 26) % 26;
     }
 
     /**
@@ -22,21 +22,18 @@ public class Caesar {
     public String encode(String message, int shift) {
         StringBuilder encoded = new StringBuilder();
 
-        final char shiftChar = normalizeShift(shift);
+        final int shiftChar = normalizeShift(shift);
 
         final int length = message.length();
         for (int i = 0; i < length; i++) {
-            //            int current = message.charAt(i); //using char to shift characters because
-            //            ascii
-            // is in-order latin alphabet
-            char current = message.charAt(i); // Java law : char + int = char
+            final char current = message.charAt(i);
 
             if (isCapitalLatinLetter(current)) {
-                current += shiftChar;
-                encoded.append((char) (current > 'Z' ? current - 26 : current)); // 26 = number of latin letters
+                final int shifted = current + shiftChar;
+                encoded.append((char) (shifted > 'Z' ? shifted - 26 : shifted)); // 26 = number of latin letters
             } else if (isSmallLatinLetter(current)) {
-                current += shiftChar;
-                encoded.append((char) (current > 'z' ? current - 26 : current)); // 26 = number of latin letters
+                final int shifted = current + shiftChar;
+                encoded.append((char) (shifted > 'z' ? shifted - 26 : shifted)); // 26 = number of latin letters
             } else {
                 encoded.append(current);
             }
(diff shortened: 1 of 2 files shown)


DenizAltunkapan (approved):
@herley-shaori thanks!

PR #7509, opened 2026-07-03
docs: correct stale WiggleSort Javadoc — [1, 2, 2] is already detected, add regression tests

### LINKED ISSUE #7507: WiggleSort Javadoc example [1, 2, 2] is stale — that case is already caught, not silently mishandled
### Description

The class-level Javadoc of `WiggleSort` (`src/main/java/com/thealgorithms/sorts/WiggleSort.java`) states that the algorithm cannot detect every non-wiggle-sortable input, and gives `[1, 2, 2]` as a concrete example that would slip through undetected:

```java
/**
 * ...
 * Not all arrays are wiggle-sortable. This algorithm will find some obviously not wiggle-sortable
 * arrays and throw an error, but there are some exceptions that won't be caught, for example [1, 2,
 * 2].
 */
```

This claim is **outdated**. Elsewhere in the same file, a later fix already added a guard specifically for this scenario:

```java
// added condition preventing off-by-one errors for odd arrays.
// https://cs.stackexchange.com/questions/150886/how-to-find-wiggle-sortable-arrays-did-i-misunderstand-john-l-s-answer?noredirect=1&lq=1
if (sortThis.length % 2 == 1 && numMedians == ceil(sortThis.length / 2.0)) {
    T smallestValue = select(Arrays.asList(sortThis), 0);
    if (!(0 == smallestValue.compareTo(median))) {
        throw new IllegalArgumentException("For odd Arrays if the median appears ceil(n/2) times, "
            + "the median has to be the smallest values in the array.");
    }
}
```

This guard was apparently added to fix the exact `[1, 2, 2]` case the class Javadoc still references as an unhandled exception, but the Javadoc comment describing the algorithm's limitations was never updated to reflect the fix.

### Steps to reproduce / verify

Running `new WiggleSort().sort(new Integer[]{1, 2, 2})` today does **not** silently return a bad result as the Javadoc implies — it correctly throws:

```
IllegalArgumentException: For odd Arrays if the median appears ceil(n/2) times, the median has to be the smallest values in the array.
```

I verified this locally by compiling `WiggleSort.java` together with its dependencies (`SortAlgorithm`, `SortUtils`, `Ceil`, `Floor`, `QuickSelect`) straight from this repo's `src/main/java` and running:

1. The exact `[1, 2, 2]` example from the Javadoc → correctly throws the exception above (not silently wrong output).
2. A brute-force sweep of thousands of small arrays (lengths 2–7, with repeated/duplicate values) → 0 cases were found where the algorithm produced a non-wiggle-sorted result without throwing.

So, as far as I can verify, the algorithm's detection is currently more complete than the Javadoc claims — at least the specific counter-example given no longer applies.

### Why this matters

- Misleading documentation on a correctness-sensitive algorithm can cause contributors or users to distrust (or mistakenly "fix") code that is actually already correct for the case described, or to skip writing a regression test for `[1, 2, 2]` because the docs say it "won't be caught" anyway.
- A quick grep of `src/test/java/com/thealgorithms/sorts/WiggleSortTest.java` shows there is currently no test asserting that `[1, 2, 2]` throws — worth adding one to lock in the current (correct) behavior and prevent future regressions.

### Suggested fix

- Update the class Javadoc to remove or correct the `[1, 2, 2]` example, since it is now caught.
- If there are still genuinely undetectable non-wiggle-sortable inputs, replace the example with a real one (or explicitly state that no counter-example is currently known) to avoid future confusion.
- Optionally, add a unit test asserting `IllegalArgumentException` for `[1, 2, 2]` to guard against regressions.

Happy to submit a PR with the Javadoc correction and a regression test if that's welcome.

herley-shaori:
Fixes #7507

### Problem

The class-level Javadoc of `WiggleSort` claims:

> "there are some exceptions that won't be caught, for example [1, 2, 2]"

This is outdated. The odd-array median guard inside `wiggleSort()` (the `sortThis.length % 2 == 1 && numMedians == ceil(sortThis.length / 2.0)` check, added with a reference to [this cs.stackexchange question](https://cs.stackexchange.com/questions/150886/)) already detects exactly that case and throws `IllegalArgumentException` — verified by running `new WiggleSort().sort(new Integer[]{1, 2, 2})` locally, plus a brute-force sweep over thousands of small arrays (lengths 2–7 with duplicates) that found no case where a non-wiggle-sorted result is silently returned.

The stale comment is misleading: it tells readers the algorithm is known-broken for an input it actually handles, and it discouraged anyone from adding a regression test for that input ("won't be caught anyway").

### Changes

- Rewrote the class Javadoc to describe the current behavior: non-wiggle-sortable inputs such as `[1, 2, 2]` are detected and rejected with `IllegalArgumentException`.
- Added two regression tests locking in that behavior:
  - `wiggleTestNonWiggleSortableOddArrayThrows` — `[1, 2, 2]` must throw (median appears ceil(n/2) times but is not the smallest value).
  - `wiggleTestTooManyDuplicatesThrows` — `[2, 2, 2, 1]` must throw (more than half the values are equal).

No behavior change to the algorithm itself — documentation and tests only.

### Verification

`mvn test -Dtest=WiggleSortTest` → 9/9 pass (7 existing + 2 new), `mvn checkstyle:check` clean, `clang-format --dry-run --Werror` clean.

### DIFF
--- a/src/main/java/com/thealgorithms/sorts/WiggleSort.java
+++ b/src/main/java/com/thealgorithms/sorts/WiggleSort.java
@@ -11,9 +11,9 @@ import java.util.Arrays;
  * https://cs.stackexchange.com/questions/125372/how-to-wiggle-sort-an-array-in-linear-time-complexity
  * Also have a look at:
  * https://cs.stackexchange.com/questions/125372/how-to-wiggle-sort-an-array-in-linear-time-complexity?noredirect=1&lq=1
- * Not all arrays are wiggle-sortable. This algorithm will find some obviously not wiggle-sortable
- * arrays and throw an error, but there are some exceptions that won't be caught, for example [1, 2,
- * 2].
+ * Not all arrays are wiggle-sortable. This algorithm detects non-wiggle-sortable inputs — for
+ * example [1, 2, 2], or arrays where more than half the values are equal — and throws an
+ * IllegalArgumentException instead of returning a wrongly ordered result.
  */
 public class WiggleSort implements SortAlgorithm {
 
--- a/src/test/java/com/thealgorithms/sorts/WiggleSortTest.java
+++ b/src/test/java/com/thealgorithms/sorts/WiggleSortTest.java
@@ -1,6 +1,7 @@
 package com.thealgorithms.sorts;
 
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import java.util.Arrays;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


Mahi1215:
hlo,can i work on this

DenizAltunkapan (approved):
@herley-shaori thank you for your contribution

PR #7514, opened 2026-07-05
Refactor/smell free solid clean code

nabila1644-iit:
## Title
Refactor BronKerbosch.java to resolve SRP/OCP/DIP violations and Bloater/Coupler smells

## What issues were addressed
- Long Method: `findMaximalCliques` mixed validation, graph copying, and algorithm init in one method
- Feature Envy: `bronKerboschPivot` and helpers were static utilities not tied to any object
- Primitive Obsession: raw `List<Set<Integer>>` used instead of a typed graph abstraction
- Data Clumps: `r`, `p`, `x` always traveled together without being packaged as a unit
- SRP/OCP/DIP violations: one class handled validation, construction, pivot logic, and enumeration; pivot strategy was hardcoded; logic depended on concrete `HashSet`/`ArrayList` types

## Why these changes improve the system
- Splitting responsibilities (Graph, GraphValidator, PivotStrategy, SearchState, Solver, façade) means each class now has exactly one reason to change
- Pivot selection is now swappable via `PivotStrategy` without touching the algorithm (OCP)
- The solver depends on the `Graph` interface, not concrete collection types (DIP)
- `SearchState` replaces the r/p/x data clump with an immutable, self-validating value object

## Summary of refactoring approach
- Extracted a `Graph` interface + `AdjacencyListGraph` implementation to decouple the algorithm from raw collections
- Extracted `GraphValidator` to isolate input-validation logic
- Introduced `PivotStrategy` interface with `MaxDegreePivotStrategy` as the default implementation
- Replaced mutable shared state with an immutable `SearchState` value object, deriving new states via `expandWith`/`processVertex` instead of in-place mutation
- Kept the original public API (`BronKerbosch.findMaximalCliques(List<Set<Integer>>)`) unchanged as a façade over the new internal design

## Functionality assurance
- Public method signature, return type, and algorithm logic (Bron–Kerbosch with pivoting) are unchanged
- Recursive mutation sequence of the original r/p/x sets was manually traced against the new immutable state transitions to confirm identical clique enumeration
- No test suite exists for this class in the repository (educational reference project), so verification was done via manual trace rather than automated tests — noted transparently for reviewers


DenizAltunkapan:
too many changes

PR #7532, opened 2026-07-20
Add MultinomialNaiveBayesClassifier Implementation

Rosander0:
Add Multinomial Naive Bayes classifier

Adds a **MultinomialNaiveBayesClassifier** implementation to **com.thealgorithms.machinelearning**, following the same package structure as the existing Linear Regression implementation.

What's included:

Multinomial Naive Bayes with Laplace (add-alpha) smoothing, configurable via constructor (defaults to alpha = 1.0)
Log-probability arithmetic throughout predict() to avoid numerical underflow from multiplying small probabilities directly
Single-sample and batch predict() overloads
Guard clauses for unfitted-classifier use, invalid alpha, and feature/sample dimension mismatches
JUnit 5 test suite covering prediction correctness on a separable toy dataset, batch/single-sample consistency, smoothing behavior on zero-count features, and all exception paths

Notes for reviewers:

This is the Multinomial variant specifically (suited to count-based features like word frequencies), not Gaussian or Bernoulli — happy to add those as follow-ups if there's interest
One test (laplaceSmoothingKeepsZeroCountFeatureLogProbabilityFinite) only asserts the prediction doesn't collapse to an invalid state, not the exact expected class — flagging in case a maintainer wants that tightened

### DIFF
--- a/src/main/java/com/thealgorithms/machinelearning/MultinomialNaiveBayesClassifier.java
+++ b/src/main/java/com/thealgorithms/machinelearning/MultinomialNaiveBayesClassifier.java
@@ -1,0 +1,141 @@
+package com.thealgorithms.machinelearning;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Multinomial Naive Bayes classifier.
+ *
+ * <p>Suited to discrete, count-based features (e.g. word frequencies in text
+ * classification). Class priors and feature likelihoods are estimated from
+ * training data with Laplace (add-alpha) smoothing to avoid zero
+ * probabilities for unseen feature/class combinations. Predictions are made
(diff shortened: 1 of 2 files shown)


Rosander0:
Hey @DenizAltunkapan , just checking in on this PR — it's been a few days. Happy to make any changes if something needs to be adjusted. Let me know!

DenizAltunkapan:
@Rosander0 Thanks for the contribution. The implementation looks solid overall, but `fit()` should clear `logPriors` and `logLikelihoods` before training. Otherwise, calling `fit()` a second time can leave classes from the previous training run in the model and affect later predictions.

Please also add a regression test for refitting the same classifier instance. After that, this should be ready to approve.

DenizAltunkapan (approved):
@Rosander0 thanks!

PR #7536, opened 2026-07-21
Fix incorrect absolute minimum calculation

klerk1t:
- [X] I have read [CONTRIBUTING.md](https://github.com/TheAlgorithms/Java/blob/master/CONTRIBUTING.md).
- [X] This pull request is all my own work -- I have not plagiarized it.
- [X] All filenames are in PascalCase.
- [X] All functions and variable names follow Java naming conventions.
- [X] All new algorithms have a URL in their comments that points to Wikipedia or other similar explanations.
- [X] All new algorithms include a corresponding test class that validates their functionality.
- [X] All new code is formatted with `clang-format -i --style=file path/to/your/file.java`

### DIFF
--- a/src/main/java/com/thealgorithms/maths/AbsoluteMin.java
+++ b/src/main/java/com/thealgorithms/maths/AbsoluteMin.java
@@ -1,7 +1,5 @@
 package com.thealgorithms.maths;
 
-import java.util.Arrays;
-
 public final class AbsoluteMin {
     private AbsoluteMin() {
     }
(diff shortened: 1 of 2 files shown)


DenizAltunkapan:
@klerk1t Nice simplification! I found one remaining edge case though.

`Math.abs(Integer.MIN_VALUE)` overflows and returns `Integer.MIN_VALUE`, so cases like:

```java
AbsoluteMin.getMinValue(Integer.MIN_VALUE, 1)
````

currently return `Integer.MIN_VALUE` instead of `1`.

I'd recommend comparing absolute values as `long` (e.g. `Math.abs((long) value)`) and adding a test covering `Integer.MIN_VALUE`.

klerk1t:
@DenizAltunkapan Thanks for pointing this out! I've updated the implementation to avoid the `Integer.MIN_VALUE` overflow issue by performing the comparisons using `long `values. I also added tests covering this case.

DenizAltunkapan (approved):
@klerk1t thank you for your contribution 🙂

PR #7540, opened 2026-07-23
Fix AbsoluteMin comparison by absolute magnitude

### LINKED ISSUE #7537: [BUG] <title>
### Description

The current implementation of com.thealgorithms.maths.AbsoluteMin returns an incorrect result for some inputs.

Example:

org.opentest4j.AssertionFailedError:
Expected :2
Actual :-5

The implementation compares values by their absolute magnitude but updates the result using Math.min(), which can produce an incorrect result.

Steps to reproduce
Add the following test to AbsoluteMinTest:

assertEquals(2, AbsoluteMin.getMinValue(-5, 2));

Excepted behavior
The method should return the value with the smallest absolute magnitude.

For the input {-5, 2}, the expected result is 2.

### Steps to reproduce

_No response_

### Excepted behavior

The method AbsoluteMin.getMinValue(int a, int b) is expected to return the number whose absolute value is smallest, not the numerically smallest number.

It should compare values using Math.abs()
It should return the original number (preserving sign) that has the minimum absolute magnitude

### Screenshots

_No response_

### Additional context

_No response_

kush2439p:
## Summary

- compare candidates by absolute magnitude and return the original signed value
- preserve the existing deterministic tie behavior by choosing the numerically smaller value
- add coverage for mixed signs, both-negative, both-positive, equal magnitudes, and zero

## Root cause

`AbsoluteMin.getMinValue` filtered candidates using their absolute values, but then updated the result with `Math.min` on the raw signed values. A larger-magnitude negative number could therefore replace the correct candidate.

## Validation

- `javac` compilation and direct behavior checks passed
- `git diff --check` passed
- `mvn -Dtest=AbsoluteMinTest test` reached test compilation, but the upstream checkout has unrelated missing-class compilation failures across existing tests before Surefire can select the focused test

Fixes #7537

### DIFF
--- a/src/main/java/com/thealgorithms/maths/AbsoluteMin.java
+++ b/src/main/java/com/thealgorithms/maths/AbsoluteMin.java
@@ -19,8 +19,7 @@ public final class AbsoluteMin {
         for (int number : numbers) {
             long absoluteNumber = Math.abs((long) number);
             long absoluteMinValue = Math.abs((long) minValue);
-            if (absoluteNumber < absoluteMinValue
-                || (absoluteNumber == absoluteMinValue && number < minValue)) {
+            if (absoluteNumber < absoluteMinValue || (absoluteNumber == absoluteMinValue && number < minValue)) {
                 // For equal absolute values, consistently choose the numerically smaller value.
                 minValue = number;
             }
--- a/src/main/java/com/thealgorithms/maths/AbsoluteMin.java
+++ b/src/main/java/com/thealgorithms/maths/AbsoluteMin.java
@@ -1,7 +1,5 @@
 package com.thealgorithms.maths;
 
-import java.util.Arrays;
-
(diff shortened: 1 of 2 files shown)


kush2439p:
This PR corrects AbsoluteMin.getMinValue to compare values by absolute magnitude while returning the original signed value.
It also documents deterministic tie handling (the numerically smaller value is chosen) and adds tests for mixed signs, negative values, positive values, ties, and zero.
Validation:
Checkstyle passed
Direct behavior checks passed
Full Maven test compilation is currently blocked by unrelated missing classes in the upstream checkout.

kush2439p:
Closing as superseded by #7536, which has already been merged into master and covers the same absolute-magnitude fix while also handling null input and Integer.MIN_VALUE overflow. Thank you.

PR #7551, opened 2026-08-02
fix: reject null array in BitonicSort with IllegalArgumentException

### LINKED ISSUE #7549: BitonicSort should reject a null array explicitly
BitonicSort currently dereferences a null input while checking its length, producing an incidental NullPointerException. The proposed change validates the input first, raises a descriptive IllegalArgumentException, and adds a focused regression test. This issue accompanies a Matcha study recommendation derived from https://stackoverflow.com/a/64124615.

zeeshanzzz788:
### Description
`BitonicSort.sort` previously dereferenced a null array when reading `array.length`, producing an incidental `NullPointerException`.

Validate null input first, throw a clear `IllegalArgumentException`, and add a regression test.

Fixes #7549


DenizAltunkapan:
already solved

PR #7555, opened 2026-08-05
Fix infinite loop in JumpSearch when key exceeds last element

SEPURI-SAI-KRISHNA:
`JumpSearch.find` never terminates when the key is greater than the last element of the array.

The jumping loop clamps its cursor to the last index:

```java
while (limit < length && key.compareTo(array[limit]) > 0) {
    limit = Math.min(limit + blockSize, length - 1);   // <-- stops advancing
}
```

Once `limit` reaches `length - 1`, `Math.min(limit + blockSize, length - 1)` evaluates back to `length - 1`, so the cursor stops moving. If the key is still greater than `array[length - 1]`, the loop condition stays true forever and the call hangs, burning CPU.

Reproducer:

```java
new JumpSearch().find(new Integer[] {1, 2, 3, 4}, 5); // never returns
```

The clamp is unnecessary — the loop condition `limit < length` already stops the scan, and the linear-search loop that follows is bounded by `i <= limit && i < length`. Letting `limit` run past the end restores termination without changing any other behaviour.

### Fix

`limit = Math.min(limit + blockSize, length - 1);` → `limit += blockSize;`

### Tests

- `testJumpSearchKeyGreaterThanLastElement` — the minimal reproducer, with `@Timeout(threadMode = SEPARATE_THREAD)` so the hang surfaces as a failure instead of stalling CI.
- `testJumpSearchKeyGreaterThanLastElementForEveryLength` — same, for lengths 1–50, since the block size depends on the array length.
- `testJumpSearchFindsEveryElement` — every index of every array of length 1–50 is still found, covering the elements that sit exactly on a jump boundary.

All three fail (time out) on `master` and pass with the fix. The existing tests never used a key above the maximum, which is why this went unnoticed.

### DIFF
--- a/src/main/java/com/thealgorithms/searches/JumpSearch.java
+++ b/src/main/java/com/thealgorithms/searches/JumpSearch.java
@@ -73,7 +73,7 @@ public class JumpSearch implements SearchAlgorithm {
         int limit = blockSize;
         // Jumping ahead to find the block where the key may be located
         while (limit < length && key.compareTo(array[limit]) > 0) {
-            limit = Math.min(limit + blockSize, length - 1);
+            limit += blockSize;
         }
 
         // Perform linear search within the identified block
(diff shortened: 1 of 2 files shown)


alxkm (approved):
Looks good. Thank you for the contribution.

PR #7556, opened 2026-08-05
Fix ExponentialSearch missing boundary elements and not returning -1

SEPURI-SAI-KRISHNA:
`ExponentialSearch.find` has two defects.

**1. It reports present elements as missing.**

After the doubling loop the candidate block is the *inclusive* index range `[range / 2, range]`, but `Arrays.binarySearch` takes an *exclusive* `toIndex`:

```java
return Arrays.binarySearch(array, range / 2, Math.min(range, array.length), key);
```

Index `range` — the element that stopped the doubling loop — is therefore never examined, so any key that lands exactly on a doubling boundary is not found:

```java
new ExponentialSearch().find(new Integer[] {-25, -9, 8, 21}, 8); // -3, expected 2
```

Randomised testing over sorted `Integer` arrays of length 1–12 hits this on roughly 4% of present keys.

**2. A missing key does not return -1.**

The Javadoc promises "The index of the key if found, otherwise -1", but the raw result of `Arrays.binarySearch` is returned, which is `-(insertion point) - 1` when absent:

```java
new ExponentialSearch().find(new Integer[] {1, 3, 5, 7, 9, 11}, 4); // -3, expected -1
```

This also breaks the `SearchAlgorithm` contract that every other implementation in the package honours. The existing "not found" test only used an empty array, which returns `-1` early and so hid the problem.

### Fix

```java
final int index = Arrays.binarySearch(array, range / 2, Math.min(range + 1, array.length), key);
return index >= 0 ? index : -1;
```

### Tests

- `testExponentialSearchElementOnRangeBoundary` — the minimal reproducer for the off-by-one.
- `testExponentialSearchFindsEveryElement` — every index of every array of length 1–50.
- `testExponentialSearchNotFoundReturnsMinusOne` — absent keys below, inside and above the array range.

All three fail on `master` and pass with the fix.

### DIFF
--- a/src/main/java/com/thealgorithms/searches/ExponentialSearch.java
+++ b/src/main/java/com/thealgorithms/searches/ExponentialSearch.java
@@ -46,6 +46,9 @@ class ExponentialSearch implements SearchAlgorithm {
             range = range * 2;
         }
 
-        return Arrays.binarySearch(array, range / 2, Math.min(range, array.length), key);
+        // The candidate block is the inclusive index range [range / 2, range], so the
+        // exclusive upper bound handed to binarySearch has to be range + 1.
+        final int index = Arrays.binarySearch(array, range / 2, Math.min(range + 1, array.length), key);
+        return index >= 0 ? index : -1;
     }
 }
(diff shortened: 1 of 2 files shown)


DenizAltunkapan (approved):
@SEPURI-SAI-KRISHNA looks clean. Thank you for your contribution

PR #7557, opened 2026-08-05
Fix out-of-bounds access and identity comparison in FibonacciSearch

SEPURI-SAI-KRISHNA:
The final probe in `FibonacciSearch.find` has two defects:

```java
if (fibMinus1 == 1 && array[offset + 1] == key) {
    return offset + 1;
}
```

**1. Out-of-bounds read.** When the key is greater than every element, the main loop advances `offset` all the way to `n - 1`, so `array[offset + 1]` reads past the end:

```java
new FibonacciSearch().find(new Integer[] {10, 20, 30, 40}, 50);
// java.lang.ArrayIndexOutOfBoundsException: Index 4 out of bounds for length 4
```

The method is documented to return `-1` when the key is absent. Randomised testing over sorted `Integer` arrays of length 1–12 throws on roughly 6% of inputs.

**2. Reference equality instead of value comparison.** `==` compares object identity, not the values. It happens to work for small `Integer` values because of the `Integer` cache (−128..127), which is why the existing tests pass, but it fails for anything else:

```java
new FibonacciSearch().find(new Integer[] {10, 20, 300}, 300);   // -1, expected 2
new FibonacciSearch().find(new String[] {"a", "b", "c"}, key);  // -1 for a non-interned "c"
```

Since `T extends Comparable<T>`, `compareTo` is the right comparison here and is consistent with the rest of the method.

### Fix

```java
if (fibMinus1 == 1 && offset + 1 < n && array[offset + 1].compareTo(key) == 0) {
    return offset + 1;
}
```

### Tests

- `testFibonacciSearchKeyGreaterThanLastElement` — a key above the maximum for lengths 1–50 must return `-1`, not throw.
- `testFibonacciSearchFindsEqualButNotIdenticalKey` — an `Integer` outside the cache and a non-interned `String`.
- `testFibonacciSearchFindsEveryElement` — every index of every array of length 1–50, with values above the `Integer` cache.

All three fail on `master` and pass with the fix.

### DIFF
--- a/src/main/java/com/thealgorithms/searches/FibonacciSearch.java
+++ b/src/main/java/com/thealgorithms/searches/FibonacciSearch.java
@@ -69,7 +69,7 @@ public class FibonacciSearch implements SearchAlgorithm {
             }
         }
 
-        if (fibMinus1 == 1 && array[offset + 1] == key) {
+        if (fibMinus1 == 1 && offset + 1 < n && array[offset + 1].compareTo(key) == 0) {
             return offset + 1;
         }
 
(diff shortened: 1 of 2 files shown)


DenizAltunkapan (approved):
@SEPURI-SAI-KRISHNA thank you for your contribution

PR #7572, opened 2026-08-15
fix: RailFenceCipher silently drops newline characters

SEPURI-SAI-KRISHNA:
### Problem

`RailFenceCipher.encrypt` builds a `rails x length` matrix and fills every cell with `'\n'` to mark it as unused. When reading the matrix back it skips every cell that still holds `'\n'`:

```java
Arrays.fill(strRail[i], '\n');
...
if (ch != '\n') {
    encryptedString.append(ch);
}
```

The placeholder is indistinguishable from a real newline in the plaintext, so any `'\n'` in the input is silently discarded. The ciphertext comes out shorter than the plaintext and the round trip is corrupted:

```java
RailFenceCipher cipher = new RailFenceCipher();
String message = "ab\ncdef";              // 7 characters

cipher.encrypt(message, 3);               // "adbcef"  -> 6 characters, the newline is gone
cipher.decrypt(cipher.encrypt(message, 3), 3);  // "abfcde" -> not the original message
```

A transposition cipher must be a permutation of its input, so dropping characters is a correctness bug rather than a formatting quirk.

### Fix

`encrypt` now collects each rail into its own `StringBuilder` instead of writing into a matrix that needs a sentinel value. Every character of the input is appended exactly once and read back in rail order, so no character can be mistaken for a blank cell. This also drops the memory use from `O(rails * length)` to `O(length)`.

Additionally, `encrypt` and `decrypt` now reject `null` input and non-positive rail counts with `IllegalArgumentException`. Previously `encrypt("abc", 0)` threw `ArrayIndexOutOfBoundsException` and `encrypt("abc", -1)` threw `NegativeArraySizeException`, both leaking internal implementation details.

`decrypt` is unchanged: its `'*'` marker is written and consumed positionally, so a `'*'` in the ciphertext was never at risk.

### Tests

The class had no test class at all. `RailFenceCipherTest` is added, covering:

- the classic `WEAREDISCOVEREDFLEEATONCE` / 3 rails vector, in both directions
- round trips over several messages and rail counts
- control characters (`\n`, `\r`, `\t`) surviving encryption for 2..5 rails — this fails on the old code
- degenerate rail counts (`rails == 1`, `rails >= length`, empty input)
- the new validation for non-positive rail counts and `null` input

### DIFF
--- a/src/main/java/com/thealgorithms/ciphers/RailFenceCipher.java
+++ b/src/main/java/com/thealgorithms/ciphers/RailFenceCipher.java
@@ -1,7 +1,5 @@
 package com.thealgorithms.ciphers;
 
-import java.util.Arrays;
-
 /**
  * The rail fence cipher (also called a zigzag cipher) is a classical type of transposition cipher.
  * It derives its name from the manner in which encryption is performed, in analogy to a fence built with horizontal rails.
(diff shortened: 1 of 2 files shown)


DenizAltunkapan (approved):
@SEPURI-SAI-KRISHNA thanks

PR #7575, opened 2026-08-16
feat(datastructures): add SelfOrganizingLinkedList implementation and tests

iamcodinghere22:
This PR implements the Self-Organizing Linked List data structure using  the "Move To Front(MTF)" heuristic.
In this implementation, whenever an element is accessed/searched, it is dynamically moved to the head of the list. This optimizes access time for frequently requested items by keeping them near the front, achieving an {O(1)} best-case lookup time for repeated accesses.
### Changes Included
* Added `SelfOrganizingLinkedList.java` under `com.thealgorithms.datastructures.lists`.
* Added comprehensive unit test coverage in `SelfOrganizingLinkedListTest.java` verifying search, insertion, edge cases (empty list, non-existent elements), and move-to-front behavior.

---

## Testing
- [x] All unit tests in `SelfOrganizingLinkedListTest.java` pass locally (`mvn test -Dtest=SelfOrganizingLinkedListTest`).
- [x] Verified full build and test suite run successfully with JaCoCo code coverage.

---

## Checklist
- [x] Code follows project coding guidelines.
- [x] Unit tests have been added for new functionality.
- [x] No duplicate code or classes introduced.


iamcodinghere22:
Hi   @DenizAltunkapan @yanglbme @alxkm,  
All checks have passed successfully and the branch is ready.  
Could one of you please review and approve this PR so it can be merged?  
Thanks

src/main/java/com/thealgorithms/datastructures/lists/SelfOrganizingLinkedList.java
```diff
@@ -0,0 +1,98 @@
+package com.thealgorithms.datastructures.lists;
+
+import java.util.Objects;
+
+/**
+ * Node structure for the generic linked list.
+ *
+ * @param <E> the type of element held in this node
+ */
+class LinkedList<E> {
```
DenizAltunkapan: Calling the node class LinkedList is confusing, it's a node, not a list, and it clashes with java.util.LinkedList. The repo already has a convention for this (SinglyLinkedListNode). Could you rename it to Node and make it a static nested class inside SelfOrganizingLinkedList? That also keeps it to one top-level class per file, which is what the other list files do.

src/main/java/com/thealgorithms/datastructures/lists/SelfOrganizingLinkedList.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /** Returns the size of the list. */
+    public int getSize() {
+        return size;
+    }
+
+    /** Returns true if the list contains no elements. */
+    public boolean isEmpty() {
+        return head == null;
```
DenizAltunkapan: Codecov flags two uncovered lines, these getters (getSize / isEmpty) look like the culprits. Worth a small test so the patch is fully covered.

src/test/java/com/thealgorithms/datastructures/lists/SelfOrganizingLinkedListTest.java
```diff
@@ -0,0 +1,85 @@
...
+        list.insert(10);
+        list.insert(20);
+        list.insert(30);
+        list.insert(40);
+
+        // Search middle element '30'
+        assertTrue(list.search(30));
+
+        // '30' should now be the new head
+        assertEquals(30, list.getHeadValue());
```
DenizAltunkapan, line 51: The tests only check getHeadValue, so they confirm the searched element ends up at the front but never that the rest of the list survives. A broken move-to-front (lost node, cycle, wrong order) would still pass all of these. Could you add a test that walks the whole list after a search and asserts the full order plus unchanged size? A duplicate-value case would be good too.

DenizAltunkapan (changes requested):
Nice addition, a self-organizing list with move-to-front is a good fit here and there's no existing one. The algorithm itself looks correct. A few things to sort out before this can go in, left inline.

iamcodinghere22:
@DenizAltunkapan 
I’ve updated the implementation and tests to address all review comments:

Node Renaming:

Renamed LinkedList to Node and encapsulated it as a private static class Node<E> nested within SelfOrganizingLinkedList.

Eliminates collision with java.util.LinkedList and keeps the file to a single top-level class per repository convention.

Getters (getSize & isEmpty):

Added assertions for getSize() and isEmpty() across initial, single-element, and multi-element states to achieve full coverage on Codecov.

Full List Integrity & Duplicate Handling:

Added test cases that walk the entire list order after Move-To-Front search operations to ensure no nodes are dropped and no cycles are created.

Added test coverage for list behavior when duplicate values exist.

DenizAltunkapan (approved):
@iamcodinghere22 thanks!

PR #7591, opened 2026-09-05
feat: add SlidingWindowAggregator, a ring buffer with windowed aggregation

alxkm:
A fixed-capacity circular buffer that additionally keeps an aggregate of every element it currently holds, so a sliding window sum, minimum or maximum is available in constant time.

Rather than the usual "add the new value, subtract the evicted one" trick, which only works for invertible operators, it uses the two-stack sliding window aggregation algorithm: the window is split into a front section holding suffix aggregates and a back section summarised by a single running aggregate. Every element takes part in at most one rotation between the two, which makes insertion and removal O(1) amortised with no allocation after construction. Any associative operator works, commutative or not, and no identity element is needed.

Comes with factories for sum, minimum and maximum, indexed access, a fail-fast iterator, and tests that check the aggregate against a brute force reference under randomly interleaved insertions and removals.

### DIFF
--- a/src/main/java/com/thealgorithms/datastructures/buffers/SlidingWindowAggregator.java
+++ b/src/main/java/com/thealgorithms/datastructures/buffers/SlidingWindowAggregator.java
@@ -1,0 +1,447 @@
+package com.thealgorithms.datastructures.buffers;
+
+import java.util.ArrayList;
+import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


DenizAltunkapan (approved):

DenizAltunkapan:
@alxkm thanks!

